PR #19, opened 2015-08-18
travis-ci

az-qbradley:


.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+jdk:
+ - oraclejdk8
+env:
+ global:
+  - TRAVIS_JDK_VERSION=oraclejdk8
+script:
+  - sbt clean update compile test
+after_success:
+ - >
+  test "${TRAVIS_PULL_REQUEST}" = 'false' &&
```
akara: We should probably set this one to 'true', assuming what I understand is correct.

project/plugins.sbt
```diff
@@ -1,3 +1,7 @@
+resolvers += Classpaths.sbtPluginReleases
+
+resolvers += Classpaths.typesafeReleases
+
```
akara: Can you remind me what these 2 resolvers are?

project/plugins.sbt
```diff
@@ -8,3 +12,4 @@ addSbtPlugin("com.sksamuel.scapegoat" %% "sbt-scapegoat" % "1.0.0")
 
 addSbtPlugin("com.github.gseitz" % "sbt-release" % "0.8.4")
 
+addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.0.0")
```
akara: Shall we remove the sbt-pgp plugin assuming we won't sign the jar?

.travis.yml
az-qbradley: This env-var test to false prevents a pre-merger run of a pull request publish to sonatype (see here http://docs.travis-ci.com/user/pull-requests/ ), it might be desirable. I think the merge of a PR will still publish ... what do you think?

akara: Hmmm, the name is misleading. It depends on how we define 'test.' What I read here is whether we should run full tests on pull requests. The answer is obviously yes. However, it should not publish to Sonatype in a pre-PR build. If these semantics are achieved, I'm good.

Issue #23, opened 2015-08-19
Unicomplex ReportStatus not usable from outside unicomplex.


akara:
The response to ReportStatus (a tuple) exposes member types that are private to unicomplex.

akara:
This is fixed by https://github.com/paypal/squbs/commit/8d8489cb350b87e079753815be2a27ac3d55b9f7.

PR #24, opened 2015-08-19
https://github.com/paypal/squbs/issues/23: Made ReportStatus more accessible.

akara:
Also changed the response type to StatusReport case class. This allows better pattern matching without lost type information.


akara:
If accepted, this issue also needs to be cherry-picked to master.

Issue #18, opened 2015-08-18
Test RouteDefinition in squbs


elong408:
As today, if we want to test RouteDefinition using spray-testkit, we need to add some workaround to create the route instance.

e.g.

val actor = TestActorRef[DummyActor]

val route = RouteDefinition
  .startRoutes(new SellercommunitysvcSvc)(actor.underlyingActor.context).route
class DummyActor extends Actor {
  override def receive: Receive = {
    case whatever =>
  }
}

Then apply the route assertion like:

REQUEST ~> route ~> check {
  ASSERTIONS
}

akara:
Please review and comment on PR https://github.com/paypal/squbs/pull/33

akara:
Fixed by PR #33.

PR #33, opened 2015-08-21
https://github.com/paypal/squbs/issues/18 : Added TestRoute…

akara:
..to create test route and test case for the same


squbs-testkit/src/main/scala/org/squbs/testkit/TestRoute.scala
```diff
@@ -0,0 +1,52 @@
...
+  def apply[T <: RouteDefinition: ClassTag](implicit system: ActorSystem): Route = {
+    val clazz = implicitly[ClassTag[T]].runtimeClass
+    val actor = TestActorRef[EmptyActor]
+    val routeDef = RouteDefinition.startRoutes{
+      clazz.asSubclass(classOf[RouteDefinition]).newInstance()
+    }(actor.underlyingActor.context)
+    routeDef.route
+  }
+
+  class EmptyActor extends Actor {
```
zhuchenwang: Do we need to make EmptyActor private? Maybe we can have a better name, like RouteSupervisor something?

akara: Good idea to make it private, as long as TestActorRef can instantiate. As for name, It's really just a dummy actor. I'd take any good/simple name that comes along, probably not RouteSupervisor - too much power :-) Will update PR if the private passes the test.

akara:
Don't know what's happening on GitHub. All conversations/comments seem to be lost. May be due to my squash. If that's the case, we need to not squash until all comments are done, then squash n commits to 1 just before the merge.

This is pertaining to the TestRoute PR.

In summary, not using an empty actor is not feasible with all the ways investigated. We tried using ImplicitSender from Actor Testkit, but would have to change the API to the worse (need implicit ActorSystem when constructuing the TestRoute before use). Other ideas are definitely welcome.

squbs-testkit/src/main/scala/org/squbs/testkit/TestRoute.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Creates the Route to be used in Spray TestKit.
+   * @param system The ActorSystem. This will be supplied implicitly if using the ScalaTestRouteTest trait from Spray.
+   * @tparam T The RouteDefinition to be tested
+   * @return The Route to be used for testing.
+   */
+  def apply[T <: RouteDefinition: ClassTag](implicit system: ActorSystem): Route = {
+    val clazz = implicitly[ClassTag[T]].runtimeClass
+    implicit val actorContext = TestActorRef[EmptyActor].underlyingActor.context
```
zhuchenwang: I'd like to to give a name to the EmptyActor, like "TestRouteActor". This will be nice for user to debug and understand the actor hierarchy.

zhuchenwang:
I only has one comment left. After that, I am good with this PR.

Thanks,

Zhuchen

squbs-testkit/src/main/scala/org/squbs/testkit/TestRoute.scala
akara: You got it! I'll make the change.

akara:
Made the changes as suggested by @zhuchenwang. Please let me know of further comments, especially from user of this feature. After we're all good, please let me know. I'll squash the commits before merge. Please DO NOT merge before I do so.

akara:
@kuanghao I reviewed chances of using RouteActor instead of the dummy actor. The chances are not too good. The RouteActor assumes the Unicomplex infrastructure in place (with init confirmation) which is not the case for RouteTest.

Using the RouteActor as is will definitely result in dead letters. The rest is still the same. We have to reach into the actor to get its context. I'd rather do that in a dummy actor. The risk is much lower and the code is kept cleaner if we clearly separate these two use cases. Please do provide further comments/suggestions, though. Thanks!

Issue #51, opened 2015-08-29
Fix warnings issued by Codacy.


akara:

akara:
Fixed for all modules except some of pipeline, unicomplex/proxy, and httpclient.

akara:
Code is in good enough shape. A few more issues will be covered by https://github.com/paypal/squbs/issues/54

PR #52, opened 2015-08-29
https://github.com/paypal/squbs/issues/51: 

akara:
Fixed a good amount of codacy warnings.
**Note:** Please do not merge this PR yet. We need to see the resulting metrics and iterate more before merging.
**Update:** PR is ready to be merged.


akara:
If all looks good, this PR should be merged. Still left some cleanup work on pipeline, unicomplex proxy, zkcluster. That should be a separate PR. Already too much for one.

squbs-pattern/src/test/scala/org/squbs/pattern/orchestration/OFutureSpec.scala
```diff
@@ -151,9 +151,8 @@ class OFutureSpec extends FunSpec with Matchers {
     val r = new IllegalStateException("recovered")
 
     val r0 = OFuture.failed[String](o) recoverWith {
-      case _ if false == true => OFuture.successful("yay")
+      case _ if false => OFuture.successful("yay")
```
keshin: Why can't we remove the recoverWith directly?
```scala
    val r0 = OFuture.failed[String](o)
    r0.isCompleted should equal(true)
    r0.value should equal(Some(Failure(o)))
```

squbs-pattern/src/test/scala/org/squbs/pattern/timeoutpolicy/TimeoutPolicySpec.scala
```diff
@@ -25,18 +25,18 @@ import scala.util.Try
 class TimeoutPolicySpec extends FlatSpecLike with Matchers{
 
   "TimeoutPolicy Object" should "works fine" in {
-    intercept[IllegalArgumentException](TimeoutPolicy(Some(""), null, null))
-    intercept[IllegalArgumentException](TimeoutPolicy(Some(""), 1 second, fixedRule, debug = null))
+    an [IllegalArgumentException] should be thrownBy TimeoutPolicy(Some(""), null, null)
```
keshin: What's the difference between intercept and `an [Exception should be thrownBy]`?

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
```diff
@@ -139,12 +139,9 @@ object ExceptionalResponse {
   def apply(t: Throwable): ExceptionalResponse = apply(t, None)
 
   def apply(t: Throwable, originalResp: Option[NormalResponse]): ExceptionalResponse = {
-    val message = t.getMessage match {
-      case null | "" => "Service Error!"
-      case other => other
-    }
-
-    ExceptionalResponse(HttpResponse(status = StatusCodes.InternalServerError, entity = message), cause = Option(t), original = originalResp)
+    val message = Option(t.getMessage) getOrElse "Service Error!"
```
keshin: how about the case of empty string?

squbs-pattern/src/test/scala/org/squbs/pattern/orchestration/OFutureSpec.scala
akara: As far as I understand, this clause is intended to test the `recoverWith` and makes sure it is functioning correctly. If that's the case, we should retain the `recoverWith` clause.

squbs-pattern/src/test/scala/org/squbs/pattern/timeoutpolicy/TimeoutPolicySpec.scala
akara: Functionally, none. Conventionally, intercept belongs to the group of assertions. The new syntax is part of the matcher syntax. The only reason for this change is for the readability, as is the intention for all of the matcher DSL. We can definitely write every check in the form of assertions. What prompted me for this change is I had to go to the documentation to understand that intercept will fail the test if the exception is not thrown. This is much clearer in the matcher DSL.  Another one is awaitAssert and awaitCond (of Akka TestKit). Lets look at these two examples:

```
awaitCond(foo == None, 10 seconds)
```

and

```
awaitAssert(foo should be (None), 10 seconds)
```

I prefer the second style despite it being a bit longer. I have seen quite a few mistakes with awaitAssert in the ActorMonitor/ActorRegistry test code where the test always passes because of misunderstandings of the API. The matcher syntax generally makes the tests quite a bit clearer.

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
akara: I missed the empty string case. Will add. Thanks very much for noticing.

squbs-actormonitor/src/test/scala/org/squbs/actormonitor/ActorMonitorSpec.scala
```diff
@@ -115,163 +108,171 @@ class ActorMonitorSpec extends TestKit(ActorMonitorSpec.boot.actorSystem) with I
...
     "2.2) getRouteConfig of TestCube/TestActor" in {
-      ActorMonitorSpec.getActorMonitorBean("user/TestCube/TestActor", "RouteConfig") should be ("NoRouter")
+      ActorMonitorSpec.getActorMonitorBean("user/TestCube/TestActor", "RouteConfig") should be (Some("NoRouter"))
     }
 
     "2.3) getParent of TestCube/TestActor" in {
       val bean = ActorMonitorSpec.getActorMonitorBean("user/TestCube/TestActor", "Parent")
-      bean should startWith ("Actor[akka://ActorMonitorSpec/user/TestCube#")
+      bean shouldBe defined
+      bean getOrElse "" should startWith ("Actor[akka://ActorMonitorSpec/user/TestCube#")
```
zhuchenwang: Instead of using getOrElse here, can we use 2 asserts to check.
Like 
bean shouldBe defined
bean.get should startWith ("Actor[akka://ActorMonitorSpec/user/TestCube#")
Because if the bean is None, it will fail at the first assertion. Otherwise, bean.get will give you something for sure.

squbs-actorregistry/src/main/scala/org/squbs/actorregistry/ActorRegistry.scala
```diff
@@ -73,15 +74,15 @@ private[actorregistry] class ActorRegistry extends Actor with Stash {
...
-      sender !  ActorIdentity("ActorLookup", result)
+      val result = processActorLookup(lookupObj).keys.headOption
+      sender() !  ActorIdentity("ActorLookup", result)
 
     case ActorLookupMessage(lookupObj, msg) =>
       processActorLookup(lookupObj) match {
-        case result if (result.isEmpty) =>
-          sender ! org.squbs.actorregistry.ActorNotFound(lookupObj)
+        case result if result.isEmpty =>
+          sender() ! org.squbs.actorregistry.ActorNotFound(lookupObj)
```
zhuchenwang: Not sure if the convention is to using `sender()` instead of `sender`

akara: This is documented here: http://doc.akka.io/docs/akka/2.3.12/project/migration-guide-2.2.x-2.3.x.html#Parens_Added_to_sender

squbs-httpclient/src/test/scala/org/squbs/httpclient/dummy/DummyService.scala
```diff
@@ -280,7 +281,7 @@ trait DummyService extends SimpleRoutingApp {
               }
             }
         }
-    }(system, null, bindingTimeout = awaitMax).onComplete {
+    } .onComplete {
```
zhuchenwang: No need the additional `.`

squbs-actorregistry/src/main/scala/org/squbs/actorregistry/ActorRegistry.scala
zhuchenwang: Got it. Thanks!

squbs-actormonitor/src/test/scala/org/squbs/actormonitor/ActorMonitorSpec.scala
akara: That's one other option. Some of the static analysis engines are sensitive to using get and they are not smart enough to analyze multiple connected statements. The getOrElse solves both issues in one shot, as None would translate to a blank string and never start with the target.

tutufool:
Just reviewed, looks good from my side.

PR #53, opened 2015-08-29
SQUBS-818: Fixing Unicomplex to make case classes referenced in Statu…

az-qbradley:
…sReport public. Also, added proper case class for StatusReport. Allows much better type matching.

Up-merge pp 0.6.5 to master


az-qbradley:
up-merge pp 0.6.5 to master

PR #55, opened 2015-09-01
Pipeline enhance

tutufool:
1. refactor some classes from unicomplex to pipeline, most of them are from org.squbs.proxy, now moved to org.squbs.pipeline in squbs-pipeline project.

2. Simplify pipeline management, remove some classes like PipelineRegistry.


squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
```diff
@@ -25,46 +25,63 @@ import org.squbs.pipeline.ConfigHelper._
...
+    pipelines().get(name) match {
       case None =>
         configs.get(name) match {
           case Some(cfg) =>
             try {
-              val processor = Class.forName(cfg.factoryClazz).newInstance().asInstanceOf[ProcessorFactory].create(cfg.settings)
-              processors.send {
+              val factoryInstance = Class.forName(cfg.factoryClass).newInstance()
+              val entry: Either[Processor, PipelineSetting] = factoryInstance match {
+                case f if f.isInstanceOf[PipelineProcessorFactory] =>
```
akara: If possible, please use
```
case f: PipelineProcessorFactory =>
```
instead. You don't have to cast afterwards either.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
```diff
@@ -25,46 +25,63 @@ import org.squbs.pipeline.ConfigHelper._
...
+                  val ppf = f.asInstanceOf[PipelineProcessorFactory]
+                  val pipelineConfig = cfg.settings.fold(SimplePipelineConfig.empty)(SimplePipelineConfig(_))
+                  Right(PipelineSetting(ppf, Option(pipelineConfig), cfg.settings))
+                case f if f.isInstanceOf[ProcessorFactory] =>
+                  val pf = f.asInstanceOf[ProcessorFactory]
```
akara: Ditto here.

squbs-pipeline/src/main/scala/org/squbs/pipeline/Processor.scala
```diff
@@ -113,5 +115,106 @@ abstract class AbstractProcessor extends Processor
...
+    }
+  }
+
+  /**
+   * Java API
+   */
+  def create(config: Config)(implicit actorRefFactory: ActorRefFactory): SimplePipelineConfig = {
+    apply(config)
+  }
+
```
akara: If this is indeed supposed to be a Java API, you cannot use implicit parameters. Please clarify.

squbs-pipeline/src/main/scala/org/squbs/pipeline/Processor.scala
```diff
@@ -113,5 +115,106 @@ abstract class AbstractProcessor extends Processor
...
+}
+
+object SimpleProcessor {
+  def empty: SimpleProcessor = new SimpleProcessor(SimplePipelineConfig.empty)
+}
+
+class SimpleProcessorFactory extends ProcessorFactory {
+  def create(settings: Option[Config])(implicit actorRefFactory: ActorRefFactory): Option[Processor] = {
+    settings match {
+      case Some(conf) => Some(SimpleProcessor(SimplePipelineConfig(conf)))
```
akara: Using map is better than match in this case.

squbs-pipeline/src/main/scala/org/squbs/pipeline/Processor.scala
```diff
@@ -113,5 +115,106 @@ abstract class AbstractProcessor extends Processor
...
+
+class SimplePipelineResolver extends PipelineProcessorFactory {
+  def create(config: SimplePipelineConfig, setting: Option[Config])(implicit actorRefFactory: ActorRefFactory): Option[Processor] = {
+    if (SimplePipelineConfig.empty.equals(config)) None
+    else Some(SimpleProcessor(config))
+  }
+
+  def resolve(config: SimplePipelineConfig): Option[Processor] = {
+    create(config, None)(null)
+  }
```
akara: I won't use null. If the ActorRefFactory is not needed, the signature should be `Option`.

squbs-pipeline/src/test/scala/org/squbs/pipeline/PipelineManagerSpec.scala
```diff
@@ -129,6 +214,22 @@ class TestProcessorFactory2 extends ProcessorFactory {
 
 class TestProcessorFactory4
 
+class TestResolver1 extends PipelineProcessorFactory {
+  def create(config: SimplePipelineConfig, setting: Option[Config])(implicit actorRefFactory: ActorRefFactory): Option[Processor] = ???
+}
```
akara: Can we get rid of the ??? by implementing it properly?

squbs-pipeline/src/test/scala/org/squbs/pipeline/PipelineManagerSpec.scala
```diff
@@ -129,6 +214,22 @@ class TestProcessorFactory2 extends ProcessorFactory {
...
+
+class TestResolver2 extends PipelineProcessorFactory {
+  def create(config: SimplePipelineConfig, setting: Option[Config])(implicit actorRefFactory: ActorRefFactory): Option[Processor] = ???
+}
```
akara: Ditto here.

squbs-pipeline/src/test/scala/org/squbs/pipeline/PipelineManagerSpec.scala
```diff
@@ -129,6 +214,22 @@ class TestProcessorFactory2 extends ProcessorFactory {
...
+
+class TestResolver4
+
+class TestHandlerFactory extends HandlerFactory with Handler {
+  override def create(config: Option[Config])(implicit actorRefFactory: ActorRefFactory): Option[Handler] = Some(this)
+
+  override def process(reqCtx: RequestContext)(implicit executor: ExecutionContext, context: ActorContext): Future[RequestContext] = ???
+}
```
akara: And here.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/Unicomplex.scala
```diff
@@ -219,7 +219,7 @@ class Unicomplex extends Actor with Stash with ActorLogging {
       import scala.collection.JavaConversions._
       extensions map { e =>
         val (phase, ex) = e.exceptions.headOption map {
-          case (iPhase, exception) => (iPhase, exception.toString)
+          case (phase, exception) => (phase, exception.toString())
         } getOrElse (("", ""))
```
akara: Using the same variable name can shadow the external variable (phase on line 221). That's why they should be using different names.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
tutufool: fixed as suggested

squbs-actormonitor/src/test/scala/org/squbs/actormonitor/ActorMonitorSpec.scala
```diff
@@ -95,9 +95,6 @@ class ActorMonitorSpec extends TestKit(ActorMonitorSpec.boot.actorSystem) with I
       case other => logger.warn(s"beforeAll invalid identity: $other")
     }
 
-    import ActorMonitorSpec._
-    val cfgBeanCount = getActorMonitorConfigBean("Count").getOrElse(-100)
-    cfgBeanCount should be > 11
```
akara: Why did we get rid of this check?

squbs-httpclient/src/main/scala/org/squbs/httpclient/endpoint/HttpClientEndpoint.scala
```diff
@@ -35,16 +35,16 @@ object Endpoint {
...
   def resolve(svcName: String, env: Environment = Default): Option[Endpoint]
 }
 
-abstract class PipelineAwareEndpointResolver(system : ActorSystem) extends EndpointResolver {
+abstract class PipelineAwareEndpointResolver(system: ActorSystem) extends EndpointResolver {
 
-  protected def getPipelineSetting(name : String) : Option[PipelineSetting] = {
-    PipelineRegistry(system).get(name)(system)
+  protected def getPipelineSetting(name: String): Option[PipelineSetting] = {
+    PipelineManager(system).getPipelineSetting(name)(system)
```
akara: Single line functions don't need an enclosing `{` and `}`

squbs-httpclient/src/test/scala/org/squbs/httpclient/pipeline/HttpClientPipelineActorSpec.scala
```diff
@@ -34,7 +34,7 @@ class HttpClientPipelineActorSpec extends TestKit(ActorSystem("HttpClientPipelin
   import system.dispatcher
 
   val endpoint = Endpoint(DummyService.dummyServiceEndpoint)
-  val processor = SimplePipelineResolver.INSTANCE.resolve(SimplePipelineConfig.empty)
+  val processor = SimplePipelineResolver.INSTANCE.create(SimplePipelineConfig.empty, None)
```
akara: Just a note - no correction needed for this PR. The use of INSTANCE is very Java-like. Is this intended to be a Java API? Please do remind me. Thx!

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
```diff
@@ -25,46 +25,61 @@ import org.squbs.pipeline.ConfigHelper._
 import scala.collection.JavaConversions._
 import scala.collection.mutable
 
-case class PipelineManager(configs: Map[String, ProxyConfig], processors: Agent[Map[String, Processor]]) extends Extension with LazyLogging {
+
+case class PipelineManager(configs: Map[String, RawPipelineSetting], pipelines: Agent[Map[String, Either[Processor, PipelineSetting]]]) extends Extension with LazyLogging {
```
akara: Please tell me more about the use of `Either` here. We have the same field of two types represented by an `Either`. Can you please elaborate on the design?

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
```diff
@@ -25,46 +25,61 @@ import org.squbs.pipeline.ConfigHelper._
...
 
   def default(implicit actorRefFactory: ActorRefFactory) = {
-    get("default-proxy")
+    getProcessor("default-proxy")
+  }
```
akara: Here too, we don't need the braces.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
```diff
@@ -25,46 +25,61 @@ import org.squbs.pipeline.ConfigHelper._
...
+
+  def getProcessor(name: String)(implicit actorRefFactory: ActorRefFactory): Option[Processor] = {
+    get(name) match {
+      case Some(Left(v)) => Option(v)
+      case Some(Right(v)) => v.factory.create(v.pipelineConfig, v.setting)
+      case None if (name.equals("default-proxy")) => None
```
akara: Do we really need this? Is there anybody besides PipelineManager.default that will call into getProcessor?

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
```diff
@@ -25,46 +25,61 @@ import org.squbs.pipeline.ConfigHelper._
 import scala.collection.JavaConversions._
 import scala.collection.mutable
 
-case class PipelineManager(configs: Map[String, ProxyConfig], processors: Agent[Map[String, Processor]]) extends Extension with LazyLogging {
+
+case class PipelineManager(configs: Map[String, RawPipelineSetting], pipelines: Agent[Map[String, Either[Processor, PipelineSetting]]]) extends Extension with LazyLogging {
 
   def default(implicit actorRefFactory: ActorRefFactory) = {
-    get("default-proxy")
+    getProcessor("default-proxy")
```
akara: This whole function could be a val. It always returns None based on the code below.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
```diff
@@ -89,18 +121,22 @@ object PipelineManager extends ExtensionId[PipelineManager] with ExtensionIdProv
...
           val aliasNames = subCfg.getOptionalStringList("aliases") getOrElse Seq.empty[String]
-          val proxyConf = ProxyConfig(key, aliasNames, subCfg.getString("processorFactory"), subCfg.getOptionalConfig("settings"))
+          val factoryClass = subCfg.getOptionalString("processorFactory") match {
+            case Some(text) => text
+            case None => subCfg.getString("factory")
+          }
+          val proxyConf = RawPipelineSetting(key, aliasNames, factoryClass, subCfg.getOptionalConfig("settings"))
           key +: aliasNames foreach { name =>
             proxyMap.get(name) match {
               case None => proxyMap.put(name, proxyConf)
```
akara: Old code, but this can also be represented as `proxyMap += (name -> proxyConf)`
Not quite used to this one anymore :-)

squbs-pipeline/src/main/scala/org/squbs/pipeline/Processor.scala
```diff
@@ -113,5 +115,99 @@ abstract class AbstractProcessor extends Processor
...
+}
+
+class SimpleProcessorFactory extends ProcessorFactory {
+  def create(settings: Option[Config])(implicit actorRefFactory: ActorRefFactory): Option[Processor] = {
+    settings.map(conf => SimpleProcessor(SimplePipelineConfig(conf)))
+  }
+}
+
+object SimplePipelineResolver {
+  val INSTANCE = new SimplePipelineResolver
```
akara: Here too. Unless it is needed by the Java API, we don't really need INSTANCE. Please clarify.

akara:
A few more comments. Couple of rules I need to ask for:
1. Please squash the commits into 1 commit.
2. Please put the issue URL into the commit text.

These rules and link to how to squash are here: https://github.com/paypal/squbs/blob/master/CONTRIBUTING.md

Thanks!

squbs-httpclient/src/test/scala/org/squbs/httpclient/pipeline/HttpClientPipelineActorSpec.scala
tutufool: The original code:

object SimplePipelineResolver {
  val INSTANCE = new SimplePipelineResolver
}

class SimplePipelineResolver extends PipelineProcessorFactory {
  def create(config: SimplePipelineConfig, setting: Option[Config])(implicit actorRefFactory: ActorRefFactory): Option[Processor] = {
    if (SimplePipelineConfig.empty.equals(config)) None
    else Some(SimpleProcessor(config))
  }
}

Shall we even remove the singleton INSTANCE and let user always create it with new?

We can also make SimplePipelineResolver from class to an object.
But that requires some extra logic when loading factory from config:

first try to new the factory class with name and then try to new it with name$
(I think we should not let user config a class name with $ even if it's an object)

Let me know your suggestions.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
```diff
@@ -25,46 +25,61 @@ import org.squbs.pipeline.ConfigHelper._
...
             try {
-              val processor = Class.forName(cfg.factoryClazz).newInstance().asInstanceOf[ProcessorFactory].create(cfg.settings)
-              processors.send {
+              val factoryInstance = Class.forName(cfg.factoryClass).newInstance()
+              val entry: Either[Processor, PipelineSetting] = factoryInstance match {
+                case f : PipelineProcessorFactory =>
+                  val pipelineConfig = cfg.settings.fold(SimplePipelineConfig.empty)(SimplePipelineConfig(_))
+                  Right(PipelineSetting(f, Option(pipelineConfig), cfg.settings))
+                case f : ProcessorFactory =>
+                  Left(f.create(cfg.settings).getOrElse(null))
```
akara: Codacy Issue found: [Avoid using null](https://www.codacy.com/app/akara-sucharitakul/squbs/file/947004612/issues/source?bid=2834325&fileBranchId=2840190#l65)

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
tutufool: For most squbs proxy that used by squbs-services, a Processor is good to go.
Because usually it's static, 100% determined by the config.
Or we can say "we eagerly create a processor here"

For squbs proxy used by http client, we might need to lazy create the processor on demand.
Because the httpclient could apply per request handlers to the pipeline.
Then, here we need a "lazy producer", with PipelineSetting + some user defined pipeline handlers, we can create the Processor on demand.

Actually, we can consolidate them into 1 , i.e. use PipelineSetting  for both cases.
The only downside is we might create a processor  for several times if multiple squbs-services share the same proxy definition.

Let me know if you have better ideas

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
tutufool: Done

squbs-httpclient/src/main/scala/org/squbs/httpclient/endpoint/HttpClientEndpoint.scala
tutufool: Done

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
tutufool: Yes, we need it in handling StartCubeService:

case StartCubeService(webContext, listeners, props, name, proxyName, initRequired) =>


      // Caution: The serviceActor may be the cubeActor in case of no proxy, or the proxy in case there is a proxy.
      val (cubeActor, serviceActor) = try {
        proxyName.fold(pipelineManager.default) {
          case "" => None
          case other => pipelineManager.getProcessor(other)
        } match {
          case None =>
            val hostActor = context.actorOf(props, name) // disable proxy
            (hostActor, SimpleActor(hostActor))
          case Some(proc) =>
            val hostActor = context.actorOf(props)
            (hostActor, ProxiedActor(context.actorOf(Props(classOf[CubeProxyActor], proc, hostActor), name)))
        }
      } catch {
        case t: Throwable =>
          log.error(t, "Cube {} proxy with name of {} initialized failed.", name, proxyName)
          val hostActor = context.actorOf(props, name) // disable proxy
          initMap += hostActor -> Some(Failure(t))
          (hostActor, SimpleActor(hostActor))
      }

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
tutufool: Yes, it could be a val if we don't expect some dynamic config load in the future.

The logic is:

If a proxy named "default-proxy" found in config, then we use the config to create a processor.
otherwise, we return NONE. (treat it as no proxy provided)

If the config is static, it could be a val.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
tutufool: Yes, your proposal is actually efficient according to the impl.
Saving a "find" operation, since we don't need a return value here.

will change it.

squbs-pipeline/src/main/scala/org/squbs/pipeline/Processor.scala
tutufool: How about doing it like this:

object SimplePipelineResolver extends PipelineProcessorFactory {
  def create(config: SimplePipelineConfig, setting: Option[Config])(implicit actorRefFactory: ActorRefFactory): Option[Processor] = {
    if (SimplePipelineConfig.empty.equals(config)) None
    else Some(SimpleProcessor(config))
  }
}

//for easy reflection from config
class SimplePipelineResolver extends PipelineProcessorFactory {
  def create(config: SimplePipelineConfig, setting: Option[Config])(implicit actorRefFactory: ActorRefFactory): Option[Processor] = {
    SimplePipelineResolver.create(config, setting)
  }
}

We don't need the class definition if we enhance the reflection logic to search "com.ebay.MyFactory$" as well.

what do you think?

tutufool: My bad, brain got stucked:

This should be OK, right?

object SimplePipelineResolver extends SimplePipelineResolver

class SimplePipelineResolver extends PipelineProcessorFactory {
  def create(config: SimplePipelineConfig, setting: Option[Config])(implicit actorRefFactory: ActorRefFactory): Option[Processor] = {
    if (SimplePipelineConfig.empty.equals(config)) None
    else Some(SimpleProcessor(config))
  }
}

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
akara: OK.

squbs-httpclient/src/test/scala/org/squbs/httpclient/pipeline/HttpClientPipelineActorSpec.scala
akara: I'm thinking more of the second choice, to make it an object.  But I get your point. Because the INSTANCE came from a factory, it may not be a good idea to use an object to it. Lets leave it as is. Thx!

akara:
Please just take care of the nulls, squash, and ensure you have the issue# in the commit. I'm good to merge if that's in place. Thx!

tutufool:
Use https://github.com/paypal/squbs/pull/63 instead

PR #57, opened 2015-09-01
allow codacy for pull requests

az-qbradley:


akara:
Looks like we have to resort back to unencrypted if we want functionality at PR time, right?

Issue #54, opened 2015-09-01
Fix null issues reported by Codacy


akara:
Especially in pipeline, unicomplex/proxy, httpclient.

akara:
Resolved all null and code analysis errors.

PR #68, opened 2015-09-10
Fixed static analysis for issue #54

akara:
Commits are now squashed and ready to merge. Please review and comment.

### DIFF
--- a/squbs-actormonitor/src/main/scala/org/squbs/actormonitor/ActorMonitorBean.scala
+++ b/squbs-actormonitor/src/main/scala/org/squbs/actormonitor/ActorMonitorBean.scala
@@ -22,6 +22,9 @@ import javax.management.MXBean
 import akka.actor.{ActorContext, ActorRef, Props}
 import org.squbs.unicomplex.JMX._
 
+import scala.annotation.tailrec
+import scala.util.{Failure, Success, Try}
+
 
 private[actormonitor] object ActorMonitorBean {
   val Pattern = "org.squbs.unicomplex:type=ActorMonitor,name="
@@ -32,7 +35,8 @@ private[actormonitor] object ActorMonitorBean {
       register(new ActorMonitorBean(actor), objName(actor))
   }
 
-  def totalBeans(implicit context: ActorContext) = ManagementFactory.getPlatformMBeanServer.queryNames(prefix + Total, null)
+  def totalBeans(implicit context: ActorContext) =
+    ManagementFactory.getPlatformMBeanServer.queryNames(prefix + Total, null)
 
   def unregisterBean(actor: ActorRef) (implicit context: ActorContext) = unregister(objName(actor))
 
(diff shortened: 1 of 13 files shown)


squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClientManager.scala
```diff
@@ -123,25 +126,27 @@ trait HttpCallActorSupport extends PipelineManager with CircuitBreakerSupport {
 
   private def verifyUri(client: HttpClient, uri: String) = {
     val baseUri = Uri(client.endpoint.uri)
```
tutufool: verifyUri is heavily used and almost all usage of endpoint.uri is to create a Uri

So, I'm thinking of change the definition from:

case class Endpoint(uri: String, config: Configuration = Configuration())

to 

case class Endpoint(uri: Uri, config: Configuration = Configuration())

Uri is a structured object which carries more information that a String.

Then we don't need to parse a string into Uri every time.

What do you think?

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClientManager.scala
```diff
@@ -123,25 +126,27 @@ trait HttpCallActorSupport extends PipelineManager with CircuitBreakerSupport {
...
-    val basePath = baseUri.path.toString
-    var trimBasePath = basePath
-    if (trimBasePath != "" && trimBasePath.charAt(0) == '/') {
-      trimBasePath = trimBasePath.substring(1)
-    }
-    if (trimBasePath != "" && trimBasePath.charAt(trimBasePath.length - 1) == '/') {
-      trimBasePath = trimBasePath.substring(0, trimBasePath.length - 1)
-    }
-    (trimBasePath, uri.charAt(0)) match {
+    val path = baseUri.path.toString()
```
tutufool: Instead of using toString and do String operation, Maybe we can just use Path.startWith and Path.+ ?
I roughly went thru impls of Path, it seems the different impl like Slash or segment should take care of the last '/'

akara: I actually changed some code based on your suggestion and I fully agree with what you're saying here. However, I don't think this change fits into the scope of #54. I suggest having a separate issue for this httpclient optimization.

Furthermore, I think the method name verifyUri is wrong. It is actually composing the Uri path. Also, the uri parameter name is wrong. It is actually a path to add to the client. We should correct all as part of the different issue.

squbs-actormonitor/src/main/scala/org/squbs/actormonitor/ActorMonitorBean.scala
```diff
@@ -32,7 +35,8 @@ private[actormonitor] object ActorMonitorBean {
       register(new ActorMonitorBean(actor), objName(actor))
   }
 
-  def totalBeans(implicit context: ActorContext) = ManagementFactory.getPlatformMBeanServer.queryNames(prefix + Total, null)
+  def totalBeans(implicit context: ActorContext) =
+    ManagementFactory.getPlatformMBeanServer.queryNames(prefix + Total, null)
```
akara: Codacy Issue found: [Avoid using null](https://www.codacy.com/app/akara-sucharitakul/squbs/file/960826412/issues/source?bid=2834325&fileBranchId=2857651#l39)

squbs-httpclient/src/main/scala/org/squbs/httpclient/json/JacksonProtocol.scala
```diff
@@ -17,21 +17,16 @@
...
 import spray.httpx.unmarshalling.Unmarshaller
 
-/**
- * Created by lma on 6/23/2015.
- */
 object JacksonProtocol {
 
   //key = class name
-  var mappers = Map[String, ObjectMapper]()
+  var mappers = Map.empty[String, ObjectMapper]
```
akara: Codacy Issue found: [Avoid mutable fields](https://www.codacy.com/app/akara-sucharitakul/squbs/file/960826435/issues/source?bid=2834325&fileBranchId=2857651#l29)

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClientManager.scala
akara: Created #72 to track fixing the Path optimization.

squbs-httpclient/src/test/scala/org/squbs/httpclient/dummy/DummyService.scala
```diff
@@ -163,7 +178,8 @@ trait DummyService extends SimpleRoutingApp {
   import DummyService._
   import org.squbs.testkit.Timeouts._
 
-  def startDummyService(implicit system: ActorSystem, address: String = dummyServiceIpAddress, port: Int = dummyServicePort) {
+  def startDummyService(implicit system: ActorSystem, address: String = dummyServiceIpAddress,
```
akara: Codacy Issue found: [Method is longer than 50 lines](https://www.codacy.com/app/akara-sucharitakul/squbs/file/968879430/issues/source?bid=2834325&fileBranchId=2857651#l181)

akara: Codacy Issue found: [You should not name methods after their object](https://www.codacy.com/app/akara-sucharitakul/squbs/file/968879430/issues/source?bid=2834325&fileBranchId=2857651#l181)

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClientManager.scala
```diff
@@ -58,7 +60,8 @@ trait HttpCallActorSupport extends PipelineManager with CircuitBreakerSupport {
              pipeline: Try[HttpRequest => Future[HttpResponse]],
              httpRequest: HttpRequest)(implicit system: ActorSystem): Future[HttpResponse] = {
     implicit val ec = system.dispatcher
-    httpClientLogger.debug("HttpRequest headers:" + httpRequest.headers.foreach(header => println(s"${header.name}=${header.value},")))
+    httpClientLogger.debug("HttpRequest headers:" +
+      httpRequest.headers.foreach(header => println(s"${header.name}=${header.value},")))
```
keshin: 1. Why we use `println` on logging?, is it a typo?
2. com.typesafe.scalalogging.Logger only supports call by value, which means, we should try to use `(message, args*)` if possible.

akara: I checked. It has been there for a long time, perhaps even from the beginning. Let me fix.

Issue #72, opened 2015-09-13
Fix verifyURI in HttpClientManager to use Spray Uri path construct


akara:
This is more efficient than the current String manipulations. We should also pass Uri and not String as parameter.

PR #73, opened 2015-09-14
#72: Optimized HttpClientManager Uri.Path usage

akara:
Renamed verifyUri to makeFullUri as it fits the intent of the function better.

### DIFF
--- a/squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClientManager.scala
+++ b/squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClientManager.scala
@@ -16,22 +16,20 @@
 package org.squbs.httpclient
 
 import akka.actor._
-import spray.http._
-import spray.can.Http
 import akka.io.IO
 import akka.pattern._
 import akka.util.Timeout
+import org.squbs.httpclient.env.Environment
+import org.squbs.httpclient.pipeline.PipelineManager
+import spray.can.Http
+import spray.http.{HttpRequest, HttpResponse, _}
 import spray.httpx.marshalling.Marshaller
+
+import scala.annotation.tailrec
 import scala.collection.concurrent.TrieMap
 import scala.concurrent._
 import scala.concurrent.duration.{Duration, FiniteDuration}
-import scala.util.Try
-import scala.util.Failure
-import spray.http.HttpResponse
-import scala.util.Success
-import spray.http.HttpRequest
-import org.squbs.httpclient.env.Environment
-import org.squbs.httpclient.pipeline.PipelineManager
+import scala.util.{Failure, Success, Try}
 
 class HttpClientManagerExtension(system: ExtendedActorSystem) extends Extension {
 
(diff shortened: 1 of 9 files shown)


squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClientManager.scala
```diff
@@ -76,72 +90,64 @@ trait HttpCallActorSupport extends PipelineManager with CircuitBreakerSupport {
...
+  private def makeFullUri(client: HttpClient, path: String): Uri = {
+    import HttpClientManager._
+    val basePath = client.endpoint.uri.path
+    val fullPath = (basePath.endsWithSlash, path.charAt(0) == '/') match {
+      case (false, false) => basePath + ('/' + path)
+      case (true, false)  => basePath + path
+      case (false, true)  => basePath + path
+      case (true, true)   => basePath + path.substring(1)
     }
+    Uri(path = fullPath)
```
akara: Debating between providing just a Uri with path (current) or a full Uri, as in
```
client.endpoint.uri.copy(path = fullPath)
```
This will simplify logging above and may be beneficial for making the request. But will need to see into the client further whether we can avoid another composition.

Both cases pass the test.

squbs-httpclient/src/main/scala/org/squbs/httpclient/pipeline/HttpClientPipeline.scala
```diff
@@ -114,7 +114,7 @@ trait PipelineManager extends LazyLogging {
   }
 
   implicit def endpointToUri(endpoint: Endpoint): String = {
```
akara: Not sure where this conversion is used. The change may have broken it.

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClientManager.scala
tutufool: I think current path is good to go, because the full Uri is only used in HttpClientCallerActor:

  override def receive: Actor.Receive = {
    case HttpClientActorMessage.Get(uri, reqSettings) =>
      IO(Http) ! hostConnectorSetup(client, reqSettings)
      context.become(receiveGetConnection(HttpMethods.GET, uri, reqSettings, client, sender()))

squbs-httpclient/src/main/scala/org/squbs/httpclient/pipeline/HttpClientPipeline.scala
tutufool: Comment it out and seems no compile errors locally.

Also it should not break it, we can simply remove the first line of the method body:

  implicit def endpointToUri(endpoint: Endpoint): String = {
    val uri = Uri(endpoint.uri)  ---- already a uri?
    val host = uri.authority.host.toString
    val port = if (uri.effectivePort == 0) 80 else uri.effectivePort
    val isSecure = uri.scheme.toLowerCase.equals("https")
    (isSecure, port) match {
      case (true, 443) => s"https://$host"
      case (true, _)   => s"https://$host:$port"
      case (false, 80) => s"http://$host"
      case (false, _)  => s"http://$host:$port"
    }
  }

Issue #80, opened 2015-09-18
HttpClient performance enhancement


tutufool:
Simplify httpclient with less actor creation.

PR #87, opened 2015-09-21
https://github.com/paypal/squbs/issues/80

tutufool:
This patch:

1. simplify duplicate code for different type of messages handling like GET, POST, etc.
2. For each httpclient call, save 3 actor creations, they're:
HttpClientCallerActor, HttpClientPipelineTargetActor and HttpClientPipelineActor


squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClientCircuitBreaker.scala
```diff
@@ -44,7 +44,7 @@ case class CircuitBreakerMetrics(var status: CircuitBreakerStatus,
 
 trait CircuitBreakerSupport{
 
-  def withCircuitBreaker(client: HttpClient, response: Future[HttpResponse])(implicit system: ActorSystem) = {
+  def withCircuitBreaker(client: HttpClient, response: => Future[HttpResponse])(implicit system: ActorSystem) = {
```
akara: Not clear why you choose to by-name the response?

tutufool: According to the definition:

  def withCircuitBreaker[T](body: ⇒ Future[T]): Future[T] = currentState.invoke(body)

This call is meant to protect the function that get the Future[T]

If we use by-value, which means we already call spray connector to get the Future[Response]

  def handle(client: HttpClient,
             pipeline: Try[HttpRequest => Future[HttpResponse]],
             httpRequest: HttpRequest)(implicit system: ActorSystem): Future[HttpResponse] = {
    implicit val ec = system.dispatcher
    httpClientLogger.debug("HttpRequest headers: " +
      httpRequest.headers.map { h => h.name + '=' + h.value } .mkString(", "))
    pipeline match {
      case Success(res) =>
        withCircuitBreaker(client, res(httpRequest))

----- we don't want to execute "res" function here, right?

akara:
After working on the perf fork by @az-qbradley, I found this PR missed a test. Can you please take a look at ChunkedRequestSpec.scala line 89? That still starts the actor with the old signature and the file is not part of this PR. There may be other places. Thanks!

tutufool:
The test failure is fixed.
which is caused by refactoring of PipelineProcessorActor, the constructor is changed.
I've sent email to Qian about how to fix this issues for similar cases in rocksqubs.

PR #107, opened 2015-10-13
#80: HttpClientState split from HttpClient

akara:
Make sure there is only one state holder per client.


akara:
Sorry for the many commits. These are not squashable due to the merges in between. Will have to stay like this as an exception unless we redo the branch.

akara:
Also, please wait for review comments from @tutufool.

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
```diff
@@ -63,50 +64,13 @@ trait HttpClientPathBuilder {
...
 
-case class HttpClient(name: String,
-                      env: Environment = Default,
-                      status: Status.Status = Status.UP,
-                      config: Option[Configuration] = None)
-                     (actorCreator: (ActorRefFactory) => Future[ActorRef] = {(arf) =>
-                        Future.successful(arf.actorOf(Props(classOf[HttpClientActor], name, env)))
-                      })(implicit actorRefFactory: ActorRefFactory) { self =>
+class HttpClient private[httpclient] (val name: String, val env: Environment = Default,
+                                           config: Option[Configuration] = None, val askTimeout: Timeout)
```
tutufool: If possible, I think maybe we can restructure class Configuration: (maybe make it flattern)

Current definition:

case class Configuration(pipeline: Option[PipelineSetting], settings: Settings)

case class Settings(hostSettings: HostConnectorSettings, connectionType: ClientConnectionType,
                    sslContext: Option[SSLContext], circuitBreakerConfig: CircuitBreakerSettings)

Some of the configs are for HttpClient level like PipelineSetting, HostConnectorSettings
Some are for HttpClientState like circuitBreakerConfig

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
```diff
@@ -63,50 +64,13 @@ trait HttpClientPathBuilder {
...
+                                          (actorCreator: (ActorRefFactory) => Future[ActorRef])
```
tutufool: This is like an declaration that "I need an actor to do things"
I'd propose to hide the implementation details into another layer like HttpClientStateHolder

This HttpClientStateHolder should:

1. hold the HttpClientActor
2. responsible to get the latest HttpClientState from TrieMap
3. consume HttpClientMessage like:
def consume(message : HttpClientMessage) : Future[HttpResponse]
(we can extract current  case msg: HttpClientMessage => from HttpClientActor into this method to avoid unnecessary pipeTo)


So here maybe we can use Future[HttpClientStateHolder]

tutufool: em, maybe HttpClientStateHolder is better than Future[HttpClientStateHolder] since HttpClientStateHolder is a wrapper lighweight class and can be created on the fly

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
```diff
@@ -118,7 +82,7 @@ case class HttpClient(name: String,
   }
 
   def get[R](uri: String)(implicit unmarshaller: FromResponseUnmarshaller[R]): Future[R] = {
-    implicit val timeout = defaultTimeout
+    implicit val timeout = askTimeout
     val fHttpResponse = fActorRef flatMap { ref => (ref ? HttpClientActorMessage.Get(uri, None)).mapTo[HttpResponse] }
```
tutufool: If we go with HttpClientStateHolder, it could be:

val fHttpResponse = HttpClientStateHolder.consume(HttpClientActorMessage.Get(uri, None))

tutufool: Also, the HttpClientMessage messages like GET, POST, etc should be redefined by carry on configuration that belongs to this httpclient.

  case class Get(uri: String, requestSettings: Option[RequestSettings] = None, config : Option[Configuration]) extends HttpClientMessage {
    val requestBuilder : (Uri) => HttpRequest = RequestBuilding.Get(_)
  }

Otherwise it will always use the default one on HttpClientState

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
```diff
@@ -206,30 +170,26 @@ case class HttpClient(name: String,
 
   def raw = new RawHttpClient(this)
 
-  val defaultTimeout = Configuration.defaultRequestSettings(endpoint.config, config).timeout.askTimeout
-
   def withConfig(config: Configuration): HttpClient = {
     implicit val timeout = defaultFutureTimeout
-    HttpClient(name, env, status, Some(config))(
+    val newAskTimeout = toTimeout(config.settings.hostSettings.connectionSettings.requestTimeout).askTimeout
+    new HttpClient(name, env, Some(config), newAskTimeout)(
```
tutufool: This method should purely update "This" HttpClient.

to update global HttpClientState config, we'd better have method on HttpClientStateHolder like:

def update(config: UpdateConfig) : Unit

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
```diff
@@ -206,30 +170,26 @@ case class HttpClient(name: String,
...
       (_) => fActorRef flatMap { ref => (ref ? HttpClientActorMessage.UpdateConfig(config)).mapTo[ActorRef] }
     )
   }
 
   def withSettings(settings: Settings): HttpClient = {
     implicit val timeout = defaultFutureTimeout
-    HttpClient(name, env, status, Some(config.getOrElse(Configuration()).copy(settings = settings)))(
+    val newAskTimeout = toTimeout(settings.hostSettings.connectionSettings.requestTimeout).askTimeout
+    new HttpClient(name, env, Some(config.getOrElse(Configuration().copy(settings = settings))), newAskTimeout)(
```
tutufool: same as above, we need to separate the logic that "update this httpclient" or "update the HttpClientState"

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
```diff
@@ -318,11 +279,55 @@ class RawHttpClient(private val client : HttpClient) {
...
+    }
+  }
+
+  import endpoint.config.settings.{circuitBreakerConfig => cbConfig}
+
+  private[httpclient] var cbStat = CircuitBreakerStatus.Closed
+  private[httpclient] val cbMetrics =
+    new CircuitBreakerMetrics(cbConfig.historyUnits, cbConfig.historyUnitDuration)(system)
+  private[httpclient] val cb: CircuitBreaker = new CircuitBreaker(system.scheduler, cbConfig.maxFailures,
+    cbConfig.callTimeout, cbConfig.resetTimeout)(system.dispatcher)
```
tutufool: These CircuitBreaker related vals will be reset as long as we do copy on HttpClientState, all statistics will be clean up.
Need to confirm this behavior is correct or not.

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
```diff
@@ -330,9 +335,17 @@ object HttpClientFactory {
...
+
+    val clientAskTimeout =
+      toTimeout(endpoint.config.settings.hostSettings.connectionSettings.requestTimeout).askTimeout
+
+
     implicit val timeout = defaultFutureTimeout
-    HttpClient(name, newEnv, Status.UP, None)(
-    (_) => (HttpClientManager(system).httpClientManager ? HttpClientManagerMessage.Get(name, newEnv)).mapTo[ActorRef]
+    new HttpClient(name, newEnv, None, clientAskTimeout)((_) =>
+      (HttpClientManager(system).httpClientManager ? HttpClientManagerMessage.Get(name, newEnv)).mapTo[ActorRef]
```
tutufool: As described above, how about return a strong typed HttpClientStateHolder instead of a Future[ActorRef]

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClientManager.scala
```diff
@@ -129,12 +124,6 @@ class HttpClientActor(svcName: String, env: Environment, clientMap: TrieMap[(Str
...
   def client = clientMap(svcName, env)
 
-  implicit def toTimeout(d: Duration): Timeout = Timeout(d match {
-    case f: FiniteDuration => f
-    case Duration.Inf => Duration.fromNanos(Long.MaxValue)
-    case _ => Duration.Zero
-  })
-
   override def receive: Actor.Receive = {
     case msg: HttpClientMessage =>
```
tutufool: this part can be extracted to a sync call block like a method in HttpClientStateHolder to avoid pipeTo.
Also, to avoid messages pile up. (several httpclient of same endpoint could send messages at the same time)

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClientManager.scala
```diff
@@ -129,12 +124,6 @@ class HttpClientActor(svcName: String, env: Environment, clientMap: TrieMap[(Str
 
   def client = clientMap(svcName, env)
```
tutufool: 1 call per request, will the global TrieMap provide enough performance?

I used to propose the restructure the TrieMap to:

TrieMap[(String, Environment), Agent[HttpClientState]] 

But his might need more memory if the cached endpoint number is too big.

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
akara: I'd really love to revisit the configuration structure. It is way messy. But this will likely break compatibility. Can you attack this in 0.8 the same time as Akka Http (client)?

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
akara: Point by point...
1. I need the HttpClientActor for all operations. That's true. But this is not a public API. As you can see, the constructor is private.
2. Having the HttpClientStateHolder hold the HttpClient Actor. I'm actually proposing the opposite. Fold the HttpClientState into the HttpClientActor. The HttpClientManager holds the map (not the extension) and fetches the HttpClientActor from the map. So instead of wrapping the actor, I'm thinking more of unwrapping and put the state where it belongs. Actors are great at dealing with state.
3. `consume(message: HttpClientMessage)`. I'm not sure I like having this layer more than interacting with the actor. Another stack frame to go through and yet a different abstraction. The pipeTo is just sending the Future directly to the temp ask actor sending this request. You'll pay that price anyway.

IMO, it is hard to justify yet another layer in the API. I'd rather unwrap than wrap, especially none of these are user-facing anyway.

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
akara: Yes, but again, this is not user API. It is hidden inside the HttpClient. I'm not sure how I can explain HttpClientStateHolder. Also, you'll have to get that object from somewhere. If we have to synchronously get the object out of the map, we'll have frequent concurrent access to the map.

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
akara: Yup. But again, this change can be done only for 0.8, not for this PR (which is 0.7.1). The client should be immutable. A new client with new settings should be created as a result of `withSettings()`. The client config should never be able to mutate. But that's too late now.

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
tutufool: Sure, I can do it later

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
tutufool: I'm a little worry about using 1 HttpClientActor to serve all messages.
Some of the messages are from HttpClient like GET(might be multiple instances of httpclient) which are high priority.
Some of the messages are management messages like updating config which are low priority.

1 Actor will make them pile up, right?

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
tutufool: let's do it in 0.8

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
akara: These vals should be re-initialized every time we copy.

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClient.scala
akara: The actor model as it stands today is not strongly typed. We're just masking it with a strongly-typed façade. Moreover, this is not user API. I'm really not sure we should drive it that way.

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClientManager.scala
akara: You'll need to go with the strategy of caching the HostConnector to solve this problem properly.

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClientManager.scala
akara: So far the TrieMap is not showing itself as an issue at 16,000 accesses per second. We could consider this but I think we need to do a thorough design review for 0.8.

akara:
In summary, there is a whole lot we can do. But this is a patch for 0.7.1. I really want to limit the amount of changes to solve just the problem at hand. Most of your suggestions would fit into 0.8.0. But then, we need to look at Akka Http. Many of our assumptions would be no longer right in view of that change. I'd do a thorough design review (no code) for HttpClient 0.8.0 with consideration on config management, actor hops, connector caching, usage of ask vs tell in every step, proper support for streaming content, etc. But that is 0.8. Lets keep the scope of this change limited to what is feasible as a patch. Even the amount of change I've made feels much. We just want to get rid of the high memory usage for unused CBStat instances.

Issue #113, opened 2015-10-22
Statistics Collection and Reporting


anilgursel:
- Need an accurate timing for sub millisecond level.
- Potential use of HdrHistogram
- Abstract StatsBucket to a common facility allows all stats to display historical data
- Potential support for longer term historical data: minutes, hours, days

anilgursel:
Provide utilities to collect metrics using:

HdrHistogram
Dropwizard

anilgursel:
Client side metrics collection addressed with https://github.com/paypal/squbs/pull/303.  Will integrate the same flow on the server side.

PR #114, opened 2015-10-23
https://github.com/paypal/squbs/issues/113: Start separating out TimeBucketMetrics

akara:
Extract functionality for more generic use. Will need to find a new home at a later stage.


akara:
Huh, it created from master and the PR takes everything from 0.7.0, which is kind of wanted, but not intended for this PR.

Issue #117, opened 2015-10-28
TestRoute does not allow passing web context


akara:
TestRoute throws exception when route `extends RouteDefinition with WebContext`. This is because the route is not started with the WebContext closure.

PR #118, opened 2015-10-28
https://github.com/paypal/squbs/issues/117: Added support for WebCont…

akara:
…ext to TestRoute, including an additional signature allowing manual passing of WebContext as String.

### DIFF
--- a/squbs-testkit/src/main/scala/org/squbs/testkit/TestRoute.scala
+++ b/squbs-testkit/src/main/scala/org/squbs/testkit/TestRoute.scala
@@ -18,8 +18,9 @@ package org.squbs.testkit
 
 import akka.actor.{Actor, ActorSystem}
 import akka.testkit.TestActorRef
-import org.squbs.unicomplex.RouteDefinition
+import org.squbs.unicomplex.{WebContext, RouteDefinition}
 import spray.routing._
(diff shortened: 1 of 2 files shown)


squbs-testkit/src/main/scala/org/squbs/testkit/TestRoute.scala
```diff
@@ -38,12 +39,32 @@ object TestRoute {
   def apply[T <: RouteDefinition: ClassTag](implicit system: ActorSystem): Route = {
     val clazz = implicitly[ClassTag[T]].runtimeClass
     implicit val actorContext = TestActorRef[TestRouteActor].underlyingActor.context
-    val routeDef = RouteDefinition.startRoutes{
-      clazz.asSubclass(classOf[RouteDefinition]).newInstance()
+    val routeDef = WebContext.createWithContext("") {
```
anilgursel: Can we update the logic of this `apply` as calling the other `apply` with webcontext as `""`?  The only difference seems to be the last line `pathPrefix(separateOnSlashes(webContext)) {routeDef.route}`, which I am not sure would cause a problem.

akara: Good point. Will do. Thanks!

Issue #160, opened 2016-01-07
JacksonProtocol failed to deserialize certain Java classes


akara:
If these are Jackson-annotated JavaBeans and and bean B extends bean A, the annotated `@JsonProperty` on bean A are not recognized.

akara:
Testing with vanilla Jackson, the serialization goes through, though.

PR #164, opened 2016-01-09
#160: Removed any Jackson module registration

akara:
Code needing certain modules need to register themselves. Also added test cases covering the issue.


tutufool:
Looks OK to me.
It makes sense to let user register what he wants instead of providing a default one.
Jackson might be out dated some day.

Issue #161, opened 2016-01-07
Get PipelineException deserializing HttpResponse with status 204


akara:
Instead, the exception should be returned back to the user with the original response so we can check it is a 204. No deserialization should have been attempted.

PR #165, opened 2016-01-09
#161: Fixed deserialization to check for no content

akara:
And send back an exception with the response. Also added test cases for no content deserialization.


squbs-httpclient/src/main/scala/org/squbs/httpclient/pipeline/HttpClientPipeline.scala
```diff
@@ -41,10 +41,11 @@ object HttpClientUnmarshal{
 
     def unmarshalTo[T: FromResponseUnmarshaller]: Try[T] = {
       Try{
-        if (response.status.isSuccess)
+        if (response.status.isSuccess && response.status.allowsEntity && response.entity.nonEmpty)
```
tutufool: Does it mean 204 response will be treated as UnsuccessfulResponseException?

tutufool:
I'm OK with code changes if 204 response should be treated as UnsuccessfulResponseException.

Issue #162, opened 2016-01-07
Java HttpClient API lacks means to easily set accept headers


akara:
... without going through hoops of nested `apply` calls on `HttpHeader`, `Accept`. Also, MediaTypes are not very accessible.

PR #166, opened 2016-01-10
#162: Made RequestSettings API independent from ActorRefFactory…

akara:
… as well as added more Java access methods to RequestSettings and facilities for Spray Java access.


squbs-pattern/src/test/java/org/squbs/pattern/spray/japi/JApiTest.java
```diff
@@ -26,6 +25,7 @@
 
 import java.util.ArrayList;
 import java.util.Arrays;
```
akara: Codacy Issue found: [Avoid unused imports such as 'java.util.Arrays'](https://www.codacy.com/app/akara-sucharitakul/squbs/file/1745745653/issues/source?bid=2888638&fileBranchId=3059628#l27)

squbs-pattern/src/main/scala/org/squbs/pattern/spray/japi/MediaTypes.scala
```diff
@@ -0,0 +1,36 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.squbs.pattern.spray.japi
+
+import spray.http.MediaType
+import spray.http.MediaTypes._
+
+object MediaTypes {
```
kittskumar: This implementation seems simplistic.  Why not reuse:

https://github.com/spray/spray/blob/master/spray-http/src/main/scala/spray/http/MediaType.scala

It provides a more complete set of MediaTypes.

squbs-httpclient/src/main/scala/org/squbs/httpclient/package.scala
```diff
@@ -36,10 +40,11 @@ package object httpclient {
...
   })
 
-  implicit class RequestToAskTimeout(val requestTimeout: Timeout) extends AnyVal {
+  implicit class RequestToAskTimeout(val reqTimeout: Option[Timeout]) extends AnyVal {
 
-    def askTimeout: Timeout = {
-      val Timeout(d) = requestTimeout
+    def askTimeout(implicit refFactory: ActorRefFactory): Timeout = {
+      import Configuration._
+      val Timeout(d) = reqTimeout getOrElse Timeout(requestTimeout(defaultHostSettings), TimeUnit.MILLISECONDS)
```
anilgursel: Just a minor comment..  This may potentially change `TimeUnit`.  If we show this somewhere on VI, just might confuse.  But, most likely in value will be in milliseconds, so should have much impact..

squbs-pattern/src/main/scala/org/squbs/pattern/spray/japi/ContentTypes.scala
```diff
@@ -0,0 +1,26 @@
...
+package org.squbs.pattern.spray.japi
+
+import spray.http.ContentTypes._
+
+object ContentTypes {
+
+  final val APPLICATION_JSON = `application/json`
+  final val APPLICATION_OCTET_STREAM = `application/octet-stream`
+  final val TEXT_PLAIN = `text/plain`
+  final val TEXT_PLAIN_UTF8 = `text/plain(UTF-8)`
```
anilgursel: I believe since this is a temporary use till we move to Akka Http, we are not having the full list.

akara: Correct. We just support the common types as a convenience here.

squbs-httpclient/src/main/scala/org/squbs/httpclient/package.scala
akara: There is another issue for 0.8 that will preserve the time unit from configuration. Unfortunately, this will need the new Typesafe config only available with Akka 2.4 to obtain the correct time unit from configuration.  We'll need to cover this change as part of 0.8.

anilgursel: I agree.

PR #186, opened 2016-03-19
#176 Upgrade to Akka Http on the server side

anilgursel:
* #176 Support Akka Http experimental mode along with Spray
* #110 Streaming Pipeline
* #111 Upgrade to Akka 2.4.2


squbs-streamingpipeline/src/main/scala/org/squbs/pipeline/streaming/package.scala
```diff
@@ -0,0 +1,48 @@
...
+import akka.NotUsed
+import akka.stream.BidiShape
+import akka.stream.scaladsl.{Merge, Broadcast, GraphDSL, BidiFlow}
+
+package object streaming {
+
+  // (BidiFlow name, defaults on/off )
+  type PipelineSetting = (Option[String], Option[Boolean])
+
+  type PipelineFlow = BidiFlow[RequestContext, RequestContext, RequestContext, RequestContext, NotUsed]
```
az-qbradley: Do we know that Mat is not needed

anilgursel: At this moment, we do not use it.  However, depending on the scenarios, we may need it later on.  Good point!

I may change that to `Any`, which would also align with the core Http flow `Flow[HttpRequest, HttpResponse, Any]`.

PR #193, opened 2016-04-20
#192 TestRoute should honor exceptionHandler and rejectionHandler

anilgursel:


squbs-testkit/src/main/scala/org/squbs/testkit/streaming/TestRoute.scala
```diff
@@ -54,8 +55,13 @@ object TestRoute {
         clazz.asSubclass(classOf[RouteDefinition]).newInstance()
       }
     }
-    if (webContext.length > 0) pathPrefix(separateOnSlashes(webContext)) {routeDef.route}
-    else routeDef.route
+
+    implicit val routingSettings = RoutingSettings(system.settings.config)
+    implicit val rejectionHandler:RejectionHandler = routeDef.rejectionHandler.getOrElse(RejectionHandler.default)
+    implicit val exceptionHandler:ExceptionHandler = routeDef.exceptionHandler.getOrElse(PartialFunction.empty[Throwable, Route])
+
```
akara: You should not need to pass the type arguments to `PartialFunction.empty`. Please try.

akara:
Other than that, I'm good.

squbs-testkit/src/main/scala/org/squbs/testkit/streaming/TestRoute.scala
anilgursel: That was a copy paste sin :)  Removing..

Issue #217, opened 2016-05-27
Let unicomplex auto-select the port when configured as test mode


anilgursel:
In some unit-tests, we need to bootstrap unicomplex, which requires a port to be passed in the configuration.  We cannot use a static port as we do not know where the tests will be running.  So, we have been grabbing a port by:

* Bind to a socket
* Get port
* Close the socket

Please see https://github.com/spray/spray/blob/master/spray-util/src/main/scala/spray/util/Utils.scala#L44 (for Akka Http, we use our own utility as it is not exposed by Akka).  

The problem with the above is that the selected port can be grabbed by another test between the time socket is closed and unicomplex binds it.  This introduces test instability.

In tests, we should instead pass the port as "0" to Spray and Akka Http bind methods and let the system select the port.  Then, tests should ask unicomplex what the port is.  Basically, reversing the port information passing.

PR #218, opened 2016-05-27
#217 Let unicomplex auto-select the port when configured as test mode

anilgursel:


squbs-unicomplex/src/main/scala/org/squbs/unicomplex/ServiceRegistry.scala
```diff
@@ -218,12 +218,18 @@ class ServiceRegistry(val log: LoggingAdapter) extends ServiceRegistryBase[Path]
...
+        serviceListeners map { case (name, ServiceListenerInfo(squbsListener, httpListener, _, exception)) =>
           ListenerState(name, squbsListener.map(_ => "Success").getOrElse("Failed"), exception.getOrElse("").toString)
         } toSeq
       }
     }
   }
+
+  override private[unicomplex] def portBindings: Map[String, Int] = {
+    serviceListeners map { case(name, ServiceListenerInfo(_, _, Some(serverBinding), None)) =>
+      (name -> serverBinding.localAddress.getPort)
```
akara: Doesn't feel like you need parentheses here.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/ServiceRegistryBase.scala
```diff
@@ -89,13 +89,16 @@ trait ServiceRegistryBase[A] {
...
+
   case class BindConfig(interface: String, port: Int, localPort: Option[Int],
                         ssLContext: Option[SSLContext], needClientAuth: Boolean)
 
   protected def bindConfig(config: Config): BindConfig = {
     val interface = if (config getBoolean "full-address") ConfigUtil.ipv4
     else config getString "bind-address"
-    val port = config getInt "bind-port"
+    // If test-mode is on, let the system pick the port
+    val port = if(config.getBoolean("test-mode")) 0 else config.getInt("bind-port")
```
akara: Question on the design: Why do we set a `test-mode` flag instead of setting `bind-port = 0` indicating this is a port that should be allocated by the system?

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
```diff
@@ -475,9 +475,15 @@ object UnicomplexBoot extends LazyLogging {
...
+    val testModeConfig = ConfigFactory.parseString {
+      s"""
+        | test-mode = ${actorSystem.settings.config getOptionalBoolean "squbs.unicomplex.test-mode" getOrElse false}
+      """.stripMargin
+    }
+
     val ackFutures =
       for ((listenerName, config) <- boot.listeners) yield {
-        Unicomplex(actorSystem).uniActor ? StartListener(listenerName, config)
+        Unicomplex(actorSystem).uniActor ? StartListener(listenerName, config.withFallback(testModeConfig))
```
akara: If we were to use 0, would we have any change here?

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/ServiceRegistry.scala
anilgursel: Will remove

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/ServiceRegistryBase.scala
anilgursel: I agree!

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
anilgursel: No change would be needed, will change.

akara:
Just one last thing. Need to find a place to document this. I think under "Testing" is the right place. We can file an issue to address the documentation separately, though - so we can at least take care of the build stability issues not holding it hostage to the docs. What do you think? If agree I'll merge.

anilgursel:
Sure.  Let's merge it and I will start adding docs in a separate PR.

PR #219, opened 2016-06-02
#217 Test utility for retrieving the port from unicomplex

anilgursel:


docs/testing.md
```diff
@@ -3,79 +3,136 @@ All tests on squbs have been written using ScalaTest 2.x. Specs2 has not yet bee
...
-import org.squbs.unicomplex.{UnicomplexBoot, JMX}
+```
 
-// First, setup an object that configures and creates the
-// UnicomplexBoot (boot, in short) for this test.
+By default, `PortGetter` returns `default-listener`'s port, which is the most common one.  If you need to retrieve another listener's bind port, you can either override `listener` method or pass listener name to `port` method:
 
-object MyTest {
+```scala
+class SampleSpec extends CustomTestKit("SampleSpec", SampleSpec.classPaths)
```
akara: Do we have an option not to pass a classpath if it is the current classpath of the test? You can use `scanResources()` to use the current classpath.

docs/testing.md
```diff
@@ -3,79 +3,136 @@ All tests on squbs have been written using ScalaTest 2.x. Specs2 has not yet bee
...
+       |  ${JMX.prefixConfig} = true # to prevent JMX name collision, if you are doing any JMX testing
        |}
     """.stripMargin
   )
 
-  // Now, start the boot with the given config.
+  // This expects a squbs-meta.conf file under src/test/resources/SampleSpec/META-INF/
+  val classPaths = Seq("SampleSpec") map {
+    getClass.getClassLoader.getResource("").getPath + "/" + _
+  }
```
akara: This is what we do for Unicomplex testing. Do we do this commonly in app tests?

squbs-testkit/src/main/scala/org/squbs/testkit/CustomTestKit.scala
```diff
@@ -29,25 +30,53 @@ object CustomTestKit {
   private[testkit] def checkInit(actorSystem: ActorSystem) {
     if (actorSystems.putIfAbsent(actorSystem.name, actorSystem) == None)
       sys.addShutdownHook {
-        actorSystem.shutdown()
+        actorSystem.terminate()
       }
   }
+
+  val emptyClassPath = Seq.empty[String]
```
akara: What does an empty classpath mean? Current classpath or no classpath (and so no components at all)?

docs/testing.md
anilgursel: `classPath` is not mandatory, but having an option to pass it is important, IMO.  e.g, when you would like to start two separate set of services for two different tests.  We can definitely default it to `src/test/resources` or `scanResources()`..  But, my preference was not to have a default classpath at all and be explicit about it.

docs/testing.md
anilgursel: Not sure how common it is used by the apps, but providing the option in important for flexibility.

squbs-testkit/src/main/scala/org/squbs/testkit/CustomTestKit.scala
anilgursel: As discussed in previous comments, it is no components.  But, I am open to defaulting to something else..  Defaulting to `src/test/resources/squbs-meta.conf` makes sense.

docs/testing.md
```diff
@@ -1,81 +1,121 @@
...
 
-2. `org.squbs.testkit.CustomTestKit` For those who need ultimate power and flexibility to control your tests and be able to run different test cases needing different configurations together and in parallel, CustomTestKit is the right answer. You can start any number of ActorSystems and Unicomplex instances, one per ActorSystem, with different configurations - all on the same JVM. You just have to ensure the ActorSystem and Unicomplex instances do not conflict.
+   * It automatically sets actor system name to spec/test class name for simplicity and also to ensure `ActorSystem` instances do not conflict.  But, it also allows actor system name to be passed in the constructor.
+   * Tests extending `CustomTestKit` can run in parallel in the same JVM.
+   * Starts and stops squbs automatically.
+   * Starts well-known actors and service in `src/test/resources/META-INF/squbs-meta.conf` by default.  But, allows passing `resources` to be scanned in the constructor.
+   * Allows custom configuration to be passed in the constructor.
```
anilgursel: Should we also read `src/test/resources/test.conf` by default?

squbs-testkit/src/main/scala/org/squbs/testkit/CustomTestKit.scala
```diff
@@ -29,25 +30,76 @@ object CustomTestKit {
...
       }
   }
+
+  val stackTraceDepth = 6 // CustomTestKit$ x 3 -> Option -> CustomTestKit$ -> CustomTestKit -> Spec
+
+  def defaultActorSystemName = actorSystemNameFrom((new Exception).getStackTrace.apply(stackTraceDepth).getClassName)
+
+  def actorSystemNameFrom(className: String) =
+    className
+      .replace('.', '-')
```
dmartyanov: Probably, it might be faster
```scala
className collect {
  case '.' => '-'
  case '_' => '-'
  case c if c != '$' => c
}
```

anilgursel: I pretty much got if from here: https://github.com/akka/akka/blob/master/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTest.scala#L35, as it is not exposed in a separate util.  Since this is test code, so I would prefer readability over speed.  But, even for the speed wise, can you elaborate more?  Is it because your version would traverse the characters once?  I believe scala compiler optimizes this and traverses only once.  @akara Any comments?

squbs-testkit/src/main/scala/org/squbs/testkit/CustomTestKit.scala
```diff
@@ -29,25 +30,76 @@ object CustomTestKit {
   private[testkit] def checkInit(actorSystem: ActorSystem) {
     if (actorSystems.putIfAbsent(actorSystem.name, actorSystem) == None)
       sys.addShutdownHook {
-        actorSystem.shutdown()
+        actorSystem.terminate()
```
akara: Just one thought. On the good side, `terminate()` is now returning a `Future`. Shall we block on that `Future`?

squbs-testkit/src/main/scala/org/squbs/testkit/CustomTestKit.scala
akara: Thanks for the comments. I don't think we have problems reading both and as @anilgursel said, this is not performance sensitive. I'd accept both ways. But @dmartyanov is right, the collection version has the potential to be faster. Great idea. Thanks @dmartyanov!

squbs-testkit/src/main/scala/org/squbs/testkit/SimpleTestKit.scala
```diff
@@ -50,6 +50,7 @@ object SimpleTestKit {
  * the component scanning from the classpath (as in the regular squbs runtime). The test process must be forked and
  * no parallel tests with different squbs configurations are allowed.
  */
+@deprecated("use org.squbs.testkit.CustomTestKit instead")
```
akara: Good move!

squbs-testkit/src/test/scala/org/squbs/testkit/CustomTestKitSpec.scala
```diff
@@ -68,10 +66,148 @@ object CustomTestKitSpec {
...
   lazy val boot = UnicomplexBoot(testConfig)
-    .scanComponents(Seq(new File("src/test/resources/CustomTestKitTest").getAbsolutePath))
+    .scanResources(getClass.getClassLoader.getResource("").getPath + "/CustomTestKitTest/META-INF/squbs-meta.conf")
     .start()
 }
 
+class CustomTestKitDefaultSpec extends CustomTestKit with FlatSpecLike with Matchers {
+
+  it should "start default-listener" in {
+    port shouldEqual port("default-listener") // not throwing an exception means, it is bound
```
akara: If you want to validate this is not throwing an exception, the following is better than a comment:

```scala
noException should be thrownBy port
```

squbs-testkit/src/test/scala/org/squbs/testkit/CustomTestKitSpec.scala
```diff
@@ -68,10 +66,148 @@ object CustomTestKitSpec {
...
+    system.settings.config.getInt("default-listener.bind-port") should be(0)
+    system.settings.config.getBoolean(s"squbs.${JMX.prefixConfig}") should be(true)
+  }
+}
+
+class CustomTestKitDefaultWithActorSystemNameSpec extends CustomTestKit("CustomTestKitDefaultWithActorSystemNameSpec")
+with FlatSpecLike with Matchers {
+
+  it should "start default-listener" in {
+    port shouldEqual port("default-listener") // not throwing an exception means, it is bound
```
akara: Ditto everywhere you have this comment.

docs/testing.md
akara: I don't think so. If somebody wants to use test.conf, they can load it.

akara:
Looks very nice! Just a few thoughts.

PR #223, opened 2016-06-15
#200 Validation Directives on Akka Http

dmartyanov:
- added unicomplex/streaming/ValidationDirectives based on wix/accord library
- added dependency on akka-http-testkit
- added tests for ValidationDirectives


anilgursel:
The spray version is in `squbs-pattern`, so we are moving this to `squbs-unicomplex` with Akka Http version?

akara:
Oh, sorry. I missed that during the review. Should have waited for you. Lets move it to squbs-pattern.

Issue #140, opened 2015-12-03
ConfigUtil.getOptionalDuration changes time unit


anilgursel:
`ConfigUtil.getOptionalDuration` should keep the time unit.

akara:
Sorry, we need the new version of Typesafe config and therefore Akka 2.4 to have the right getDuration signature retaining the time unit. This cannot be implemented with Akka 2.3.X

anilgursel:
`scala.concurrent.duration.Duration` has `length` and `unit` methods; however, `java.time.Duration` does not.  So, it gets problematic and requires a complicated logic, which is probably not worth it.

Please see https://github.com/typesafehub/config/issues/165.

akara:
I'd make a few changes to ConfigUtil to solve this issue, as follows:

```scala
private val DurationTag = typeTag[Duration]
private val FiniteDurationTag = typeTag[FiniteDuration]

...
case DurationTag => Some(Duration(underlying.getString(path))
case FiniteDurationTag => Some(Duration(underlying.getString(path)).asInstanceOf[FiniteDuration])
```

Then we have to add tests to test the finite case and the infinite case. The infinite case should throw a ClassCastException.

akara:
This means we're not going through the `getDuration` signature at all and still achieve the desired results. Please let me know what you think.

dmartyanov:
@akara this approach looks good to me,  we don't have to use `getDuration` signature if it loses some data.
I have only one concern about `ClassCastException` for infinite case, because essentially it is not a failure of upcast,  it is string value that doesn't correspond to  any value from the set of permissible values defined by `FiniteDuration` type. We can return just `None` as for any other incorrect string.
In general I suppose that `getOption[T]` shouldn't throw exceptions if it's used properly, as it were done in `getOptionalString` and other methods.

akara:
Good point. Let's check before casting or catch the exception. Thanks!

dmartyanov:
If we need to inject more type cast logic into `RichConfig` I would like to suggest the following:
Extract method `getTry[T]: Try[T]` and implement all the logic of fetching config values and type casting with proper exceptions handling.  Execute `getTry[T]` inside `getOption[T]` and divide all the exceptions into 2 groups for Failure case: 
1. Exceptions of data in config and it's representation, they should be recovered with None
2. Exceptions of incorrect method call,  like `IllegalArgumentException`,  they should be thrown downstream

akara:
[...]

PR #224, opened 2016-06-16
#140 ConfigUtil.getOptionalDuration changes time unit

dmartyanov:
- added Duration typetag to unicomplex/ConfigUtil and pattern/ConfigUtil
- added FiniteDuration typetag to unicomplex/ConfigUtil and pattern/ConfigUtil
- added tests
- refactored config calls

### DIFF
--- a/squbs-pattern/src/main/scala/org/squbs/pattern/util/ConfigUtil.scala
+++ b/squbs-pattern/src/main/scala/org/squbs/pattern/util/ConfigUtil.scala
@@ -17,42 +17,64 @@ package org.squbs.pattern.util
 
 import java.net.{Inet4Address, NetworkInterface}
 
+import com.typesafe.config.ConfigException.{Missing, WrongType}
 import com.typesafe.config.{Config, ConfigException, ConfigMemorySize}
+import com.typesafe.scalalogging.LazyLogging
 
 import scala.collection.JavaConversions._
 import scala.concurrent.duration._
 import scala.reflect.runtime.universe._
+import scala.util.{Failure, Success, Try}
 
-object ConfigUtil {
+object ConfigUtil extends LazyLogging {
 
   private val StringTag = typeTag[String]
   private val StringListTag = typeTag[Seq[String]]
   private val ConfigTag = typeTag[Config]
   private val ConfigListTag = typeTag[Seq[Config]]
-  private val DurationTag = typeTag[FiniteDuration]
+  private val DurationTag = typeTag[Duration]
+  private val FiniteDurationTag = typeTag[FiniteDuration]
   private val ConfigMemorySizeTag = typeTag[ConfigMemorySize]
 
   implicit class RichConfig(val underlying: Config) extends AnyVal {
 
-    def getOption[T](path: String)(implicit tag: TypeTag[T]): Option[T] =
-      if (underlying.hasPath(path))
-        (tag match {
-            case StringTag => Some(underlying.getString(path))
-            case StringListTag => Some(underlying.getStringList(path).toSeq)
-            case TypeTag.Int => Some(underlying.getInt(path))
-            case TypeTag.Boolean => Some(underlying.getBoolean(path))
-            case TypeTag.Double => Some(underlying.getDouble(path))
-            case ConfigTag => Some(underlying.getConfig(path))
-            case ConfigListTag => Some(underlying.getConfigList(path).toSeq)
-            case DurationTag => Some(Duration.create(underlying.getDuration(path, MILLISECONDS), MILLISECONDS))
-            case ConfigMemorySizeTag => Some(underlying.getMemorySize(path))
-            case _ => throw new IllegalArgumentException(s"Configuration option type $tag not implemented")
-          }).asInstanceOf[Option[T]]
-      else None
+    def getTry[T](path: String)(implicit tag: TypeTag[T]): Try[T] = Try {
+      (typeTag[T] match {
+        case StringTag => underlying.getString(path)
+        case StringListTag => underlying.getStringList(path).toSeq
+        case TypeTag.Int => underlying.getInt(path)
+        case TypeTag.Boolean => underlying.getBoolean(path)
+        case TypeTag.Double => underlying.getDouble(path)
+        case ConfigTag => underlying.getConfig(path)
+        case ConfigListTag => underlying.getConfigList(path).toSeq
+        case FiniteDurationTag => Duration(underlying.getString(path)).asInstanceOf[FiniteDuration]
+        case DurationTag => Duration(underlying.getString(path))
+        case ConfigMemorySizeTag => underlying.getMemorySize(path)
+        case _ =>
+          throw new IllegalArgumentException(s"Configuration option type ${typeTag[T].tpe} not implemented")
+      }).asInstanceOf[T]
+    } recover {
+      case e: IllegalArgumentException => throw e
+      case e: Missing => throw e
+      case e: WrongType => throw e
+      case e => throw new WrongType(underlying.origin,
+        s"Path: $path, value ${underlying.getString(path)} is not a ${typeTag[T].tpe}", e)
+    }
+
+    def getOption[T: TypeTag](path: String): Option[T] =
+      getTry[T](path) match {
+        case Success(value) => Some(value)
+        case Failure(e: ConfigException.Missing) => None
+        case Failure(e: IllegalArgumentException) => throw e
+        case Failure(e) =>
+          logger.warn("Value at path {} has an illegal format for type{}: {}",
+            path, typeTag[T].tpe, underlying.getString(path))
+          None
+      }
 
     def get[T: TypeTag](path: String, default: => T) = getOption[T](path).getOrElse(default)
 
-    def get[T: TypeTag](path: String) = getOption[T](path).getOrElse(throw new ConfigException.Missing(path))
+    def get[T: TypeTag](path: String) = getTry[T](path).get
 
     def getOptionalString(path: String): Option[String] = {
       try {
--- a/squbs-pattern/src/test/scala/org/squbs/pattern/util/ConfigUtilSpec.scala
+++ b/squbs-pattern/src/test/scala/org/squbs/pattern/util/ConfigUtilSpec.scala
@@ -20,6 +20,7 @@ import com.typesafe.config.{Config, ConfigException, ConfigFactory, ConfigMemory
 import org.scalatest.{FunSpecLike, Matchers}
 
 import scala.concurrent.duration._
+import scala.util.Success
 
 class ConfigUtilSpec extends FunSpecLike with Matchers {
 
@@ -45,6 +46,7 @@ class ConfigUtilSpec extends FunSpecLike with Matchers {
       |    }
       |  ]
       |  timeout = 20s
+      |  timeout-inf = Inf
       |  mem-size = 4m
       |}
     """.stripMargin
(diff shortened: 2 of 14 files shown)


squbs-pattern/src/main/scala/org/squbs/pattern/util/ConfigUtil.scala
```diff
@@ -22,37 +22,46 @@ import com.typesafe.config.{Config, ConfigException, ConfigMemorySize}
...
+        case _ => throw new IllegalArgumentException(s"Configuration option type $tag not implemented")
+      }).asInstanceOf[T]
+    }
+
+    def getOption[T: TypeTag](path: String): Option[T] =
+      getTry[T](path) match {
+        case Success(value) => Some(value)
+        case Failure(e: IllegalArgumentException) => throw e
+        case Failure(e) => None
+      }
```
akara: I'm just thinking out loud. Should the result of `ConfigException.Missing` and a `ParseException` or `NumberFormatException` be the same? That would lead to more of `getTry` usage to distinguish between the two.

akara: Thinking from the usage stand point, there is no question if `getOption` returns a `None` in case of `ConfigException.Missing`. However, if there is a value for that config entry but it is in the wrong format, the user will generally log the issue and continue as `None`. Would it mean we should log the issue for them (just in case of `getOption`)? In that case, this code should be:

```scala
def getOption[T: TypeTag](path: String): Option[T] =
  getTry[T](path) match {
    case Success(value) => Some(value)
    case Failure(e: ConfigException.Missing) => None
    case Failure(e: IllegalArgumentException) => throw e
    case Failure(e) =>
      log.warn("Path {} has an illegal format for type {}: {}", path, typeTag, value)
      None
}
```

squbs-pattern/src/main/scala/org/squbs/pattern/util/ConfigUtil.scala
```diff
@@ -22,37 +22,46 @@ import com.typesafe.config.{Config, ConfigException, ConfigMemorySize}
...
-            case TypeTag.Boolean => Some(underlying.getBoolean(path))
-            case TypeTag.Double => Some(underlying.getDouble(path))
-            case ConfigTag => Some(underlying.getConfig(path))
-            case ConfigListTag => Some(underlying.getConfigList(path).toSeq)
-            case DurationTag => Some(Duration.create(underlying.getDuration(path, MILLISECONDS), MILLISECONDS))
-            case ConfigMemorySizeTag => Some(underlying.getMemorySize(path))
-            case _ => throw new IllegalArgumentException(s"Configuration option type $tag not implemented")
-          }).asInstanceOf[Option[T]]
-      else None
+    def getTry[T](path: String)(implicit tag: TypeTag[T]): Try[T] = Try {
```
akara: You can spare yourself a bit by declaring `def getTry[T: TypeTag](path: String): Try[T] = ...`
Then you can access the tag by calling `typeTag[T] match {` on line 40 below instead of just `tag`.

squbs-pattern/src/test/scala/org/squbs/pattern/util/ConfigUtilSpec.scala
```diff
@@ -53,163 +55,206 @@ class ConfigUtilSpec extends FunSpecLike with Matchers {
...
-    it ("should get None for non-existing string config") {
-      config.getOptionalString("str") should be (None)
+    it("should get None for non-existing string config") {
+      config.getOptionalString("str") should be(None)
     }
 
-    it ("should get None for non-existing string config by \"getOption\"") {
-      config.getOption[String]("str") should be (None)
+    it("should get Failure(ConfigException.Missing) for non-existing string config by \"getTry\"") {
+      config.getTry[String]("str").failed.get.isInstanceOf[ConfigException.Missing] shouldBe true
```
akara:
```scala
config.getTry[String]("str").failed.get shouldBe a [ConfigException.Missing]
```

squbs-pattern/src/test/scala/org/squbs/pattern/util/ConfigUtilSpec.scala
```diff
@@ -53,163 +55,206 @@ class ConfigUtilSpec extends FunSpecLike with Matchers {
...
-    it ("should get None for non-existing string list config") {
-      config.getOptionalStringList("str-list") should be (None)
+    it("should get None for non-existing string list config") {
+      config.getOptionalStringList("str-list") should be(None)
     }
 
-    it ("should get None for non-existing string list config by \"getOption\"") {
-      config.getOption[Seq[String]]("str-list") should be (None)
+    it("should get Failure(ConfigException.Missing) for non-existing string list config by \"getTry\"") {
+      config.getTry[Seq[String]]("str-list").failed.get.isInstanceOf[ConfigException.Missing] shouldBe true
```
akara: Ditto.

squbs-pattern/src/test/scala/org/squbs/pattern/util/ConfigUtilSpec.scala
```diff
@@ -53,163 +55,206 @@ class ConfigUtilSpec extends FunSpecLike with Matchers {
...
-    it ("should get None for non-existing int config") {
-      config.getOptionalInt("int-num") should be (None)
+    it("should get None for non-existing int config") {
+      config.getOptionalInt("int-num") should be(None)
     }
 
-    it ("should get None for non-existing int config by \"getOption\"") {
-      config.getOption[Int]("int-num") should be (None)
+    it("should get Failure(ConfigException.Missing) for non-existing int config by \"getTry\"") {
+      config.getTry[Int]("int-num").failed.get.isInstanceOf[ConfigException.Missing] shouldBe true
```
akara: Too.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/ConfigUtil.scala
```diff
@@ -33,30 +33,38 @@ object ConfigUtil {
...
-          case TypeTag.Double => Some(underlying.getDouble(path))
-          case ConfigTag => Some(underlying.getConfig(path))
-          case ConfigListTag => Some(underlying.getConfigList(path).toSeq)
-          case RegexTag => Try(new Regex(underlying.getString(path))).toOption
-          case ConfigMemorySizeTag => Some(underlying.getMemorySize(path))
-          case DurationTag => Some(Duration.create(underlying.getDuration(path, MILLISECONDS), MILLISECONDS))
-          case _ => throw new IllegalArgumentException(s"Configuration option type $tag not implemented")
-        }).asInstanceOf[Option[T]]
-      else None
+    def getTry[T](path: String)(implicit tag: TypeTag[T]): Try[T] = Try {
```
akara: Similar comments in the file as above.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
```diff
@@ -360,40 +361,35 @@ object UnicomplexBoot extends LazyLogging {
...
+        val className = serviceConfig.getString("class-name")
+        val clazz = Class.forName(className, true, getClass.getClassLoader)
+        val proxyName = serviceConfig.getOption[String]("proxy-name") map (_.trim)
+        val webContext = serviceConfig.getString("web-context")
+        val pipeline = serviceConfig.getOption[String]("pipeline")
+        val defaultFlowsOn = serviceConfig.getOption[Boolean]("defaultPipelineOn")
+        val streamingPipelineSettings = (pipeline, defaultFlowsOn)
+
+        val listeners = serviceConfig.getOption[Seq[String]]("listeners").fold(Seq("default-listener")) { list =>
+          if (list.contains("*")) aliases.values.toSeq.distinct
```
akara: Just a marker note. This changes the alias logic. I need to come back and read in detail later today.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/ConfigUtilSpec.scala
```diff
@@ -62,180 +66,223 @@ class ConfigUtilSpec extends FunSpecLike with Matchers {
...
     }
     
     it ("should get None for non-existing string config") {
       config.getOptionalString("str") should be (None)
     }
 
-    it ("should get None for non-existing string config by \"getOption\"") {
-      config.getOption[String]("str") should be (None)
+    it("should get Failure(ConfigException.Missing) for non-existing string config by \"getTry\"") {
+      config.getTry[String]("str").failed.get.isInstanceOf[ConfigException.Missing] shouldBe true
```
akara: Ditto on all lines like this.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
```diff
@@ -418,50 +414,43 @@ object UnicomplexBoot extends LazyLogging {
...
+      logger.warn(s"Duplicate alias $alias for listener $listener already declared for listener $l. Ignoring.")
+    )
+  }
+
+  def resolveDuplicates[T](in: Seq[(String, T)], duplicateHandler: (String, T, T) => Unit): Map[String, T] = {
+    in.groupBy(_._1).map {
+      case (key, Seq((k, v))) => key -> v
+      case (key, head::tail) =>
+        tail.foreach { case (k, ass) => duplicateHandler(k, ass, head._2)}
+        key -> head._2
```
akara: OK. All good with the above.

squbs-pattern/src/test/scala/org/squbs/pattern/util/ConfigUtilSpec.scala
```diff
@@ -53,163 +55,206 @@ class ConfigUtilSpec extends FunSpecLike with Matchers {
...
 
-    it ("should throw ConfigMissing exception for non existing config string") {
+    it("should throw ConfigException.Missing exception for non existing config string") {
       intercept[ConfigException.Missing](
         config.get[String]("str")
       )
     }
 
-    it ("should throw IllegalArgumentException for unexpected type of existing config value") {
+    it("should throw IllegalArgumentException for unexpected type of existing config value") {
```
akara: We should ensure both the exception type and the message is correct:

```scala
the [IllegalArgumentException] thrownBy config.get[Unit]("testConfig.str") should
  have message "Configuration option type Unit not implemented"
```

squbs-pattern/src/main/scala/org/squbs/pattern/util/ConfigUtil.scala
```diff
@@ -22,37 +22,46 @@ import com.typesafe.config.{Config, ConfigException, ConfigMemorySize}
...
+      (tag match {
+        case StringTag => underlying.getString(path)
+        case StringListTag => underlying.getStringList(path).toSeq
+        case TypeTag.Int => underlying.getInt(path)
+        case TypeTag.Boolean => underlying.getBoolean(path)
+        case TypeTag.Double => underlying.getDouble(path)
+        case ConfigTag => underlying.getConfig(path)
+        case ConfigListTag => underlying.getConfigList(path).toSeq
+        case FiniteDurationTag => Duration(underlying.getString(path)).asInstanceOf[FiniteDuration]
```
akara: Another "think out loud." Should we ensure the exception type is consistent? In this case this line would be:

```scala
  case FiniteDurationTag =>
    val s = underlying.getString(path)
    try {
      val d = Duration(underlying.getString(path))
      d.asInstanceOf[FiniteDuration]
    } catch {
      case _: NumberFormatException | _: ClassCastException =>
        throw new ConfigException.WrongType(underlying.origin, s"Path: $path, value: $s is not a FiniteDuration.)
    }
```

But then we have to do something similar when parsing Duration. What do you think?

squbs-pattern/src/main/scala/org/squbs/pattern/util/ConfigUtil.scala
```diff
@@ -22,37 +22,46 @@ import com.typesafe.config.{Config, ConfigException, ConfigMemorySize}
...
+        case DurationTag => Duration(underlying.getString(path))
+        case ConfigMemorySizeTag => underlying.getMemorySize(path)
+        case _ => throw new IllegalArgumentException(s"Configuration option type $tag not implemented")
```
akara: Just double-checked. We need to use `${tag.tpe}` and not `$tag`.

```
scala> val t = typeTag[Unit]
t: reflect.runtime.universe.TypeTag[Unit] = TypeTag[Unit]

scala> t.toString
res0: String = TypeTag[Unit]

scala> t.tpe.toString
res1: String = Unit

scala> s"$t"
res2: String = TypeTag[Unit]

scala> s"${t.tpe}"
res3: String = Unit
```

squbs-pattern/src/main/scala/org/squbs/pattern/util/ConfigUtil.scala
dmartyanov: Thank you for feedback, I agree,  completely ignoring exception in `getOption` doesn't look like a good idea,  we should differentiate `ConfigException.Missing` from format exceptions as well. Will fix that.
The problem of injection `Logger` into `ConfigUtil` worries me a little bit. 
Making `object ConfigUtil extends LazyLogging` doesn't look perfect,  because logger's name would not be representative and helpful in localisation of exceptions. Probably, we may consider defining `ConfigUtil` as a trait with `LazyLogging` self annotation and mix it into classes instead of importing. There is a problem in this approach as well, we use either `akka.event.LoggingAdapter` or `com.typesafe.scalalogging.LazyLogging` depends on entity. What do you think about it ?

squbs-pattern/src/test/scala/org/squbs/pattern/util/ConfigUtilSpec.scala
dmartyanov: Many thanks for that!

squbs-pattern/src/main/scala/org/squbs/pattern/util/ConfigUtil.scala
dmartyanov: I used this style no to call `typeTag[T]` twice( `case _ => ...`). Will get it fixed.

squbs-pattern/src/test/scala/org/squbs/pattern/util/ConfigUtilSpec.scala
dmartyanov: Thanks!

squbs-pattern/src/main/scala/org/squbs/pattern/util/ConfigUtil.scala
dmartyanov: To my opinion the recovering of exception with an exception of another type reduces standalone value of `getTry` method. As we are not handling `ConfigException.WrongType` exception as a separate case, this change in fact doesn't have any impact on current codebase. Also we have a separate exception for incorrect `Regex` value,  and probably more cases will appear. The  mapping for all types of exception should not be implemented here.

akara: IMO, it depends on the semantics. `getTry` with varying exception types
would expose users to the underlying implementation and they would have to
deal with the exception of two different types differently. For instance,
if the value is "12345foo", `getTry[ConfigMemorySize]` would return
`Failure(e: ConfigException.wrongType)` whereas `getTry[FiniteDuration]`
would return `Failure(e: NumberFormatException)`. Both of them are actually
parsing exceptions. This is the inconsistency I'm talking about. If the
exception is outside the realm of parsing, we should let its type show
through.

If you really feel this is an overreach into `getTry` and that the user
should know or test that similar parsing exceptions would return you
different exception types depending on which type you're parsing, I'll
respect that.

Thanks!
-Akara

On Sun, Jun 19, 2016 at 10:22 PM, Dmitry Martyanov <notifications@github.com

dmartyanov: I agree I always prefer to bring more order to the logic. 
Only one comment just to keep it in mind: 
If we recover `FiniteDuration` specific exceptions with `ConfigException.WrongType`,  we should do the same with `java.util.regex.PatternSyntaxException` for incorrect Regex string as well. That leads us to the situation when  `getTry` could return one of  4 options: `Success(value)`,  `Failure(IllegalArgumentException)`,  `Failure(ConfigException.Missing)` and `Failure(ConfigException.WrongType)` (like it is in `com.typesafe.config.Config`). 
Looks good to me. The changes have been committed.

squbs-pattern/src/main/scala/org/squbs/pattern/util/ConfigUtil.scala
akara: In that case, on line 41 above, you should probably do:

```scala
def getTry[T: TypeTag](path: String): Try[T] = Try {
```

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/ServiceRegistryBase.scala
```diff
@@ -98,8 +98,8 @@ trait ServiceRegistryBase[A] {
     else config getString "bind-address"
     val port = config getInt "bind-port"
     // assign the localPort only if local-port-header is true
-    val localPort = config getOptionalBoolean "local-port-header" flatMap { useHeader =>
-      if (useHeader) Some(port) else None
+    val localPort = config getOption[Boolean] "local-port-header" collect {
```
akara: I have just looked through the failures and tried to figure out what could be triggering it. The problem turns out to be on this line. Can you please change this to:

```scala
val localPort = config.getOption[Boolean]("local-port-header") collect {
```

Somehow scoverage bails when we use infix notation in combination with type parameters and parameters. Just changing the style fixes the issue.

Issue #226, opened 2016-06-25
PipelineManagerSpec fails sporadically


akara:
Due to PipelineManager's use of Akka Agent, it is eventually getting to the up-to-date state. The test relies on the state to be up-to-date. If this is not happening in a timely manner, the test will fail.

PR #227, opened 2016-06-25
#226: Fix for PipelineManagerSpec sporadic failures. 

akara:
Note that processors became a TrieMap as there is potential concurrent update at startup. This map is not concurrently accessed otherwise at runtime. Ideally, the PipelineManager should be an Actor. But that would be a future step.

### DIFF
--- a/squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
+++ b/squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
@@ -17,18 +17,19 @@
 package org.squbs.pipeline
 
 import akka.actor._
-import akka.agent.Agent
 import com.typesafe.config.{Config, ConfigObject}
 import com.typesafe.scalalogging.LazyLogging
 import org.squbs.pipeline.ConfigHelper._
 
 import scala.collection.JavaConversions._
+import scala.collection.concurrent.TrieMap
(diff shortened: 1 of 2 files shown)


dmartyanov:
LGTM

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineManager.scala
```diff
@@ -53,13 +54,14 @@ case class PipelineManager(configs: Map[String, RawPipelineSetting],
...
-      Option[Either[Option[Processor], PipelineSetting]] = {
-    pipelines().get(name) match {
+  Option[Either[Option[Processor], PipelineSetting]] = {
+    pipelines.get(name) match {
       case None =>
         configs.get(name) match {
           case Some(cfg) =>
             try {
               val factoryInstance = Class.forName(cfg.factoryClass).newInstance()
+              println(s"Creating factory for $name")
```
anilgursel: Would this be a `logger.debug` instead ?

akara: Thanks! @anilgursel for catching this. It was just an artifact left over from debugging the issue. Removed now.

akara:
No new tests or documents needed. This is just a bug fix.

Issue #184, opened 2016-02-25
Automatically managing lifecycle of perpetual streams


akara:
Given a perpetual stream, one that starts when server starts and stops when server stops, we need a way to shutdown **ANY** such stream cleanly, no matter of the materialized value being a `Future` or not.

There is an Akka issue open for this purpose: https://github.com/akka/akka/issues/18612
We need this solve in order to implement clean shutdown.

PR #203, opened 2016-05-06
#184 Added PerpetualStream and complete tests + docs

neovasant:

PR #231, opened 2016-07-01
#184: PerpetualStream implementation, tests, and docs

akara:

- [X] Commits are squashed.
- [X] Tests added.


akara:
This PR supersedes #203.

docs/streams-lifecycle.md
```diff
@@ -0,0 +1,202 @@
+# Streams Lifecycle
+Akka Streams/Reactive stream needs to be integrated with runtime lifecycle of the server. For this, an automated or semi-automated integration is provided through `PerpetualStream`. For fine-grained control over stream sources directly, `LifecycleManaged` provides a wrapping that can control any source component for proper stopping and shutdown so the flow can start/stop gracefully.
```
anilgursel: Should add a link in the first sentence, something like "... with runtime lifecycle of the server (please see [Runtime Lifecycle & API](lifecycle.md)) ."

docs/streams-lifecycle.md
```diff
@@ -0,0 +1,202 @@
...
+    // Do all your cleanup
+    // For safety, call super
+    super.shutdown()
+    // The Future from super.shutdown may not mean anything.
+    // Feel free to create your own future that identifies the
+    // stream being done. Return your Future instead.
+  }
+```
+
+`shutdown` needs to do the followings:
```
anilgursel: followings -> following

docs/streams-lifecycle.md
```diff
@@ -0,0 +1,202 @@
...
+
+```
+
+**Java**
+
+```java
+import static org.squbs.stream.TriggerEvent.DISABLE;
+import static org.squbs.stream.TriggerEvent.ENABLE;
+
+    final Source<?, ?> inSource = <your-original-source>;
```
anilgursel: Indentation.

docs/streams-lifecycle.md
```diff
@@ -0,0 +1,202 @@
...
+
+```
+
+## Custom Lifecycle Event(s) for Trigger
+If you want to respond to more lifecycle events beyond `Active` and `Stopping`, for example you want `Failed` to also stop the flow, you can modify the lifecylce event mapping.
+
+```scala
+import org.squbs.stream.TriggerEvent._
+
+  val inSource = <your-original-source>
```
anilgursel: Indentation.

akara: I think I need to be clear what indentation we want in these cases. The `import` statements are generally placed at the beginning of the file for Scala. For Java, they are always placed at the beginning. The rest is code which is in some class. I can do a `...` between the import and the code to make it clearer. But the code indentation is always further in than the `import` statements, except for the Scala case of inline imports. Please let me know how you want to see it. I've addressed all other comments pending sync. Thanks!

anilgursel: We need to do one of the following:
* add `class Sample {`
* Just indent at the same level with `import` if there is no `class Sample` etc.

I lean towards the second option.  I understand the current indentation has a subtle consideration; however, it looks more of a problem and causes distraction during reading.

akara: Thanks, will go with option 2.

Issue #213, opened 2016-05-22
Refactor `RouteSelectorStage` with a `PartitionStage`


anilgursel:
`PartitionStage` was introduced in Akka 2.4.2.  Instead of maintaining our own graph stage logic, we should utilize it.

PR #236, opened 2016-07-21
#213 Refactor `RouteSelectorStage` with a `PartitionStage`

anilgursel:


squbs-unicomplex/src/main/scala/org/squbs/unicomplex/streaming/FlowHandler.scala
```diff
@@ -88,20 +88,25 @@ class Handler(routes: Agent[Seq[(Path, ActorWrapper, PipelineSetting)]], localPo
...
-      val pathExtractor = (rc: RequestContext) => normPath(rc.request.uri.path)
-      val pathMatcher = (p1: Path, p2: Path) => pathMatch(p1, p2)
 
       val routesMerge = b.add(Merge[RequestContext](actorWrappers.size + 1))
-      val rss = b.add(RouteSelectorStage(paths, pathExtractor, pathMatcher))
+
+      def partitioner(rc: RequestContext) = {
+        val index = paths.indexWhere(p => pathMatch(normPath(rc.request.uri.path), p))
+        // If no match, give the last port's index (which is the NOT_FOUND use case)
+        if(index >= 0) index else paths.size
```
anilgursel: Did not want to do it with `zipWithIndex.find` as the readability is not as clear.  

Another option would be `(index :: paths.size :: Nil).filter(_ >= 0).head`, but still harder to understand.

I with `indexOf` returned an `Option`.

Issue #225, opened 2016-06-20
Timeout policy documentation should be more elaborate


Harikiranvuyyuru:
Timeout policy documentation should clarify that timeout policy doesn't enforce timeouts (results in timeout exceptions) and that it provides only values for timeout

PR #237, opened 2016-07-25
 #225 Timeout policy documentation should be more elaborate

Harikiranvuyyuru:
Modified timeout policy documentation to clarify that usage of timeout policy itself will not cause timeouts to occur, and timeout occurs only when the values derived from timeout policy are used for setting the timeout values.

### DIFF
--- a/docs/timeoutpolicy.md
+++ b/docs/timeoutpolicy.md
@@ -1,6 +1,6 @@
 #Timeout Policy
 
-Timeouts are a key part of every asynchronous system and every distributed system. They are usually configured statically and hard to get right. squbs provides the Timeout Policy facility as part of the squbs-pattern package to help dynamically determine the correct timeout given a policy instead of static setting.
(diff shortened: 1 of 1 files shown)


docs/timeoutpolicy.md
```diff
@@ -1,6 +1,6 @@
 #Timeout Policy
 
-Timeouts are a key part of every asynchronous system and every distributed system. They are usually configured statically and hard to get right. squbs provides the Timeout Policy facility as part of the squbs-pattern package to help dynamically determine the correct timeout given a policy instead of static setting.
+Timeouts are a key part of every asynchronous system and every distributed system. They are usually configured statically and hard to get right. squbs provides the Timeout Policy facility as part of the squbs-pattern package to help dynamically determine the correct timeout given a policy instead of static setting.This dynamically determined timeout value from the policy, can then be utilized for setting the timeouts.Please note that only when the values derived from timeout policy are used for timeout settings, transactions will timeout according to the policy.
```
anilgursel: Would you please add two spaces after full stops.  

Not sure if we need a comma after "This dynamically determined timeout value from the policy,".

The last sentence is not very clear to me :(

PR #239, opened 2016-07-26
#238 adds support for user defined akka stage - broadcast buffer with comm…

SarathChandran:


docs/persistent-buffer.md
```diff
@@ -151,6 +151,51 @@ implicit val serializer = new PersonSerializer()
 val buffer = new PersistentBuffer[Person](new File("/tmp/myqueue")
 ```
 
+##Broadcast Buffer
+`BroadcastBuffer` is a variant of persistent buffer. This works similar to `PersistentBuffer` except that the output is broadcasted to multiple subscribers. Hence it is a combination of buffer and broadcast stages. The configuration takes additional parameter named `subscribers` which specifies the number of out ports.
```
akara: Just to keep the semantics consistent with `Broadcast`, you may want to name the parameter `outputPorts`. Also we should note that each of the output ports run independently from the input port and other output ports.

docs/persistent-buffer.md
```diff
@@ -151,6 +151,51 @@ implicit val serializer = new PersonSerializer()
...
+```scala
+implicit val serializer = QueueSerializer[ByteString]()
+
+val in = Source(1 to 100000)
+val flowCounter = Flow[Any].map(_ => 1L).reduce(_ + _).toMat(Sink.head)(Keep.right)
+
+def committer(buffer: BroadcastBuffer[T]) = Flow[Event[T]].map(element => {
+      buffer.commit(element.subscriberId, element.commitOffset)
+      element
+    })
```
akara:
```scala
def committer(buffer: BroadcastBuffer[T]) = Flow[Event[T]].map { element =>
  buffer.commit(element.subscriberId, element.commitOffset)
  element
}
```

akara: Another "thinking out loud" whether we should have the committer be part of the API instead of defined in user code. For that, you need to add an API to BroadcastBuffer:

```scala
def committer = Flow[Event[T]].map { element =>
  commit(element.subscriberId, element.commitOffset)
  element
}
```

So the user code will just be:

```scala
  val buffer = new BroadcastBuffer[ByteString](config)
  val committer = buffer.committer
  val mr = builder.add(merge)
```

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -0,0 +1,118 @@
...
+    setHandler(in, inHandler)
+
+    out.zipWithIndex foreach { case (currentOut, subscriberId) =>
+      setHandler(currentOut, outHandler(currentOut, subscriberId))
+    }
+
+    override def preStart(): Unit = {
+      // Start upstream demand
+      pull(in)
+    }
```
akara: Here is where you can add your `committer` operation.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -0,0 +1,118 @@
...
+        * emits the remaining elements from the buffer
+        * at subscriber level in out port.
+        */
+      private def flushStage() = {
+        out.zipWithIndex foreach { case (currentOut, subscriberId) =>
+          var element: Option[Entry[T]] = queue.dequeue(subscriberId)
+          while (element.isDefined) {
+            element foreach { e =>
+              emitCount(subscriberId).incrementAndGet()
+              emit[Event[T]](currentOut, Event(subscriberId, e.index, e.entry))
```
akara: Using `emit` is dangerous here. We need to re-design this without emit.

akara: This is called only when upstream finishes, which means it may got killed. You may not want to keep emitting events in this loop. Rather, you want to stop the stream. When it restarts, it will pickup. This is the main reason why the stream is persistent.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -85,25 +95,66 @@ class PersistentQueue[T](config: QueueConfig)(implicit val serializer: QueueSeri
...
+  /**
+    * Commits the queue's index, this index is mounted when
+    * the queue is initialized next time
+    *
+    * @param subscriberId
+    * @param index to be committed for next read
+    */
+  def commit(subscriberId: Int, index: Long) = {
+    if (!indexMounted) mountIndexFile()
+    indexStore.writeLong(subscriberId.toLong << 8, index)
```
akara: Is it really `<< 8`? Subscriber id 0 will be offset 0, 1 will be offset 256, 2 will be offset 512. Since each offset contains a `Long`, would it be better to do a `* 8` on the id? If you want to use bit shifting, should it rather be `<< 3`?

squbs-pattern/src/main/scala/org/squbs/pattern/stream/QueueConfig.scala
```diff
@@ -39,7 +40,8 @@ object QueueConfig {
     val blockSize = config.getOption[ConfigMemorySize]("block-size") map (_.toBytes) getOrElse defaultBlockSize
     val indexSpacing = config.getOption[ConfigMemorySize]("index-spacing") map (_.toBytes.toInt) getOrElse cycle.defaultIndexSpacing
     val indexCount = config.get[Int]("index-count", cycle.defaultIndexCount)
-    QueueConfig(persistDir, cycle, wireType, blockSize, indexSpacing, indexCount)
+    val subscribers = config.get[Int]("subscribers", defaultSubscribers)
```
akara: Should we name this `outputPorts` in code and `output-ports` in config instead?

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -85,25 +95,66 @@ class PersistentQueue[T](config: QueueConfig)(implicit val serializer: QueueSeri
...
+  def clearStorage(): Unit = {
+    val reader = queue.createTailer()
+    reader.moveToIndex(minIndex())
+    val current = cache.resourceFor(reader.cycle)
+
+    var previous = cache.resourceFor(reader.toStart.cycle)
+    while (!current.path.equals(previous.path)) {
+      previous = cache.resourceFor(reader.toStart.cycle)
+      previous.path.delete()
+    }
```
akara: While loops are always fishy. Try using tail recursion.

squbs-pattern/src/test/scala/org/squbs/pattern/stream/BroadcastBufferSpec.scala
```diff
@@ -0,0 +1,423 @@
...
+      Option(file.listFiles).map(_.toList).getOrElse(Nil).foreach(delete(_))
+    file.delete
+  }
+
+  def createElement(n: Int): T
+
+  def format(element: T): String
+
+  class QAUtil {
+    var tempPath: File = Files.createTempDirectory("persistent_queue").toFile
```
akara: Does this really need to be mutable?

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentBufferSpec.scala
```diff
@@ -103,26 +106,36 @@ abstract class PersistentBufferSpec[T: ClassTag, Q <: QueueSerializer[T]: Manife
...
     }.toMat(Sink.head)(Keep.right)
+    val total = Flow[Any].map( _ => 1L).reduce(_ + _).toMat(Sink.head)(Keep.right)
+    def committer(buffer: PersistentBuffer[T]) = Flow[Event[T]].map(element => {
+      buffer.commit(element.subscriberId, element.commitOffset)
+      element
+    })
 
-    val streamGraph = RunnableGraph.fromGraph(GraphDSL.create(counter(t1 = _)) { implicit builder =>
-      sink =>
+    val streamGraph = RunnableGraph.fromGraph(GraphDSL.create(counter(t1 = _), total)((a,b) => (a,b)) { implicit builder =>
```
akara: Easy form `((_, _))` instead of ((a, b) => (a, b)).

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentBufferSpec.scala
```diff
@@ -259,20 +285,24 @@ abstract class PersistentBufferSpec[T: ClassTag, Q <: QueueSerializer[T]: Manife
...
     val injectError = Flow[Int].map { n =>
       if (n == failTestAt) throw new NumberFormatException("This is a fake exception")
       else n
     }
 
     val buffer = new PersistentBuffer[T](tempPath.toFile)
+    def committer(buffer: PersistentBuffer[T]) = Flow[Event[T]].map(element => {
+      buffer.commit(element.subscriberId, element.commitOffset)
+      element
+    })
```
akara: Yeah, you can cut down a ton of code, by making the committer part of the API.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
akara: One more, really most important thing of all: For a given subscriber/port, if previous index does not commit and a subsequent one commits, this case it will override the previous non-commit. That means that message can be lost even thought it is not committed. We need to discuss how to handle such scenarios.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
SarathChandran: Will this be good enough?
```scala
def clearStorage(): Unit = {
    val reader = queue.createTailer(); reader.moveToIndex(minIndex())
    val current = cache.resourceFor(reader.cycle)
    def first = cache.resourceFor(reader.toStart.cycle)

    @tailrec
    def delete(previous: Resource) : Unit = {
      if (!current.path.equals(previous.path)) {
        previous.path.delete()
        delete(first)
      }
    }
    delete(first)
  }
```

squbs-pattern/src/main/java/org/squbs/pattern/stream/QueueResource.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package org.squbs.pattern.stream;
+
+import java.io.File;
+
+import net.openhft.chronicle.queue.impl.RollingResourcesCache;
+import net.openhft.chronicle.queue.impl.single.SingleChronicleQueue;
+
+public class QueueResource {
+    public static RollingResourcesCache resourceCache(SingleChronicleQueue queue) {
```
SarathChandran: The main purpose of this java file is because of Java 8 Functional interface. Can this be written in scala without adding dependencies?

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
SarathChandran: @akara , When i was actually testing with multiple outputPorts, when one of the ports completed, the stage completed leaving the remaining ports elements unpushed. I was wondering why despite using a boolean seq for each outputPorts previous, then i decided to go with emit. I need your help here.

docs/persistent-buffer.md
```diff
@@ -151,6 +151,47 @@ implicit val serializer = new PersonSerializer()
 val buffer = new PersistentBuffer[Person](new File("/tmp/myqueue")
 ```
 
+##Broadcast Buffer
+`BroadcastBuffer` is a variant of persistent buffer. This works similar to `PersistentBuffer` except that the output is broadcasted to multiple output ports. Hence it is a combination of buffer and broadcast stages. The configuration takes additional parameter named `output-ports` which specifies the number of output ports.
```
anilgursel: ... takes an additional parameter ...

anilgursel: I think we need some explanation here how it is different than a usual

```scala
persistenBuffer ~> broadcast ~> x
                   broadcast ~> y
                   broadcast ~> z
```

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -0,0 +1,128 @@
...
+              queue.close()
+              completeStage()
+              flushStage()
+            }
+          case Some(element) =>
+            push(outlet, Event(outputPortId, element.index, element.entry))
+        }
+      }
+
+      /**
```
anilgursel: No need to use javadoc style for a `private` method.  Single star would suffice.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -0,0 +1,128 @@
...
+    out.zipWithIndex foreach { case (currentOut, outputPortId) =>
+      setHandler(currentOut, outHandler(currentOut, outputPortId))
+    }
+
+    override def preStart(): Unit = {
+      // Start upstream demand
+      pull(in)
+    }
+
+    def detect() = Flow[Event[T]].map { element =>
```
anilgursel: Not sure where `detect` is used?  Could not see anything in the doc nor any code usage.

SarathChandran: @anilgursel this method would be removed, this was just added to discuss the usage with akara. sorry about that.

squbs-pattern/src/test/scala/org/squbs/pattern/stream/BroadcastBufferSpec.scala
```diff
@@ -0,0 +1,425 @@
...
+  }
+
+  it should "buffer a stream of one million elements using GraphDSL and custom config" in {
+    val util = new QAUtil
+    import util._
+
+    val streamGraph = RunnableGraph.fromGraph(GraphDSL.create(flowCounter) { implicit builder =>
+      sink =>
+        import GraphDSL.Implicits._
+        val buffer = newBuffer
```
anilgursel: I believe having a `val config` at the top and directly calling `new BroadcastBuffer(...)` would read easier

docs/persistent-buffer.md
```diff
@@ -152,7 +154,9 @@ val buffer = new PersistentBuffer[Person](new File("/tmp/myqueue")
 ```
 
 ##Broadcast Buffer
-`BroadcastBuffer` is a variant of persistent buffer. This works similar to `PersistentBuffer` except that the output is broadcasted to multiple subscribers. Hence it is a combination of buffer and broadcast stages. The configuration takes additional parameter named `subscribers` which specifies the number of out ports.
+`BroadcastBuffer` is a variant of persistent buffer. This works similar to `PersistentBuffer` except that the output is broadcasted to multiple output ports. Hence it is a combination of buffer and broadcast stages. The configuration takes an additional parameter named `output-ports` which specifies the number of output ports.
+
+A broadcast buffer is specially required when a commit guarantee has to be ensured for each individual output port. In addition to this, each output port can publish messages at an independent phase proportional to the speed of the downstream consumer.
```
SarathChandran: @anilgursel does it look better with these requirements

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -31,56 +32,55 @@ class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T])
...
         }
       }
 
-      /**
-        * emits the remaining elements from the buffer
-        * at subscriber level in out port.
-        */
+      // emits the remaining elements from the buffer
+      // at output port level.
       private def flushStage() = {
```
SarathChandran: @az-qbradley @anilgursel , this is the place where @akara mentioned to do away with usage of emit. One thing is clear that, after restart, all output ports will resume from the place it left its messages. However, for ensuring the total counts the records have processed in test specs, this part of code was added. It ensures all output ports have finished processing all the records from source.

could you look through alternative ways to get this done

squbs-pattern/src/test/scala/org/squbs/pattern/stream/BroadcastBufferSpec.scala
```diff
@@ -298,45 +301,47 @@ abstract class BroadcastBufferSpec[T: ClassTag, Q <: QueueSerializer[T] : Manife
       else n
```
SarathChandran: @anilgursel @az-qbradley While running tests after adding a short fix "(i suspect that concurrency issues still exists)" 

https://github.com/paypal/squbs/pull/239/commits/5b10dc01043dd0cc838cb34e007109123422e97e#diff-717b79f6434ac986220216a2e1c86d40R119

```scala
if (!indexMounted || indexStore.refCount() <= 0) mountIndexFile()
```

This particular test case takes a long time to run. Count you please check if something is going wrong here. Because prior to moving the commit's impl within buffer it ran without any issues.

docs/persistent-buffer.md
anilgursel: Thanks!  It clarifies more.  A minor comment if you agree as well:

"In addition to this, elements can be emit from each output port at an independent rate depending on the downstream demand."

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -0,0 +1,121 @@
...
+import java.util.concurrent.atomic.AtomicInteger
+
+import akka.stream._
+import akka.stream.scaladsl.Flow
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+import com.typesafe.config.Config
+import com.typesafe.scalalogging.Logger
+import org.slf4j.LoggerFactory
+
+class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T])
```
anilgursel: Sorry, I missed this yesterday..  Can you please add Javadoc here, with the style that Akka Stream stages have, e.g., '''Emits when'''.  See https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala#L31 as an example.

Apparently, we missed it in `PersistentBuffer` as well.  I will take care of it.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentBuffer.scala
```diff
@@ -82,4 +84,11 @@ class PersistentBuffer[T] private(private[stream] val queue: PersistentQueue[T])
       }
     })
   }
+
+  def commit() = Flow[Event[T]].map { element =>
```
anilgursel: This changes the API for `PersistenBuffer`.  If we would like to provide this option, we should also have a default auto-commit option.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -85,25 +98,66 @@ class PersistentQueue[T](config: QueueConfig)(implicit val serializer: QueueSeri
...
-  def dequeue: Option[T] = {
-    var output: Option[T] = None
-    if (reader.readDocument(new ReadMarshallable {
-      override def readMarshallable(wire: WireIn): Unit = output = serializer.readElement(wire)
-    })) {
-      if (!indexMounted) mountIndexFile()
-      indexStore.writeLong(0L, reader.index)
-      output
-    } else None
+  def dequeue(outputPortId: Int = 0): Option[Entry[T]] = {
```
anilgursel: My understanding is that the index is not auto incremented anymore..  If so, how do we make sure some messages are not processed multiple times?  Let's assume we have the graph `a ~> persistentBuffer ~> b ~> c ~> commit`, until the message reaches to the `commit` stage, b might demand more elements, in such scenario, will `PersistentBuffer` emit the last element again ?

SarathChandran: @anilgursel The idea of `commit` stage itself is to re-process already emitted elements if the elements are not committed due to some unexpected failure. Hence it is expected that if the message does not reach the `commit` stage messages should be reprocessed. (P.S This intent should be included in documentation, i will add that quickly). Also providing an auto-commit as you said, will solve the case of not sending the same failed messages more than once, (because messages can be committed right after it is emitted, thinking of providing this in config). Hope i am getting it correct here.

anilgursel: I understand the failure scenarios.  However, for the success scenarios, we should not cause a message to be emitted multiple times.  I am not sure how we guarantee this with the current implementation.  Auto-commit would address that; however, once that's off, we still have the problem.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -0,0 +1,116 @@
...
+                completeStage()
+              }
+            }
+          case Some(element) =>
+            push(outlet, Event(outputPortId, element.index, element.entry))
+        }
+      }
+    }
+
+    def inHandler() = new InHandler {
```
akara: No state change, no need for `()`

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -18,10 +18,18 @@ package org.squbs.pattern.stream
...
 import net.openhft.chronicle.core.OS
 import net.openhft.chronicle.queue.ChronicleQueueBuilder
+import net.openhft.chronicle.queue.impl.RollingResourcesCache.Resource
 import net.openhft.chronicle.wire.{ReadMarshallable, WireIn, WireOut, WriteMarshallable}
+import org.slf4j.LoggerFactory
+
+import scala.annotation.tailrec
+
+case class Entry[T](index: Long, entry: T)
+case class Event[T](outputPortId: Int, commitOffset: Long, entry: T)
```
akara: `Event` has no use in `PersistentQueue.scala`. We should move it to `PersistentBuffer.scala`

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -0,0 +1,116 @@
...
+      override def onUpstreamFailure(ex: Throwable): Unit = {
+        val logger = Logger(LoggerFactory.getLogger(this.getClass))
+        logger.error("Received upstream failure signal: " + ex)
+        queue.close()
+        completeStage()
+      }
+    }
+  }
+
+  def commit() = Flow[Event[T]].map { element =>
```
akara: `commit` can be a `val`.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -56,6 +67,9 @@ class PersistentQueue[T](config: QueueConfig)(implicit val serializer: QueueSeri
   private var indexFile: IndexFile = _
   private var indexStore: MappedBytesStore = _
 
+  val totalOutputPorts = outputPorts
+  val doCommit = !autoCommit
```
akara: do we really need a `doCommit` just as an inversion of `autoCommit`? We could make `config` a val reference any `queue.doCommit` with `!queue.config.autoConfig` instead.

akara: Actually, we're violating encapsulation here. The fact that the queue is autoCommit or not should be handled by the queue. It is not the buffer's job to know. So we should be handling it in a commit internal to queue. You may do:

```scala
def commit(): Unit = if (!autoCommit) internalCommit

def internalCommit(): Unit = { // The current commit logic }
```

Then you can just blindly call commit() from outside.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -64,8 +78,12 @@ class PersistentQueue[T](config: QueueConfig)(implicit val serializer: QueueSeri
 
   if (path.isFile) {
     mountIndexFile()
-    val startIdx = indexStore.readLong(0L)
-    reader.moveToIndex(startIdx)
+    0 until outputPorts foreach { outputPortId =>
+      val startIdx = read(outputPortId)
+      logger.info(s"Setting idx for outputPort ${outputPortId} - " + startIdx)
```
akara: `      logger.info("Setting idx for outputPort {} - {}", outputPortId, startIdx)`

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -85,25 +103,69 @@ class PersistentQueue[T](config: QueueConfig)(implicit val serializer: QueueSeri
...
+          }
+        }
+    })) output else None
+  }
+
+  /**
+    * Commits the queue's index, this index is mounted when
+    * the queue is initialized next time
+    *
+    * @param outputPortId
```
akara: Please observe IntelliJ warnings. We need a description behind this.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -85,25 +103,69 @@ class PersistentQueue[T](config: QueueConfig)(implicit val serializer: QueueSeri
...
+    indexStore.writeLong(outputPortId.toLong << 3, index)
   }
 
+  // Reads the given outputPort's queue index
+  private def read(outputPortId: Int) : Long = {
+    if (!indexMounted) mountIndexFile()
+    indexStore.readLong(outputPortId.toLong << 3)
+  }
+
+  private def minIndex() = Vector.tabulate(outputPorts)(outputPortId => read(outputPortId)).min
```
akara: `Iterator.tabulate(...).min` will be much more efficient. It is lazy and does not need to allocate space and keep the indexes around. A `min` call is a one-pass call anyway.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -85,25 +103,69 @@ class PersistentQueue[T](config: QueueConfig)(implicit val serializer: QueueSeri
...
+    * min read index from all outputPorts
+    */
+  def clearStorage(): Unit = {
+    val reader = queue.createTailer(); reader.moveToIndex(minIndex())
+    val current = cache.resourceFor(reader.cycle)
+    def first() = cache.resourceFor(reader.toStart.cycle)
+
+    @tailrec
+    def delete(previous: Resource) : Unit = {
+      if (!current.path.equals(previous.path)) {
```
akara: Just because this is Scala code, we use `==` instead of `equals` which has the same meaning (unlike Java)

```scala
if (current.path != previous.path) {
```

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -85,25 +103,69 @@ class PersistentQueue[T](config: QueueConfig)(implicit val serializer: QueueSeri
...
+  /**
+    * Commits the queue's index, this index is mounted when
+    * the queue is initialized next time
+    *
+    * @param outputPortId
+    * @param index to be committed for next read
+    */
+  def commit(outputPortId: Int, index: Long) = {
+    if (!indexMounted) mountIndexFile()
+    indexStore.writeLong(outputPortId.toLong << 3, index)
```
akara: The `toLong` is probably excessive here.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -85,25 +103,69 @@ class PersistentQueue[T](config: QueueConfig)(implicit val serializer: QueueSeri
...
   }
 
+  // Reads the given outputPort's queue index
+  private def read(outputPortId: Int) : Long = {
+    if (!indexMounted) mountIndexFile()
+    indexStore.readLong(outputPortId.toLong << 3)
```
akara: Ditto.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -0,0 +1,116 @@
...
+      override def onPush(): Unit = {
+        val element = grab(in)
+        queue.enqueue(element)
+        0 until outputPorts foreach { outputPortId =>
+          if (isAvailable(out(outputPortId))) {
+            queue.dequeue(outputPortId) foreach { element =>
+              push(out(outputPortId), Event(outputPortId, element.index, element.entry))
+            }
+          }
+        }
```
akara: Not sure I'm too picky here :-) If so, please ignore:

We reference `out(outputPortId)` multiple times. Not too bad for a vector, but still. This may be a better solve:

```scala
          if (isAvailable(port)) {
            queue.dequeue(id) foreach { element =>
              push(out(id), Event(id, element.index, element.entry))
            }
          }
        }
```

The reason I pull an iterator is to make it lazy and one-use as this `onPush` is called extremely frequently. It will be in the performance critical path.

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentBufferSpec.scala
```diff
@@ -15,131 +15,112 @@
...
+                           bc1 ~> counter(t2 = _)
         ClosedShape
     })
-    val countF = streamGraph.run()
+    val (countF, totalF) = streamGraph.run()
     val count = Await.result(countF, awaitMax)
+    val totalProcessed = Await.result(totalF, awaitMax)
 
     println("Time difference (ms): " + (t1 - t2) / 1000000d)
+    println(s"Total count ${count} vs total processed ${totalProcessed}")
```
akara: `count` and `totalProcessed` are variables, not expressions. No curly braces needed for eval.

```scala
println(s"Total count $count vs total processed $totalProcessed")
```

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentBufferSpec.scala
```diff
@@ -148,160 +129,122 @@ abstract class PersistentBufferSpec[T: ClassTag, Q <: QueueSerializer[T]: Manife
...
-      })
-    val (doneF, firstF) = graph2.run()(ActorMaterializer())
-    Await.ready(doneF, awaitMax)
-    val firstRec = Await.result(firstF, awaitMax)
-    println("First record after recovery: " + format(firstRec))
-    val recordsLost = elementCount + elementsAfterFail - recordCount.get
-    println(s"Number of records lost due to unexpected shutdown: $recordsLost")
-    recordsLost should be < 3
-    delete(tempPath.toFile)
+    resumeGraphAndDoAssertion(recordCount.get, (elementCount + 1))
```
akara: Remove unnecessary parentheses around `elementCount + 1`

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentBufferSpec.scala
```diff
@@ -148,160 +129,122 @@ abstract class PersistentBufferSpec[T: ClassTag, Q <: QueueSerializer[T]: Manife
...
-    discrepancy should be < 3L
-
-    val recordsLost = inCount.get + elementsAfterFail - outCount.get
-    println(s"Number of records lost due to unexpected shutdown: $recordsLost")
-    recordsLost should be < 3
-    delete(tempPath.toFile)
+    val countF = graph.run()(mat)
+    Try { Await.result(countF, awaitMax) }
+    val restartFrom = if (inCounter.get == elementCount) elementCount + 1 else inCounter.get
+    println(s"Restart from count ${restartFrom}")
```
akara: Ditto here `$restartFrom` instead of `${restartFrom}`

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentBufferSpec.scala
```diff
@@ -148,160 +129,122 @@ abstract class PersistentBufferSpec[T: ClassTag, Q <: QueueSerializer[T]: Manife
...
+  }
+
+  private def assertions(beforeShutDown: Long, afterRecovery: Long, totalRecords: Long) = {
+    println(s"Last record processed before shutdown => ${beforeShutDown}")
+    println(s"Records processed after recovery => ${afterRecovery}")
+    val processedRecords = beforeShutDown + afterRecovery
+    val lostRecords = totalRecords - processedRecords
+    println(s"Total records lost due to unexpected shutdown => ${lostRecords}")
+    println(s"Total records processed => ${processedRecords}")
+    if (!autoCommit) processedRecords should be >= totalRecords
```
akara: This whole block has a lot of unnecessary wrapped expressions although they are just variables.

squbs-pattern/src/test/scala/org/squbs/pattern/stream/StreamSpecUtil.scala
```diff
@@ -0,0 +1,73 @@
+package org.squbs.pattern.stream
```
akara: Please add standard header.

akara:
Just a thought. What would be the use of fetching an `Entry[T]` in case `autoCommit = true`? You want an entry because you can use it to commit back. If you do not commit back, would it be better to fetch a `T` and not an `Entry[T]` as there is no use of the `Entry[T]` container anyway? @SarathChandran, @anilgursel, @az-qbradley please let me know your thoughts.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
SarathChandran: @akara this function is from `GraphStageLogic` I am wondering if we can change the definition. When testing with `val inHandler` it says `no handler defined in stage`, but removing the function brackets with def worked though. Should i change here?

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
SarathChandran: `BroadcastBuffer.scala` also has made use of `Event`. Since `PersistentQueue` is the common api for both, I thought this would be correct.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
SarathChandran: did this change, but just curious to know how `0 until outputPorts` and `out.iterator.zipWithIndex` gets evaluated. really interesting to know that there is a difference in this two :)

squbs-pattern/src/test/scala/org/squbs/pattern/stream/StreamSpecUtil.scala
```diff
@@ -0,0 +1,73 @@
...
+  def incrementFlow(counter: AtomicInteger) = Flow[Event[T]].map(event => {
+    counter.incrementAndGet()
+    event
+  })
+
+  def clean() = if (tempPath.isFile) delete(tempPath)
+
+  private def delete(file: File) {
+    if (file.isDirectory)
+      Option(file.listFiles).map(_.toList).getOrElse(Nil).foreach(delete(_))
```
akara:
```scala
Option(file.listFiles).map(_.toList).getOrElse(Nil).foreach(delete)
```

squbs-pattern/src/test/scala/org/squbs/pattern/stream/StreamSpecUtil.scala
```diff
@@ -0,0 +1,73 @@
...
+  lazy val flowCounter = Flow[Any].map(_ => 1L).reduce(_ + _).toMat(Sink.head)(Keep.right)
+  lazy val merge = Merge[Event[T]](outputPorts)
+  lazy val throttle = Flow[Event[T]].throttle(flowRate, flowUnit, burstSize, ThrottleMode.shaping)
+  lazy val head = Sink.head[Event[T]]
+  lazy val last = Sink.last[Event[T]]
+
+  def incrementFlow(counter: AtomicInteger) = Flow[Event[T]].map(event => {
+    counter.incrementAndGet()
+    event
+  })
```
akara:
```scala
  def incrementFlow(counter: AtomicInteger) = Flow[Event[T]].map {event =>
    counter.incrementAndGet()
    event
  }
```

squbs-pattern/src/test/scala/org/squbs/pattern/stream/BroadcastBufferSpec.scala
```diff
@@ -0,0 +1,428 @@
...
+                                 bcBuffer ~> throttle ~> commit ~> incrementFlow(atomicCounter(1)) ~> count2
+                           bc ~> fireFinished()
+
+        ClosedShape
+    })
+    graph.run()(mat)
+    Await.result(shutdownF, awaitMax)
+    Thread.sleep(1000) // Wait a sec to make sure the shutdown completed. shutdownF only says shutdown started.
+    val beforeShutDown = SinkCounts(atomicCounter(0).get, atomicCounter(1).get)
+    resumeGraphAndDoAssertion(beforeShutDown, (elementCount + 1))
```
akara: `elementCount + 1` does not need to be in parentheses. Please remove parentheses.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
akara: @anilgursel: Once and only once is almost impossible to attain - unless you use strict transactional semantics. So we should aim for at least once or at most once semantics. Auto commit achieves at least once semantics. Manual commit achieves at most once semantics where stream idempotence has to be observed. Thanks!

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentQueueSpec.scala
```diff
@@ -22,50 +22,83 @@ import akka.util.ByteString
...
-class PersistentQueueSpec extends FlatSpec with Matchers with BeforeAndAfterAll {
+class PersistentQueueSpec extends FlatSpec with Matchers {
 
-  val tempPath = Files.createTempDirectory("persistent_queue")
   implicit val serializer = QueueSerializer[ByteString]()
-  val queue = new PersistentQueue[ByteString](QueueConfig(tempPath.toFile))
 
   def delete(file: File) {
     if (file.isDirectory)
       Option(file.listFiles).map(_.toList).getOrElse(Nil).foreach(delete(_))
```
akara: Ditto, `delete(_)` can just be `delete`

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
akara: Let me check this out.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
akara: OK - lack of a better place :-)

Issue #78, opened 2015-09-17
Remove hardcoded header logging in spray client


anilgursel:
https://github.com/paypal/squbs/blob/master/squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpClientManager.scala#L77

If there are other places, those should be masked as well.

anilgursel:
We will be addressing optional header logging for Akka Http client in 0.9.0 with https://github.com/paypal/squbs/issues/247.

anilgursel:
We are removing it just in case it may contain potential sensitive information.

PR #248, opened 2016-08-24
#78 Remove hardcoded header logging in spray client

anilgursel:


akara:
Code change looks good. I just lost track of the background why we are removing this. Can you add a quick comment to the issue? Thanks!

anilgursel:
Done!

PR #249, opened 2016-08-24
#245 Upgrade to Akka 2.4.9

anilgursel:


akara:
Nice work! Put an async boundary around PersistentBuffer! Code is all good. Will merge once build passes.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -39,36 +39,43 @@ import org.slf4j.LoggerFactory
   * to enable this, set the `auto-commit` to `false` and add a commit stage after downstream consumer.
   *
   */
-class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T])(implicit serializer: QueueSerializer[T])
+class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T],
+                                 onPushCallback: () => Unit = (() => {}))(implicit serializer: QueueSerializer[T])
```
akara: Does IntelliJ raise a warning here? I think we have too many layers of parentheses.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -39,36 +39,43 @@ import org.slf4j.LoggerFactory
...
 
     private var upstreamFinished = false
-    private val finished = Array.fill[Boolean](outputPorts)(false)
 
     override def preStart(): Unit = pull(in)
 
     def outHandler(outlet: Outlet[Event[T]], outputPortId: Int) = new OutHandler {
       override def onPull(): Unit = {
-        queue.dequeue(outputPortId) match {
+        queue.dequeue(outputPortId, finished(outputPortId)) match {
```
akara: Marker comment: Check the encapsulation of `queue`.
Will update/delete this comment as we get further into the review.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -39,36 +39,43 @@ import org.slf4j.LoggerFactory
...
+              finished = finished.updated(outputPortId, true)
               if(finished.reduce(_ && _)) {
-                queue.close()
                 completeStage()
               }
             }
-          case Some(element) => push(outlet, Event(outputPortId, element.index, element.entry))
+          case Some(element) =>
+            push(outlet, Event(outputPortId, element.index, element.entry))
+            queue.lastPushedIndex(outputPortId) = element.index
```
akara: This updates the queue state from outside the queue. Is this the right path?

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
anilgursel: You talking about `onPushCallback: () => Unit = (() => {})` ?  No, it does not give any warning.  I purposefully put an extra parenthesis, thinking it reads better.  But, I can remove it.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
anilgursel: I do not like this at all.  Actually, I can change the design little bit on this..  I can instead keep `lastPushedIndex` in `BroadcastBuffer` and pass a boolean flag `lastPushedIndex(outputPortId) == index` as a parameter to `queue.commit`.

anilgursel: But, if I move `lastPushedIndex` to `BroadcastBuffer`, I need to mark it `@volatile`, and unfortunately, there will be too many writes, so it would be costly..  I could do it only after `upstreamFinished` turns `true`, but, still it would be quite many writes..

anilgursel: Thought about `agent`s, but would not help much here.  One thing I can do..  I can have an actor and send all pushed (outpordId, index) tuple to the actor...  Also, send all commit index and outpordId from commit function as well (after stream finishes, or for a specific port queue is done)..  So, the actor would know all push/commit indexes for each outport..  Each time we send a commit index for an outport id, we check if commit index reached to the pushed index for each outport, and then close the queue.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -77,9 +84,13 @@ class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T])(
       override def onPush(): Unit = {
         val element = grab(in)
         queue.enqueue(element)
+        onPushCallback()
         out.iterator.zipWithIndex foreach { case (port, id) =>
           if (isAvailable(port))
-            queue.dequeue(id) foreach { element => push(out(id), Event(id, element.index, element.entry)) }
+            queue.dequeue(id, finished(id)) foreach { element =>
+              push(out(id), Event(id, element.index, element.entry))
+              queue.lastPushedIndex(id) = element.index
```
akara: Should we even do this all the time? Or should we only do the update when `upstreamFinished`? This line causes an assignment to a volatile and that looks costlier than reads.

Anyway, if we want to redo it to be more async, we can as well accept this.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -39,36 +39,43 @@ import org.slf4j.LoggerFactory
...
+  def withOnPushCallback(onPushCallback: () => Unit) = new BroadcastBuffer[T](queue, onPushCallback)
+
+  def withOnCommitCallback(onCommitCallback: Int => Unit) = new BroadcastBuffer[T](queue.withOnCommitCallback(onCommitCallback), onPushCallback)
 
   private[stream] val outputPorts = queue.totalOutputPorts
   private[stream] val in = Inlet[T]("BroadcastBuffer.in")
   private[stream] val out = Vector.tabulate(outputPorts)(i ⇒ Outlet[Event[T]]("BroadcastBuffer.out" + i))
   val shape: UniformFanOutShape[T, Event[T]] = UniformFanOutShape(in, out: _*)
+  @volatile private var finished = IndexedSeq.fill[Boolean](outputPorts)(false)
+  @volatile private var upstreamFailed = false
```
akara: Is there any difference in terms of upstreamFinished, upstreamFailed, or downstreamFinished? Can we use one `@volatile` variable like `closing` for all use cases?

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentBuffer.scala
```diff
@@ -39,22 +39,29 @@ import org.slf4j.LoggerFactory
   * to enable this, set the `auto-commit` to `false` and add a commit stage after downstream consumer.
   *
   */
-class PersistentBuffer[T] private(private[stream] val queue: PersistentQueue[T])
+class PersistentBuffer[T] private(private[stream] val queue: PersistentQueue[T],
+                                  onPushCallback: () => Unit = (() => {}))
```
akara: Here too, can we get rid of a layer of parentheses?

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentBuffer.scala
```diff
@@ -39,22 +39,29 @@ import org.slf4j.LoggerFactory
...
+  def withOnPushCallback(onPushCallback: () => Unit) = new PersistentBuffer[T](queue, onPushCallback)
+
+  def withOnCommitCallback(onCommitCallback: () => Unit) = new PersistentBuffer[T](queue.withOnCommitCallback(i => onCommitCallback()), onPushCallback)
+
   private[stream] val in = Inlet[T]("PersistentBuffer.in")
   private[stream] val out = Outlet[Event[T]]("PersistentBuffer.out")
   val shape: FlowShape[T, Event[T]] = FlowShape.of(in, out)
   val defaultOutputPort = 0
+  @volatile private var reachedEndOfQueue = false
+  @volatile private var upstreamFailed = false
```
akara: Ditto

akara:
Having all these comments given, I'm happy to merge this PR and address those issues with the next PR. Please just let me know.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
anilgursel: Doing this even after `upstreamFinished` is costly, it would still be quite many time.  I will address this by making queue close async.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
anilgursel: There is.  In case of `upstreamFailed` we should not close the queue in commit phase as it would be closed in `upstreamFailed` hook of `BroadcastBuffer`.

anilgursel:
Let's merge this version.  I do not want to drag it any further.  I will create another issue and make the queue `close` async that should address most of the comments.  I'll do it on Monday.

PR #265, opened 2016-09-21
#260 Remove synchronous close in PersistentQueue

anilgursel:

### DIFF
--- a/squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
+++ b/squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
@@ -17,6 +17,7 @@ package org.squbs.pattern.stream
 
 import java.io.File
 
+import akka.actor.{Props, ActorSystem}
 import akka.stream._
 import akka.stream.scaladsl.Flow
 import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
@@ -40,42 +41,51 @@ import org.slf4j.LoggerFactory
   *
   */
 class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T],
-                                 onPushCallback: () => Unit = (() => {}))(implicit serializer: QueueSerializer[T])
+                                 onPushCallback: () => Unit = () => {})(implicit serializer: QueueSerializer[T],
+                                                                        system: ActorSystem)
   extends GraphStage[UniformFanOutShape[T, Event[T]]] {
 
-  def this(config: Config)(implicit serializer: QueueSerializer[T]) = this(new PersistentQueue[T](config))
+  def this(config: Config)(implicit serializer: QueueSerializer[T],
+                           system: ActorSystem) = this(new PersistentQueue[T](config))
 
-  def this(persistDir: File)(implicit serializer: QueueSerializer[T]) = this(new PersistentQueue[T](persistDir), () => {})
+  def this(persistDir: File)(implicit serializer: QueueSerializer[T],
+                             system: ActorSystem) = this(new PersistentQueue[T](persistDir), () => {})
 
   def withOnPushCallback(onPushCallback: () => Unit) = new BroadcastBuffer[T](queue, onPushCallback)
 
   def withOnCommitCallback(onCommitCallback: Int => Unit) = new BroadcastBuffer[T](queue.withOnCommitCallback(onCommitCallback), onPushCallback)
 
-  private[stream] val outputPorts = queue.totalOutputPorts
-  private[stream] val in = Inlet[T]("BroadcastBuffer.in")
-  private[stream] val out = Vector.tabulate(outputPorts)(i ⇒ Outlet[Event[T]]("BroadcastBuffer.out" + i))
+  private val outputPorts = queue.totalOutputPorts
+  private val in = Inlet[T]("BroadcastBuffer.in")
+  private val out = Vector.tabulate(outputPorts)(i ⇒ Outlet[Event[T]]("BroadcastBuffer.out" + i))
+  private val outWithIndex = out.zipWithIndex
   val shape: UniformFanOutShape[T, Event[T]] = UniformFanOutShape(in, out: _*)
-  @volatile private var finished = IndexedSeq.fill[Boolean](outputPorts)(false)
   @volatile private var upstreamFailed = false
+  @volatile private var upstreamFinished = false
+  private val queueCloserActor = system.actorOf(Props(classOf[PersistentQueueCloserActor[T]], queue))
 
   def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
 
-    private var upstreamFinished = false
+    private val finished = Array.fill[Boolean](outputPorts)(false)
+    private val lastPushed = Array.fill[Long](outputPorts)(0)
 
     override def preStart(): Unit = pull(in)
 
     def outHandler(outlet: Outlet[Event[T]], outputPortId: Int) = new OutHandler {
       override def onPull(): Unit = {
-        queue.dequeue(outputPortId, finished(outputPortId)) match {
+        queue.dequeue(outputPortId) match {
           case None => if (upstreamFinished) {
-              finished = finished.updated(outputPortId, true)
-              if(finished.reduce(_ && _)) {
-                completeStage()
-              }
+            finished(outputPortId) = true
+            queueCloserActor ! PushedAndCommitted(outputPortId, lastPushed(outputPortId), queue.read(outputPortId))
+            if (finished.reduce(_ && _)) {
+              queueCloserActor ! UpstreamFinished
+              completeStage()
             }
+          }
           case Some(element) =>
             push(outlet, Event(outputPortId, element.index, element.entry))
-            queue.lastPushedIndex(outputPortId) = element.index
+            lastPushed(outputPortId) = element.index
+            if (queue.autoCommit) queue.commit(outputPortId, element.index)
         }
       }
     }
@@ -85,36 +95,59 @@ class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T],
         val element = grab(in)
         queue.enqueue(element)
         onPushCallback()
-        out.iterator.zipWithIndex foreach { case (port, id) =>
+        outWithIndex foreach { case (port, id) =>
           if (isAvailable(port))
-            queue.dequeue(id, finished(id)) foreach { element =>
-              push(out(id), Event(id, element.index, element.entry))
-              queue.lastPushedIndex(id) = element.index
+            queue.dequeue(id) foreach { element =>
+              push(port, Event(id, element.index, element.entry))
+              lastPushed(id) = element.index
+              if (queue.autoCommit) queue.commit(id, element.index)
             }
         }
         pull(in)
       }
 
-      override def onUpstreamFinish(): Unit = upstreamFinished = true
+      override def onUpstreamFinish(): Unit = {
+        upstreamFinished = true
+        var isAllAvailable = true
+        outWithIndex foreach { case (port, outportId) =>
+          if (isAvailable(port)) {
+            finished(outportId) = true
+            queueCloserActor ! PushedAndCommitted(outportId, lastPushed(outportId), queue.read(outportId))
+          } else {
+            isAllAvailable = false
+          }
+        }
+
+        if (isAllAvailable) {
+          queueCloserActor ! UpstreamFinished
+          completeStage()
+        }
+      }
 
       override def onUpstreamFailure(ex: Throwable): Unit = {
         val logger = Logger(LoggerFactory.getLogger(this.getClass))
         logger.error("Received upstream failure signal: " + ex)
         upstreamFailed = true
-        queue.close()
+        queueCloserActor ! UpstreamFailed
         completeStage()
       }
     })
 
-    out.zipWithIndex foreach { case (currentOut, outputPortId) =>
+    outWithIndex foreach { case (currentOut, outputPortId) =>
       setHandler(currentOut, outHandler(currentOut, outputPortId))
     }
   }
 
-  val commit = Flow[Event[T]].map { element =>
-    queue.commit(element.outputPortId, element.commitOffset, finished(element.outputPortId), upstreamFailed)
-    element
-  }
+  val commit = if (queue.autoCommit) Flow[Event[T]]
+               else {
+                  Flow[Event[T]].map { element =>
+                    if (!upstreamFailed) {
+                      queue.commit(element.outputPortId, element.commitOffset)
+                      if(upstreamFinished) queueCloserActor ! Committed(element.outputPortId, element.commitOffset)
+                    }
+                    element
+                  }
+               }
 
   def clearStorage() = queue.clearStorage()
 }
--- a/squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
+++ b/squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
@@ -68,11 +68,10 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
   private var indexMounted = false
   private var indexFile: IndexFile = _
   private var indexStore: MappedBytesStore = _
+  private var closed = false
 
   val totalOutputPorts = outputPorts
-  val lastPushedIndex = Array.ofDim[Long](totalOutputPorts)
-  @volatile private var reachedToPushedIndex = IndexedSeq.fill[Boolean](totalOutputPorts)(false)
-  private var closed = false
+  val autoCommit = config.autoCommit
 
   private def mountIndexFile(): Unit = {
     indexFile = IndexFile.of(path, OS.pageSize())
@@ -107,17 +106,14 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
     *
     * @return The first element in the queue, or None if the queue is empty.
     */
-  def dequeue(outputPortId: Int = 0, reachedEndOfQueue: Boolean = false): Option[Entry[T]] = {
+  def dequeue(outputPortId: Int = 0): Option[Entry[T]] = {
     var output: Option[Entry[T]] = None
     if (reader(outputPortId).readDocument(new ReadMarshallable {
       override def readMarshallable(wire: WireIn): Unit =
         output = {
           val element = serializer.readElement(wire)
           val index = reader(outputPortId).index
-          element map { e =>
-            if (autoCommit) internalCommit(outputPortId, index, reachedEndOfQueue)
-            Entry[T](index, e)
-          }
+          element map { e => Entry[T](index, e) }
         }
     })) output else None
   }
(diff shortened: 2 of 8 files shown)


squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -107,17 +105,14 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
...
-  def dequeue(outputPortId: Int = 0, reachedEndOfQueue: Boolean = false): Option[Entry[T]] = {
+  def dequeue(outputPortId: Int = 0): Option[Entry[T]] = {
     var output: Option[Entry[T]] = None
     if (reader(outputPortId).readDocument(new ReadMarshallable {
       override def readMarshallable(wire: WireIn): Unit =
         output = {
           val element = serializer.readElement(wire)
           val index = reader(outputPortId).index
-          element map { e =>
-            if (autoCommit) internalCommit(outputPortId, index, reachedEndOfQueue)
```
akara: Hmmm, we're removing the `internalCommit` which means we make the queue itself unaware of whether to guard for autoCommit and rely on the callers to do the guarding. What do you think?

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentBufferSpec.scala
```diff
@@ -121,24 +121,24 @@ abstract class PersistentBufferSpec[T: ClassTag, Q <: QueueSerializer[T]: Manife
...
+      e
+    }.toMat(Sink.ignore)(Keep.right)
 
     val shutdownF = finishedGenerating.future map { d => mat.shutdown(); d }
 
     val graph = RunnableGraph.fromGraph(GraphDSL.create(Sink.ignore) { implicit builder =>
       sink =>
         import GraphDSL.Implicits._
         val buffer = new PersistentBuffer[T](config).withOnPushCallback(() => pBufferInCount.incrementAndGet()).withOnCommitCallback(() =>  commitCount.incrementAndGet())
+        val pBuffer = builder.add(buffer.async)
```
akara: What is exactly the difference to add or not to add?

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -55,27 +59,33 @@ class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T],
   private[stream] val in = Inlet[T]("BroadcastBuffer.in")
   private[stream] val out = Vector.tabulate(outputPorts)(i ⇒ Outlet[Event[T]]("BroadcastBuffer.out" + i))
   val shape: UniformFanOutShape[T, Event[T]] = UniformFanOutShape(in, out: _*)
-  @volatile private var finished = IndexedSeq.fill[Boolean](outputPorts)(false)
   @volatile private var upstreamFailed = false
+  private val queueCloserActor = system.actorOf(Props(classOf[PersistentQueueCloserActor[T]], queue))
```
anilgursel: During closing would mean once upstream is finished?  That might still mean that we send many messages afterwards.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -87,21 +97,32 @@ class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T],
         onPushCallback()
         out.iterator.zipWithIndex foreach { case (port, id) =>
           if (isAvailable(port))
-            queue.dequeue(id, finished(id)) foreach { element =>
+            queue.dequeue(id) foreach { element =>
               push(out(id), Event(id, element.index, element.entry))
-              queue.lastPushedIndex(id) = element.index
+              queueCloserActor ! Pushed(id, element.index)
```
anilgursel: I agree.  However, when should we start this?  We would have the problem with the `commit` messages as well.  I do not want to introduce another `volatile` for it.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -112,7 +133,10 @@ class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T],
   }
 
   val commit = Flow[Event[T]].map { element =>
-    queue.commit(element.outputPortId, element.commitOffset, finished(element.outputPortId), upstreamFailed)
+    if(!queue.autoCommit && !upstreamFailed) {
```
anilgursel: The queue does not check autocommit anymore.  It is the responsibility of the `BroadcastBuffer` or `PersistentBuffer` to decide on the `commit`.  The reason I moved it up is to not have to pass the listener actor ref to the queue.

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentBufferSpec.scala
anilgursel: I was trying this while debugging an issue.  This should not make a difference actually.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
anilgursel: Yes.  That's why I mentioned on the Gitter chat that this may be controversial.  However, I do this because I do not want to pass the actor ref to the queue.  So, queue knowing about the actor and actor knowing about the queue does not sound right neither.  Moreover, queue knowing about the actor makes the code quite complicated.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
anilgursel: Also, I do not actually need to check `queue.autoCommit` inside the map, I should check it outside of Flow, during blueprint creation.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
anilgursel: So, for `Pushed`, it is easy to just send after `finished(outputPortId)` turns `true` or in `onUpstreamFinish` if `isAvailable` is `true` for the port.  However, the problem is with `Committed` (for the non-auto-commit scenario).  To be able to send `Committed` only after certain event, we need to introduce `volatile` with many writes.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
akara: Yes, that should be fine. This is the "best effort" stage. Key is not to cause runtime overhead. We care less about shutdown-time overhead.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
akara: Fair point. It is more of a compromise than controversial 😄

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentBufferSpec.scala
akara: I'd keep it simple in that case. `simple == better`.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
anilgursel: "We care less about shutdown-time overhead." -> What we think the shutdown-time overhead might actally be big percentage of the stream lifecycle depending on the the logic in the stream and the speed of downstream vs upstream.  In other words, `upstreamFinished` might get called when downstream is still in the first 10%.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
anilgursel: I am not sure if I agree with it being a "compromise" :)  If we let commit to be called manually, I find it more natural the "stage" controlling the auto-commit behavior.  Passing the in queue configuration does not sound right though, I agree.

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentBufferSpec.scala
anilgursel: I agree.

anilgursel:
@akara I have reduced the `Pushed` message traffic to only shutdown time already, and it is true shutdown time, it will only send constant number of messages (based on the output port number), not for all messages after `upstreamFinished`.

However, doing the same for `Committed` message gets tricky and requires `volatile` for elements that are updated with every message.  It may turn out to be more expensive than the actor message traffic.  

Please let me know your suggestions around that.

anilgursel:
Actually, one thing I can do is that making the following `PersistentQueue` method accessible from `BroadcastBuffer` and `PersistentBuffer`:

```scala
// Reads the given outputPort's queue index
private def read(outputPortId: Int) : Long = {
  if (!indexMounted) mountIndexFile()
  indexStore.readLong(outputPortId << 3)
}
```

Then, I do not need to have `volatile` and no need to send `Committed` message each time..  Let me work on this and see how it looks.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
akara: Sure. But given the choice between shutdown-time overhead and all-time overhead, I'd still choose the first. Of course, unless the cost per message is much higher, then we have to weigh our options.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
akara: We have to think about the auto-commit case. But yeah, if there is no drawback for that I'm good to stay with the model of externally controlled commits.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -55,27 +59,33 @@ class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T],
   private[stream] val in = Inlet[T]("BroadcastBuffer.in")
   private[stream] val out = Vector.tabulate(outputPorts)(i ⇒ Outlet[Event[T]]("BroadcastBuffer.out" + i))
```
akara: I'd create one more variable that you use a lot during stream processing:

```scala
private[stream] val outWithIndex = out.zipWithIndex
```

Then wherever you use `out.zipWithIndex`, use `outWithIndex` instead. You won't deal with creating tuples all the time and running further logic, for instance, in `onPush` which is very frequently called.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -87,21 +97,34 @@ class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T],
         onPushCallback()
         out.iterator.zipWithIndex foreach { case (port, id) =>
```
akara: Here you won't need an `iterator` and can just use `outWithIndex` directly.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -87,21 +97,34 @@ class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T],
...
           if (isAvailable(port))
-            queue.dequeue(id, finished(id)) foreach { element =>
+            queue.dequeue(id) foreach { element =>
               push(out(id), Event(id, element.index, element.entry))
```
akara:
```scala
push(port, Event(id, element.index, element.entry))
```

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -87,21 +97,34 @@ class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T],
...
+              if (queue.autoCommit) queue.commit(id, element.index)
             }
         }
         pull(in)
       }
 
-      override def onUpstreamFinish(): Unit = upstreamFinished = true
+      override def onUpstreamFinish(): Unit = {
+        upstreamFinished = true
+        out.zipWithIndex filter { case (port, _) => isAvailable(port) } foreach { case (_, outportId) =>
```
akara: Use `outWithIndex`. Also, we use `filter` and `foreach` on a non-lazy collection. This will cause two loops. So better to use `outWithIndex.iterator` to make it lazy. But even with lazy.

Between that and using `collect` that yields `Unit`, I'm wondering what is more efficient.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -87,21 +97,34 @@ class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T],
...
+          finished(outportId) = true
+          queueCloserActor ! Pushed(outportId, lastPushed(outportId))
+          queueCloserActor ! Committed(outportId, queue.read(outportId))
+        }
+
+        if (out.forall(isAvailable(_))) {
```
akara: `if (out.forall(isAvailable))` is adequate.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -111,10 +134,16 @@ class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T],
...
 
-  val commit = Flow[Event[T]].map { element =>
-    queue.commit(element.outputPortId, element.commitOffset, finished(element.outputPortId), upstreamFailed)
-    element
-  }
+  val commit = if (queue.autoCommit) Flow[Event[T]]
+               else {
+                  Flow[Event[T]].map { element =>
+                    if (!upstreamFailed) {
+                      queue.commit(element.outputPortId, element.commitOffset)
```
akara: If `upStreamFailed` and downstream is still processing and actively committing, do you want to do a noop on that commit?

akara:
Reviewed the rest. Other than the few comments provided on BroadcastBuffer, LGTM!

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
anilgursel: `out.zipWithIndex` in `onPush` had been there prior to this PR, but I agree and will make the change.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
anilgursel: I will use `iterator`.  Regarding `collect`, the loop will be for the number of outports, which is mostly a low single digit.  Also, `onUpstreamFinish` will be called only once, so not sure if it would make any significant performance difference.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
anilgursel: 👍

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
anilgursel: If `upstreamFailed`, we need to close the `queue` somewhere, and since there is no hook at the `commit` stage, we need to close it inside `onUpstreamFailed` of `BroadcastBuffer`.  Once we close the queue, there is not guarantee in ordering and concurrency and we do not want to call `commit` if the queue is already closed (it would cause an exception).

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
anilgursel: Changing the logic to 

```scala
override def onUpstreamFinish(): Unit = {
        upstreamFinished = true
        var isAllAvailable = true
        outWithIndex foreach { case (port, outportId) =>
          if (isAvailable(port)) {
            finished(outportId) = true
            queueCloserActor ! Pushed(outportId, lastPushed(outportId))
            queueCloserActor ! Committed(outportId, queue.read(outportId))
          } else {
            isAllAvailable = false
          }
        }

        if (isAllAvailable) {
          queueCloserActor ! UpstreamFinished
          completeStage()
        }
      }
```

akara:
I'm good with the changes. Thank you very much!

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueueCloserActor.scala
```diff
@@ -24,7 +24,9 @@ class PersistentQueueCloserActor[T](queue: PersistentQueue[T]) extends Actor wit
   val commitIndex = Array.ofDim[Long](queue.totalOutputPorts)
 
   override def receive: Receive = {
-    case Pushed(outportId, index) => pushIndex(outportId) = index
+    case PushedAndCommitted(outportId, pIndex, cIndex) =>
+      pushIndex(outportId) = pIndex
+      if(cIndex > commitIndex(outportId)) commitIndex(outportId) = cIndex
```
akara: Please double-check that this comparison is always valid `cIndex > commitIndex(outportId)`. This must hold true even when the file rolls. Likely it does, but we need to be extra careful.

squbs-pattern/src/test/scala/org/squbs/pattern/stream/BroadcastBufferSpec.scala
```diff
@@ -22,21 +22,26 @@ import akka.actor.ActorSystem
...
+   (typeName: String, autoCommit: Boolean = true) extends FlatSpec with Matchers with BeforeAndAfterAll with Eventually {
 
   implicit val system = ActorSystem(s"Broadcast${typeName}BufferSpec")
   implicit val mat = ActorMaterializer()
   implicit val serializer = QueueSerializer[T]()
   import StreamSpecUtil._
   import system.dispatcher
 
+  implicit override val patienceConfig =
+    PatienceConfig(timeout = scaled(Span(30, Seconds)), interval = scaled(Span(500, Millis)))
```
anilgursel: I should actually not need this anymore.  But, if I still need, I will use `awaitMax`.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueueCloserActor.scala
anilgursel: Good point, I will confirm!

Issue #243, opened 2016-08-17
BroadcastBuffer does not need `event` type for auto commit enabled


anilgursel:

akara:
Fixed by #281.

PR #271, opened 2016-09-28
#270 : persistent queue auto cleanup handling in compliance to the recent ve…

### LINKED ISSUE #270: persistent queue auto cleanup handling
The current version of persistent queue provides `clearStorage` api to clients for cleaning up the queue files. Instead, implement a standard to cleanup resources with all `outputPorts` (subscribers) acknowledge onReleased callback.
reference document: https://groups.google.com/forum/#!topic/java-chronicle/0Nz5P-nvLgM

SarathChandran:

- [X] Commits are squashed.
- [x] Tests added/updated.

Description:
Implements a standard resources cleanup function which clean queue files when all outputPorts (subscribers) acknowledge onReleased callback.


squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
```diff
@@ -138,9 +138,9 @@ class BroadcastBuffer[T] private(private[stream] val queue: PersistentQueue[T],
     }
   }
 
-  val commit = if (queue.autoCommit) Flow[Event[T]]
+  def commit[Q] = if (queue.autoCommit) Flow[Event[Q]]
```
anilgursel: Why do we need this to be a `def`?  For a given `BroadcastBuffer` instance, there should only be one `queue` and the `autoCommit` should be immutable.

anilgursel: Also, not sure why we need a separate `[Q]`, instead of `[T]`.  If we need to use a different character, why `Q?

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentBuffer.scala
```diff
@@ -79,10 +79,10 @@ class PersistentBuffer[T] private(private[stream] val queue: PersistentQueue[T],
         onPushCallback()
         if (downstreamWaiting) {
           queue.dequeue() foreach { element =>
-            push(out, Event(defaultOutputPort, element.index, element.entry))
+            push(out, element)
             downstreamWaiting = false
-            lastPushed = element.index
-            if(queue.autoCommit) queue.commit(defaultOutputPort, element.index)
+            lastPushed = element.commitOffset
```
anilgursel: `commitOffset` does not sound the right name here.  Gets little confusing.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -137,39 +135,45 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
...
+    Option(indexStore) foreach { store => if (store.refCount > 0) store.release() }
+    Option(indexFile) foreach { file => file.release() }
   }
 
   private[stream] def isClosed = closed
 
   /**
-    * Removes processed queue files based on
-    * min read index from all outputPorts
+    * Removes processed queue files all outputPorts
```
anilgursel: I am not clear about this comment.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -137,39 +135,45 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
...
-    def delete(previous: Resource) : Unit = {
-      if (current.path != previous.path) {
-        previous.path.delete()
-        delete(first())
+  class ResourceManager extends StoreFileListener {
+    var counter = Map.empty[Int, AtomicInteger]
+    var lastFile: Option[File] = None
+
+    override def onAcquired(cycle: Int, file: File): Unit = {
+      logger.info(s">>>>>>>>>> File Acquired ${(cycle, file)}")
```
anilgursel: I would not have `>>>>>>>>>>` in the log message.  Also, using the `logger.info("File Acquired {} {}", cycle, file)` syntax would be better to prevent unnecessary string interpolation when the log level is `warn` etc..
 `

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -137,39 +135,45 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
...
+      super.onAcquired(cycle, file)
+    }
+
+    override def onReleased(cycle: Int, file: File): Unit = {
+      counter += (cycle -> count(cycle))
+      if (counter(cycle).intValue() == outputPorts) {
+        logger.info(s">>>>>>>>>> File Released ${(cycle, file)}")
```
anilgursel: Ditto

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -137,39 +135,45 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
...
+        counter = Map.empty[Int, AtomicInteger]
```
anilgursel: Is it possible that it just started a new cycle, and only few output ports are moved to the new cycle and one of those already released the new cycle while some other output port is still on the old cycle and not release.  Some time later, the behind port is releasing the old cycle? 

In other words, do we want to do `counter = Map.empty[Int, AtomicInteger]` or set it to `0` only for that cycle?

squbs-pattern/src/test/scala/org/squbs/pattern/stream/BroadcastBufferSpec.scala
```diff
@@ -60,7 +60,7 @@ abstract class BroadcastBufferSpec[T: ClassTag, Q <: QueueSerializer[T] : Manife
     val streamGraph = RunnableGraph.fromGraph(GraphDSL.create(flowCounter) { implicit builder =>
       sink =>
         import GraphDSL.Implicits._
-        val commit = buffer.commit // makes a dummy flow if autocommit is set to false
+        val commit = buffer.commit[T] // makes a dummy flow if autocommit is set to false
```
anilgursel: U see a compile issue with this?

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -106,28 +104,28 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
     *
     * @return The first element in the queue, or None if the queue is empty.
     */
-  def dequeue(outputPortId: Int = 0): Option[Entry[T]] = {
-    var output: Option[Entry[T]] = None
+  def dequeue(outputPortId: Int = 0): Option[Event[T]] = {
```
anilgursel: Do we need to make this change with this PR and can you please explain the reasoning behind it?  Wondering https://github.com/paypal/squbs/issues/243 would affected with this.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
akara: I think I can answer that question. Take this stream, for instance:

```scala
source ~> bcBuffer ~> enrich ~> commit ~> sink
```

If `enrich` translates `Event[T]` to `Event[Q]`, the current code would not allow the commit as the event has to be of the same type. By decoupling the type from the type of the `BroadcastBuffer`, it allows you to run any translation as long as the Event envelope is retained, which means you still have the index to commit to.

squbs-pattern/src/test/scala/org/squbs/pattern/stream/BroadcastBufferSpec.scala
akara: As a `def`, this should be fine, right?

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentBuffer.scala
SarathChandran: the `case class Entry` used previously was a unnecessary and everything could have been simply in `case class Event`. And by its definition this holds `(outputPort, commitOffset, element)` and that's the reason the rename is reflected here.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -137,39 +135,45 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
...
 
-  private def minIndex() = Iterator.tabulate(outputPorts)(outputPortId => read(outputPortId)).min
-
   /**
     * Closes the queue and all its persistent storage.
     */
   def close(): Unit = {
     closed = true
-    if (written) appender.close()
-    reader.foreach { m => m.close() }
```
akara: Can you please remind me why we take out the `appender.close()` and the reader closings?

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
anilgursel: Got it!

But, should the second generic type parameter name be `Q`?  We do not seem to be following Scala convention of `A`, `B`, `C` here, and rather following Java one, which I am totally fine.  But, in that case, shouldn't the second generic type be `S`?  http://docs.oracle.com/javase/tutorial/java/generics/types.html

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentBuffer.scala
anilgursel: Yes, just the name.  I would prefer `index`.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
anilgursel: I am actually ok with the change.  If we are not using `Entry` anymore, can we make sure to delete the class.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/BroadcastBuffer.scala
SarathChandran: sure @anilgursel will change it to `S` type name

squbs-pattern/src/test/scala/org/squbs/pattern/stream/BroadcastBufferSpec.scala
SarathChandran: I was not able to compile without adding type. I am adding the error i got in short `expected Flow[Event[T],....], actual Flow[Event[Nothing],...]`

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
SarathChandran: yes @anilgursel `Entry` case class is removed https://github.com/paypal/squbs/pull/271/files#diff-717b79f6434ac986220216a2e1c86d40L31

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
SarathChandran: The scenario you are mentioning is possible. each `reader` processing at own pace might end up processing different queue cycles. to make sure that we keep track on the release status of each `reader` i have used counter Map.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
anilgursel: thanks!

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
anilgursel: But `counter = Map.empty[Int, AtomicInteger]` is resetting the entire map not just that cycle.  So, for the scenario I mentioned, we may end up not deleting the file for the next cycle?

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
SarathChandran: @akara the appender and tailer where implementing Autoclosable in v4.3.2 and the current version they have removed Autoclosable and the method is no longer callable.
http://static.javadoc.io/net.openhft/chronicle-queue/4.3.2/net/openhft/chronicle/queue/ExcerptTailer.html
http://static.javadoc.io/net.openhft/chronicle-queue/4.5.13/net/openhft/chronicle/queue/ExcerptTailer.html

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -165,39 +162,44 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
...
+
+    override def onReleased(cycle: Int, file: File): Unit =
+      if (minCycle.equals(cycle)) deleteFiles(cycle, fileNameToLong(file))
+
+    private def minCycle = Iterator.tabulate(outputPorts)(reader(_).cycle).min
+
+    private def fileNameToLong(file: File): Long = file.getName
+      .stripSuffix(SingleChronicleQueue.SUFFIX).split("-").last.toLong
+
+    private def deleteFiles(cycle: Int, fileNameInLong: Long) = Option(persistDir.listFiles) map {
```
anilgursel: Why do we not delete just the specific file when all outports are done with that cycle?  Here we are instead traversing through the directory and delete all files that has a cycle number less than it.  Is there a known scenario that requires this behavior?

If we do `DAILY`, this may not be an issue; however, if roll cycle is `MINUTELY`, and there is a difference in the pace of downstream branches, this may turn out to be costly file operations.

I am ok to do this traversing when things go wrong (unexpected shutdown/failure etc), but not for the happy path.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -165,39 +162,44 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
...
-        delete(first())
+  class ResourceManager extends StoreFileListener {
+
+    override def onAcquired(cycle: Int, file: File): Unit = {
+      logger.info("File Acquired {} - {}", cycle.toString, file.getPath)
+      super.onAcquired(cycle, file)
+    }
+
+    override def onReleased(cycle: Int, file: File): Unit =
+      if (minCycle.equals(cycle)) deleteFiles(cycle, fileNameToLong(file))
```
SarathChandran: @anilgursel onReleased is called when one of the `outputPorts` have completed the cycle, so if at least if 1 of the `outputPorts` is less that the current cycle (means, the cycle's file still under process). We back off the delete for this particular callback. 

We wait until the slowest reader completes the cycle. Hence not having `<` check. Also since the slowest reader calling onReleased at last will always be equal to minCycle (all readers would have passed beyond that cycle, expect for this one), I think having `=` should be good. 

Taking fast executions to account. There is could be a chance that all readers have moved beyond the callback's cycle value. probably `if (minCycle() >= cycle)` can be a proper check here.

anilgursel: "all readers would have passed beyond that cycle, expect for this one" -> is there a guarantee that the slowest one has not already moved to the next cycle?  In other words, will moving to the next cycle be blocked until `onRelease` completes?

anilgursel: Yes, you are right, actually I had a typo or I thought it wrong in the first comments.  `if (minCycle() >= cycle)` is the right one I think.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
SarathChandran: Since we don't know how many times an `onReleased` callback gets called for a cycle (unexpected shutdown calls `onReleased` even if file is half read) the counter implementation i had previously didn't work out well. 

This was because `no. of outputPorts != no. of callbacks` so even if minCycle condition matches (for unexpected shutdown callbacks) we shouldn't actually delete the current cycle's file. 

Since unexpected cases are non-deterministic, we have to remove all previous files. Mostly in happy path, its just one file.

anilgursel: So, the problem is for unexpected shutdown/failure cases where we close the queue.  So, when we close the queue in such scenarios, `onRelease` gets called as well, but we should not be counting that, is that a true understanding?

If so, then how about checking for `isClosed`?  If it is already `closed`, then we should not increment the counter.  For the happy scenario, where we actually close the queue when everyone reaches to the end, we can manually delete the file.

I understand it is only one file we delete; however, we are doing a traversal on the files and also doing string manipulation.  That looked little expensive to me and there should be ways to eliminate.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
akara: @anilgursel Yes, we were discussing cycle wrapping. That's kind of the ultimate scenario and we need to see whether chronicle handles that case. It would be an awfully long time. I'm also good without that.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -165,39 +162,44 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
...
+
+    private def minCycle = Iterator.tabulate(outputPorts)(reader(_).cycle).min
```
akara: If you keep it as var, you have to have events that trigger changing that var. A cycle change for each reader will have to cause this var to change. But besides that, the code can be written without tabulating by using `reader.iterator.map(_.cycle).min`. That might make this code a bit more understandable. Let me check whether kicking a var around with events would be worth it.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
akara: Agree that we should find ways to eliminate processing. But this should not be on the performance-critical path. It will happen once a day, once an hour, or some granularity like that. I can easily buy into a little less optimization for clean code in such cases. But I'll leave it to your judgement.

akara:
@SarathChandran I think the main thing is to change the commit message and the PR title as @anilgursel suggested. The other parts are minor to me. But please do take care of those. Thanks very much @anilgursel for the other pair of eyes.

anilgursel:
@SarathChandran I might have missed it, is there any test case for testing the auto delete?  We should have on for `PersistentBuffer` and one for `BroadcastBuffer`, I mean, making sure that we actually delete the files.  There seems to be a `TEST_SECONDLY` option, which might help on this https://github.com/OpenHFT/Chronicle-Queue/blob/master/src/main/java/net/openhft/chronicle/queue/RollCycles.java#L21.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
anilgursel: `reader(_).cycle` won't be doing any file read and no heavy calculation, `cycle` is just returning the value.  So, I think `reader.iterator.map(_.cycle).min` should be good.  The other option is to keep an array and updating those at https://github.com/paypal/squbs/blob/master/squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala#L145 where we do `cycle` change.

I agree, `reader.iterator.map(_.cycle).min` is a lot simpler and cleaner.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
anilgursel: @akara It will not happen often; however, once happens, it might affect the latency on the happy path.  Also, `MINUTELY` can see this more often.  I would be more comfortable if potentially heavier processing happens only for non-happy path or during close.  I agree on the cleaner code argument.  However, not sure if readability would be degraded significantly if we followed the other approach. 

@SarathChandran Please help me understand.  From the Gitter chat, my understanding is that there are two problems with the previous approach:

1. State loss if a restart happens.  -> We actually already keep `cycle` information, and if a restart happens it should be initialized properly, please see https://github.com/paypal/squbs/blob/master/squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala#L92.

2. `onRelease` could be called with `queue.close` for unexpected failure/shutdown.  -> The very first thing we do when `close` is called is to set the `var close` to `true`, please see https://github.com/paypal/squbs/blob/master/squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala#L174.  So, if we skip any clean up operation if `isClosed` is true, I would assume we eliminate the corner cases.  So, this would probably cause the happy `close` scenario not to clean the files.  But, we know where `close` is called in happy path scenario in `PersistentQueueCloserActor`:

   - https://github.com/paypal/squbs/blob/master/squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueueCloserActor.scala#L33
   - https://github.com/paypal/squbs/blob/master/squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueueCloserActor.scala#L38

So, we can directly call the clean up from there.

I think this would eliminate some of the processing cost for happy path, while the readability of the code won't be bad(IMO).  If you think otherwise, I am ok to merge (given the other comments are addressed) and we can discuss more later and tackle with another issue.

SarathChandran: @anilgursel above cases are totally true. But, the problem i faced is we do not have `cycleNumber vs fileName` mapping. So after a restart, even though we have cycle's info, the fileName is not available to us. This was the reason i used a `var counter map`. The proper case is to persist this mapping itself to maintain state. (which again i think is costlier (reads/writes for each onReleased)  than current approach). 

Ideally, if fileName info is available for given cycle number, i would not have chosen traversal at all.
Also, (IMHO) most of production systems will never use short span cycles like minutely, which safely puts us cases for hourly and above modes. But this choice is totally left to consumers :(

anilgursel: I am actually trying to minimize the impact on the happy flow (when it is not closing or failure etc).  For `MINUTELY`, we would see more impact.  I am just trying to see if we can eliminate a few potential latency increase every hour or every day without complicating the code.  I think here are the scenarios:

1. Unexpected failure/shutdown: we should not be deleting any files and we are able to prevent it by checking the `closed` flag.  When the stream restarts, we know which output port is on which cycle.  When the slowest outport moves the next cycle, `onRelease` will still be called with the `file` and we will be able to check if other outports are already ahead and close the file if needed.  

2. Happy flow, the cycle rolls (no failures) to the next hour: by keeping which outport is at which cycle currently, we will be able to check if we can delete the `file` passed in `onRelease`.  Once `onRelease` is called for the slowest outport, we delete the file passed in as a parameter.

3. The queue is closed because `upstream` is finished and `commit` index reached to the `pushed` index for each output port.  We manually call `delete`, if needed we can do a traversal here and I would not be much concerned about any cost.  However, since this is a happy shutdown scenario, we would probably not need the files with the next restart of the system or stream, so we could actually delete all the files in the directory.

If there is any other scenario I am missing, please let me know.

anilgursel:
@SarathChandran Let's tackle the `if (minCycle() >= cycle)` issue in the PR and work on the optimization with another PR afterwards.  Let's get this merged.

PR #272, opened 2016-09-29
#242 PersistentBuffer & BroadcastBuffer commit safety

anilgursel:

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
+++ b/squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
@@ -69,6 +69,8 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
   private var indexFile: IndexFile = _
   private var indexStore: MappedBytesStore = _
   private var closed = false
+  private val cycle = Array.ofDim[Int](outputPorts)
+  private val lastCommitIndex = Array.ofDim[Long](outputPorts)
 
   val totalOutputPorts = outputPorts
   val autoCommit = config.autoCommit
@@ -86,6 +88,8 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
       logger.info("Setting idx for outputPort {} - {}", outputPortId.toString, startIdx.toString)
       reader(outputPortId).moveToIndex(startIdx)
       dequeue(outputPortId) // dequeue the first read element
+      lastCommitIndex(outputPortId) = startIdx
+      cycle(outputPortId) = queue.rollCycle().toCycle(startIdx)
     }
   }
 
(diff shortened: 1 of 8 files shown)


squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -126,11 +130,34 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
...
   def commit(outputPortId: Int, index: Long): Unit = {
+    verifyCommitOrder(outputPortId, index)
     if (!indexMounted) mountIndexFile()
     indexStore.writeLong(outputPortId << 3, index)
     onCommitCallback(outputPortId)
   }
 
+  private def verifyCommitOrder(outputPortId: Int, index: Long): Unit = {
+    val lastCommit = lastCommitIndex(outputPortId)
+    if(index == lastCommit + 1) lastCommitIndex(outputPortId) = index
```
anilgursel: Do we actually want this if check even for `Lenient` use case?  If we would like to add a log, we should..  However, if log is not needed, then I can control this at the buffer level and use a different function..

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -126,11 +130,34 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
...
+        lastCommitIndex(outputPortId) = index
+      } else {
+        config.commitOrderPolicy match {
+          case Lenient =>
+            logger.warn(s"Missing or out of order commits.  previous: ${lastCommitIndex(outputPortId)} latest: $index cycle: ${cycle(outputPortId)}")
+            lastCommitIndex(outputPortId) = index
+          case Strict =>
+            val msg = s"Missing or out of order commits.  previous: ${lastCommitIndex(outputPortId)} latest: $index cycle: ${cycle(outputPortId)}"
+            logger.error(msg)
+            throw new CommitOrderException(msg, lastCommitIndex(outputPortId), index, cycle(outputPortId))
```
anilgursel: Not closing the queue here as the `SupervisionDecider` might resume the stream.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
akara: Looking at the code, it costs pretty much one single branch and one non-concurrent assignment. I think that should not be an issue on modern processors.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
akara: Thanks for the clarification. Perhaps you want to stick this into a comment on the code, too. Future maintainers won't make the mistake of closing the queue. They usually don't look at your github comments on the original commits before they make the mistake 😄

squbs-pattern/src/test/scala/org/squbs/pattern/stream/BroadcastBufferCommitOrderSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+    import util._
+    val buffer = new BroadcastBuffer[Int](ConfigFactory.parseString("commit-order-policy = strict").withFallback(config))
+    val streamGraph = RunnableGraph.fromGraph(GraphDSL.create(flowCounter) { implicit builder =>
+      sink =>
+        import GraphDSL.Implicits._
+        val commit = buffer.commit // makes a dummy flow if autocommit is set to false
+      val bcBuffer = builder.add(buffer.async)
+        val mr = builder.add(merge)
+        in ~> bcBuffer ~> filterARandomElement ~> commit ~> mr ~> sink
+        bcBuffer ~> commit ~> mr
```
akara: Nice simulation of the failure!

Issue #273, opened 2016-10-11
Support for Flow (similar to Route) for handling http requests


akara:
User ask for submitting a `Flow[HttpRequest, HttpResponse]` instead of `Route` in registered service handler.

akara:
Fixed by #292

PR #292, opened 2016-10-23
#273: Support for FlowDefinition

akara:

- [x] Commits are squashed.
- [x] Tests added.

Code, tests, and documentation supporting the FlowDefinition registration.


docs/flow-definition.md
```diff
@@ -0,0 +1,74 @@
...
+	// This example does both request and response chunking.
+    case req @ HttpRequest(_, Uri(_, _, Path(`chunkPath`), _, _), _, _, _) =>
+
+      val responseChunks = req.entity.dataBytes.filter(_.nonEmpty)
+        .map { b => (1, b.length) }
+        .reduce { (a, b) => (a._1 + b._1, a._2 + b._2) }
+        .map { case (chunkCount, byteCount) =>
+          ChunkStreamPart(s"Received $chunkCount chunks and $byteCount bytes.\r\n")
+        }
+        .concat(Source.single(ChunkStreamPart("This is the last chunk!")))
```
anilgursel: This example shows the request chunking well; however, I do not think it is a good example for response chunking :(  I mean, doesn't feel natural.

docs/flow-definition.md
```diff
@@ -0,0 +1,74 @@
...
+}
+```
+
+Of course, you're still encouraged to use the higher level `RouteDefinition`. But using low level artifacts has a few benefits in terms of memory consumption.
+
+##Rules and Behavior of the FlowDefinition
+
+There are a few rules you have to keep in mind when implementing a `FlowDefinition`:
+
+1. It is the responsibility of the application to generate exactly one response for every request and that the ordering of responses matches the ordering of the associated requests (which is relevant if HTTP pipelining is enabled where processing of multiple incoming requests may overlap).
```
anilgursel: This should be two items.  Exactly one response and ordering!

docs/flow-definition.md
```diff
@@ -0,0 +1,74 @@
...
+2. A single `FlowDefinition`'s flow can be materialized multiple times, causing multiple instances of the `Flow` itself. If they access any state in the `FlowDefinition`, it is important to note such access can be concurrent, both for reads and writes. It is not safe for such accesses to read or write mutable state inside the `FlowDefinition`. The use of Akka `Actor`s or `Agent`s is highly encouraged in such situations.
+3. The `FlowDefinition` has access to the `ActorContext` with the field `context` by default. These can be used to create new actors or access other actors.
+4. If the `WebContext` trait is mixed in, the `FlowDefinition` will have access to another field `webContext` will be available. This can be seen in the code above and is used to determine the web context or path from the root where this `FlowDefinition` is handling requests.
```
anilgursel: The first sentence is not complete.

Also, in the second sentence, I would drop "this can be seen in the code above".  IMO, just showing here is good enough an do not need to emphasize it in the code.

docs/flow-definition.md
```diff
@@ -0,0 +1,74 @@
...
+
+Of course, you're still encouraged to use the higher level `RouteDefinition`. But using low level artifacts has a few benefits in terms of memory consumption.
+
+##Rules and Behavior of the FlowDefinition
+
+There are a few rules you have to keep in mind when implementing a `FlowDefinition`:
+
+1. It is the responsibility of the application to generate exactly one response for every request and that the ordering of responses matches the ordering of the associated requests (which is relevant if HTTP pipelining is enabled where processing of multiple incoming requests may overlap).
+2. A single `FlowDefinition`'s flow can be materialized multiple times, causing multiple instances of the `Flow` itself. If they access any state in the `FlowDefinition`, it is important to note such access can be concurrent, both for reads and writes. It is not safe for such accesses to read or write mutable state inside the `FlowDefinition`. The use of Akka `Actor`s or `Agent`s is highly encouraged in such situations.
+3. The `FlowDefinition` has access to the `ActorContext` with the field `context` by default. These can be used to create new actors or access other actors.
```
anilgursel: "These" ?

docs/flow-definition.md
anilgursel: Also, it would make it easier to read if we start each sentence with a bold title, e.g.,:

1. **Exactly one response:** It is the ...
1. **Response Ordering:** The ordering of responses ....
1. **Concurrent State Access:** A single `FlowDefinition`'s flow can be ...

docs/flow-definition.md
```diff
@@ -0,0 +1,74 @@
...
+4. If the `WebContext` trait is mixed in, the `FlowDefinition` will have access to another field `webContext` will be available. This can be seen in the code above and is used to determine the web context or path from the root where this `FlowDefinition` is handling requests.
+5. The `HttpRequest` object is handed to this flow unmodified. The `webContext` is in the `Path` of the request. It is the `FlowDefinition`'s job (as seen above) to handle the request with the knowledge of the `webContext`.
+
+##FlowDefinition Registration
+
+This flow above can be used as part of a cube by declaring it as a service in the `src/main/resources/META-INF/squbs-meta.conf` as follows:
```
anilgursel: I am not sure if we should reference to "cube" here.  I would just say "A `FlowDefinition` should be registered in `squbs-meta.conf` as below:".

docs/flow-definition.md
```diff
@@ -0,0 +1,74 @@
...
+cube-version = "0.0.1"
+squbs-services = [
+    {
+        class-name = org.mycube.SampleFlowSvc
+        web-context = sampleflow
+    }
+]
+```
+
+##Summary
```
anilgursel: Having a summary is against the conciseness of the documentation.  I would drop this section.

docs/flow-definition.md
```diff
@@ -0,0 +1,74 @@
...
+        .reduce { (a, b) => (a._1 + b._1, a._2 + b._2) }
+        .map { case (chunkCount, byteCount) =>
+          ChunkStreamPart(s"Received $chunkCount chunks and $byteCount bytes.\r\n")
+        }
+        .concat(Source.single(ChunkStreamPart("This is the last chunk!")))
+
+      HttpResponse(StatusCodes.OK, entity = Chunked(ContentTypes.`text/plain(UTF-8)`, responseChunks))
+
+    // The last one is to see error handling in action.
+    case HttpRequest(_, Uri(_, _, Path(`exceptionPath`), _, _), _, _, _) =>
```
anilgursel: Why do we need to have this case for the example?  I do not see any discussion about exception handling in the documentation.  And even if we talk about exception handling, I do not see a value of it in the code.

docs/flow-definition.md
```diff
@@ -0,0 +1,74 @@
...
+  val chunkPath = s"$prefix/chunks"
+  val exceptionPath = s"$prefix/throwit"
+
+  def flow = Flow[HttpRequest].map {
+  
+    // A simple ping/pong request/response.
+    case HttpRequest(_, Uri(_, _, Path(`pingPath`), _, _), _, _, _) =>
+      HttpResponse(StatusCodes.OK, entity = "pong")
+
+	// This example does both request and response chunking.
```
anilgursel: Is there a "tab" here?

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/streaming/ServiceRegistry.scala
```diff
@@ -270,4 +272,58 @@ trait RouteDefinition extends Directives {
...
+      if (flowDefTry.isFailure) context.stop(self)
+  }
+}
+
+object FlowDefinition {
+  private[unicomplex] val context = new ThreadLocal[Option[ActorContext]] {
+    override def initialValue(): Option[ActorContext] = None
+  }
+
+  def startFlows[T](fn: => T)(implicit refFactory: ActorContext): T = {
```
anilgursel: `startFlows` or `startFlow` ?

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/streaming/FailedFlow1Spec.scala
```diff
@@ -0,0 +1,72 @@
...
+import scala.concurrent.Await
+import scala.util.Failure
+
+object FailedFlow1Spec {
+
+  val dummyJarsDir = getClass.getClassLoader.getResource("classpaths/streaming").getPath
+
+  val classPath = dummyJarsDir + "/DummyFailedFlowSvc1/META-INF/squbs-meta.conf"
+
+  val (_, _, port) = temporaryServerHostnameAndPort()
```
anilgursel: For any new test case, I would suggest not using `temporaryServerHostnameAndPort()`, instead  set `default-listener.bind-port = 0` to prevent any port conflict related unit-test failures:

```scala
val portBindings = Await.result((Unicomplex(system).uniActor ? PortBindings).mapTo[Map[String, Int]], awaitMax)

portBindings("default-listener")
```

We have a `PortGetter` utility in `squbs-testkit`; however, not accessible from unicomplex.  So, we need to do it manually here.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/streaming/FlowDefinitionSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+    .initExtensions.start()
+
+}
+
+class TestFlowDefinition extends FlowDefinition with WebContext {
+  val firstPath = s"/$webContext/first"
+  val secondPath = s"/$webContext/second"
+  val thirdPath = s"/$webContext/third"
+
+  @volatile var count = 0
```
anilgursel: 😨   

We not only use `volatile` but also violate the http spec with a non-idempotent `GET` 😄

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/streaming/InvalidProxyFlowSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+import scala.concurrent.Await
+import scala.util.Failure
+
+object InvalidProxyFlowSpec {
+
+  val dummyJarsDir = getClass.getClassLoader.getResource("classpaths/streaming").getPath
+
+  val classPath = dummyJarsDir + "/InvalidProxyFlowSvc/META-INF/squbs-meta.conf"
+
+  val (_, _, port) = temporaryServerHostnameAndPort()
```
anilgursel: Ditto

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/streaming/RootCtxFlowSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+import org.squbs.unicomplex.{JMX, Unicomplex, UnicomplexBoot}
+
+import scala.concurrent.Await
+
+object RootCtxFlowSpec{
+
+  val dummyJarsDir = getClass.getClassLoader.getResource("classpaths/streaming").getPath
+  val classPath = dummyJarsDir + "/RootCtxFlowSpec/META-INF/squbs-meta.conf"
+
+  val (_, _, port) = temporaryServerHostnameAndPort()
```
anilgursel: Ditto

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/streaming/UnicomplexSpec.scala
```diff
@@ -155,6 +158,26 @@ class UnicomplexSpec extends TestKit(UnicomplexSpec.boot.actorSystem) with Impli
 
       Await.result(entityAsString(s"http://127.0.0.1:$port/dummysvcactor/ping"), timeout.duration) should be ("pong")
 
+      Await.result(entityAsString(s"http://127.0.0.1:$port/dummyflowsvc/ping"), timeout.duration) should be ("pong")
+
+      val requestChunks = Source.single("Hi this is a test")
+        .mapConcat { s => s.split(' ').toList }
+        .map (HttpEntity.ChunkStreamPart(_))
+
+      val errResp = Await.result(get(s"http://127.0.0.1:$port/dummyflowsvc/throwit"), timeout.duration)
```
anilgursel: Should we also have a test case with `exceptionHandler` and `rejectionHandler`?

akara:
This is not just docs. It is the same PR made to the 0.8.x branch.

docs/flow-definition.md
akara: No it is not. I pulled this code out of the test, to synthetically show response chunking.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/streaming/FlowDefinitionSpec.scala
akara: Copy of the `RouteDefinitionSpec`. But since this one can have multiple flows, the counter can incremented concurrently. That's why the `@volatile`

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/streaming/FailedFlow1Spec.scala
akara: Thanks. I was looking for that.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/streaming/UnicomplexSpec.scala
akara: Does that apply for a Flow?

docs/flow-definition.md
```diff
@@ -0,0 +1,74 @@
...
+squbs is previously basing the Http functionality on Spray. With Akka Http defining the low level request handling using `Flow[HttpRequest, HttpResponse, NotUsed]`, it is natural for squbs to support such an artifact to match with the capabilities of Akka Http moving forward.
+
+##Defining the Flow
+
+Lets jump straight into an example. To define the flow to be integrated into the request flow in squbs, you want to define a class with a public, no-argument constructor extending the `FlowDefinition` trait:
+
+```scala
+class SampleFlowSvc extends FlowDefinition with WebContext {
+
+  val prefix = if (webContext.isEmpty) "" else '/' + webContext
```
akara: Removing `WebContext` and discussing it later.

docs/flow-definition.md
akara: Looks like something odd. Thanks for noticing.

docs/flow-definition.md
akara: I've changed the code to chunk a sentence into words instead.

docs/flow-definition.md
akara: Taking it out.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/streaming/UnicomplexSpec.scala
akara: I doubt that is applicable. Please let me know if it is.

akara:
Fixed all except for two comments. The `FlowDefinitionSpec` is provided as a parallel to `RouteDefinitionSpec` with the same logic. Yes, I agree the logic is against REST rules. But if we fix it we want to fix it for both.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/streaming/UnicomplexSpec.scala
anilgursel: I understand.  If an exception mapper functionality is needed, we guide towards routing?

Also, what happens when an exception is thrown inside the flow?  The materialized request stream will be shutdown and it automatically translated to Internal Error?

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/streaming/FlowDefinitionSpec.scala
akara: Also, looking at it, I won't really see a problem using GET on a counter. In case of bidding, each GET would see an incremental number. The only problem here is that GET is "officially" not supposed to modify state. We're not alone on this, though. And I'd keep it simple. It proves the point.

PR #295, opened 2016-10-31
Issue294: Fixed exception when combining shutdownHook and internal shutdown.

akara:

- [X] Commits are squashed.
- [ ] Tests added.

Fixed exception at shutdown if the shutdown comes from internal sources. In that case, the ActorSystem would have been shut down before the shutdownhook is called. When the shutdownhook is called, it can no longer find the ActorSystem in the table. So here we check and only shut down the actor system if found. If not found, we skip - assuming it has already been shutdown.


akara:
Closing PR as repo does not seem to be in sync.

Issue #302, opened 2016-11-07
Cannot obtain port from auto-assignment if some listeners error out


akara:
In `MultiListenerSpec`, the `third-listener` has been configured with intentions of failure. If we use auto port assignment by setting `bind-port = 0`, the following ask will always timeout:

```scala
val portBindings = Await.result((Unicomplex(system).uniActor ? PortBindings).mapTo[Map[String, Int]], awaitMax)
```

This would disallow `MultiListenerSpec` from using auto port assignment. We can test this issue is fixed by setting `bind-port = 0` for all listeners and using the above ask to query the ports. If this does not timeout, the issue is fixed.

PR #321, opened 2016-11-12
#302: Fixed match error on ServerBindingInfo when dealing with bind error cases

akara:

- [X] Commits are squashed.
- [x] Tests added.

The portBindings call will only return ports successfully bound. Errors will be left out.

### DIFF
--- a/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/ServiceRegistry.scala
+++ b/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/ServiceRegistry.scala
@@ -184,9 +184,7 @@ class ServiceRegistry(val log: LoggingAdapter) extends ServiceRegistryBase[Path]
   }
 
   override private[unicomplex] def portBindings: Map[String, Int] = {
-    serverBindings map { case (name, ServerBindingInfo(Some(sb), None)) =>
-      name -> sb.localAddress.getPort
-    }
+    serverBindings collect { case (name, ServerBindingInfo(Some(sb), None)) => name -> sb.localAddress.getPort }
   }
 }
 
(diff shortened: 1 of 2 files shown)


squbs-unicomplex/src/main/scala/org/squbs/unicomplex/ServiceRegistry.scala
```diff
@@ -184,9 +184,7 @@ class ServiceRegistry(val log: LoggingAdapter) extends ServiceRegistryBase[Path]
   }
 
   override private[unicomplex] def portBindings: Map[String, Int] = {
-    serverBindings map { case (name, ServerBindingInfo(Some(sb), None)) =>
-      name -> sb.localAddress.getPort
-    }
+    serverBindings collect { case (name, ServerBindingInfo(Some(sb), None)) => name -> sb.localAddress.getPort }
```
anilgursel: Just out of curiosity, what was happening here ?

akara: The `ServerBindingInfo`'s `binding` field is set to None in case you have a binding error. Using `map` will cause a `MatchError`.

When you do

```scala
val portBindings = Await.result((Unicomplex(system).uniActor ? PortBindings).mapTo[Map[String, Int]], awaitMax)
```

this will error out with a MatchError and nothing will ever be sent back causing a timeout.

Using `collect` will give you the subset of the `serverBindings` that bound successfully and the map sent back will only have the listeners that bound properly.

akara: Actually, it would also be good to validate the map only contains 2 and the "third-listener" should not be in the map. I'll add that before you merge.

akara: Validation all added.

Issue #301, opened 2016-11-07
default-listener.local-port-header does not work with auto-assigned port


akara:
In test mode, where ports are auto-assigned, the `local-port-header` always has the configured port, the the actual auto-assigned port. We need to consider whether we need this to return the real port, and at what cost. If there is no runtime cost, we should do so.

Fixing this means `LocalPortListenerSpec` can run off of auto-assigned ports. Currently it can't.

PR #322, opened 2016-11-12
#301: Added case for local-port-header to get real/effective port from URL…

akara:

- [x] Commits are squashed.
- [x] Tests added.

…if port configured to 0 (auto-assign)

### DIFF
--- a/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/FlowHandler.scala
+++ b/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/FlowHandler.scala
@@ -153,8 +153,14 @@ class FlowHandler(routes: Seq[(Path, FlowWrapper, PipelineSetting)], localPort: 
         } getOrElse NotFound
       })
 
-      val zipF = localPort map {
-        port => (hr: HttpRequest, po: Int) => RequestContext(hr, po).addRequestHeaders(LocalPortHeader(port))
+      val zipF = localPort map { port =>
+        if (port != 0) {
+          // Port is configured, we use the configured port.
+          (hr: HttpRequest, po: Int) => RequestContext(hr, po).addRequestHeaders(LocalPortHeader(port))
+        } else {
+          // Else we use the port from the URI.
+          (hr: HttpRequest, po: Int) => RequestContext(hr, po).addRequestHeaders(LocalPortHeader(hr.uri.effectivePort))
+        }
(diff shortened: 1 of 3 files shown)


squbs-unicomplex/src/test/scala/org/squbs/unicomplex/LocalPortListenerSpec.scala
```diff
@@ -30,8 +30,8 @@ import scala.concurrent.Await
 
 object LocalPortListenerSpecActorSystem {
 
-  // We need to explicitly generate a port for this test as the local port header takes the port form the config
-  // and not the binding results.
+  // Case 1 default-listener: Explicitly generate a port for this test. The local port header takes the port form the config.
+  // Case 2 fourth-listener: Implicitly ask Unicomplex to create a port.
```
akara: Note we keep port1 manually assigned so we can have the test coverage for both auto and manual assigned use cases. They have slightly different code paths.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/FlowHandler.scala
```diff
@@ -153,8 +153,11 @@ class FlowHandler(routes: Seq[(Path, FlowWrapper, PipelineSetting)], localPort:
         } getOrElse NotFound
       })
 
-      val zipF = localPort map {
-        port => (hr: HttpRequest, po: Int) => RequestContext(hr, po).addRequestHeaders(LocalPortHeader(port))
+      val zipF = localPort map { port =>
+        if (port != 0)
```
anilgursel: It would be great to add a short comment here, why we check.  

Also, if the body is not in the same line, shouldn't we be using curly braces?

anilgursel:
Just a minor styling comment, otherwise LGTM.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/FlowHandler.scala
akara: This is the scalastyle we use for this case:

```xml
 <check level="warning" class="org.scalastyle.scalariform.IfBraceChecker" enabled="false">
  <parameters>
   <parameter name="singleLineAllowed"><![CDATA[true]]></parameter>
   <parameter name="doubleLineAllowed"><![CDATA[true]]></parameter>
  </parameters>
 </check>
```

Issue #122, opened 2015-10-28
Extract common components from squbs-httpclient project 


anilgursel:
`EnvironmentResolver` has uses cases outside of http client.  So, we should extract/move it to another project.

PR #334, opened 2016-11-18
#122 Extract common components to a separate project

anilgursel:

- [x] Commits are squashed.
- [x] Tests added.

Description:
- Moving functionalities that are not specific to httpclient out of squbs-httpclient project:
   - Metrics
   - Endpoint Resolver
   - Environment Resolver
   - Circuit Breaker metrics
- `ConfigUtil` moves to `squbs-ext` so that it can be used by other projects without a unicompelx dependency.
- Http client does not depend on unicomplex anymore https://github.com/paypal/squbs/issues/308.
- Metrics need akka-http-core; however, marking it as `"provided"` to make sure it does not bring the dependency to classpath as `squbs-ext` can be used by non-http projects (Endpoint resolver etc).  Http projects (e.g., squbs-httpclient) should specify the http-core dependency.
- Same for dependency on `squbs-streamingpipeline`.  We do not want to bring http transitively, so dependency on `squbs-streamingpipeline` is `"provided"`.
- Given that we extracted Metrics to `squbs-ext`, this allowed me to add metrics tests for server side https://github.com/paypal/squbs/issues/113.


akara:
Just a question. Did you cover serializers/deserializers?

anilgursel:
No, this PR does not cover serializers and deserializers.  We have https://github.com/paypal/squbs/issues/329 to tackle that.

anilgursel:
Looks like my other PR caused a merge conflict..  Let me resolve it..

anilgursel:
But, actually, let me get your comments and address those, afterwards I will rebase and resolve the conflicts.  Otherwise, there will be multiple commits.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/pipeline/MetricsFlowSpec.scala
```diff
@@ -0,0 +1,204 @@
...
+  val port = portBindings("default-listener")
+  val poolClientFlow = Http().cachedHostConnectionPool[Int]("localhost", port)
+
+  it should "collect request count and time metrics when webContext is empty" in {
+    val future = Source(hello("") :: hello("") :: Nil).
+      via(poolClientFlow).
+      runWith(Sink.ignore)
+
+    future map { _ =>
+      assertJmxValue("/-request-count", "Count", 2)
```
akara: Ditto here. The assertion function is harder to read than the matcher DSL. Below, too.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/pipeline/MetricsFlowSpec.scala
```diff
@@ -0,0 +1,204 @@
...
+      runWith(Sink.ignore)
+
+    future map { _ =>
+      assertJmxValue("sample6-request-count", "Count", 6)
+      assertJmxValue("sample6-2XX-count", "Count", 2)
+      assertJmxValue("sample6-5XX-count", "Count", 4)
+    }
+  }
+
+  def assertJmxValue(beanName: String, key: String, expectedValue: Any) = {
```
akara: Should we add these to some test utils? Of course, not the `assertJmxValue` but rather `jmxEntry(beanName: String, key: String): Option[AnyRef]` and re-use it across tests?

squbs-streamingpipeline/src/main/scala/org/squbs/pipeline/streaming/PipelineExtension.scala
```diff
@@ -86,8 +86,8 @@ object PipelineExtension extends ExtensionId[PipelineExtensionImpl] with Extensi
       flowMap = flowMap + (name -> flowFactory)
     }
 
-    val pre = system.settings.config.getOptionalString("squbs.pipeline.streaming.defaults.pre-flow")
-    val post = system.settings.config.getOptionalString("squbs.pipeline.streaming.defaults.post-flow")
+    val pre = Try(system.settings.config.getString("squbs.pipeline.streaming.defaults.pre-flow")).toOption
+    val post = Try(system.settings.config.getString("squbs.pipeline.streaming.defaults.post-flow")).toOption
```
anilgursel: Yes, I would.  But, `ConfigUtil` is defined in `squbs-extra` which depends on `squbs-streamingpipelne` because of the `MetricFlow`.  I do not want to create a cyclic dependency for `ConfigUtil`.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/pipeline/MetricsFlowSpec.scala
anilgursel: Thought about it.  But, need to pass domain name, system etc..  A few lines of function requires long arguments, I was not sure which one would be more readable..

squbs-ext/src/test/scala/org/squbs/metrics/MetricsFlowSpec.scala
```diff
@@ -0,0 +1,140 @@
...
+      assertJmxValue("sample6-RequestTimeoutException-count", "Count", 2)
+    }
+  }
+
+  def assertJmxValue(beanName: String, key: String, expectedValue: Any) = {
+    val oName =
+      ObjectName.getInstance(s"${MetricsExtension(system).Domain}:name=${MetricsExtension(system).Domain}.$beanName")
+    val actualValue = ManagementFactory.getPlatformMBeanServer.getAttribute(oName, key)
+    actualValue shouldEqual expectedValue
+  }
```
anilgursel: 👍 

I do not remember what was my reasoning going the other way..  but, I absolutely agree, this is cleaner..

Instead of `jmxEntry` and I am naming it as `jmxValue` though..

squbs-streamingpipeline/src/main/scala/org/squbs/pipeline/streaming/PipelineExtension.scala
akara: Got it. NP.

PR #336, opened 2016-11-23
#8 Http Client Documentation

anilgursel:

- [x] Commits are squashed.
- [ ] Tests added.


anilgursel:
> The deleted docs contain examples for GET, POST, PUT, HEAD. We may need to ensure having such samples here, too.

I disagree.  There is nothing squbs specific there, so I would point to Akka HTTP documentation for such samples.  My goal is to make sure the doc looks simple and give the message that it is no different than Akka HTTP API.

docs/httpclient.md
```diff
@@ -1,1575 +1,118 @@
...
-		- [MarkDown HttpClient](#markdown-httpclient)
-		- [MarkUp HttpClient](#markup-httpclient)
-		- [Close HttpClient](#close-httpclient)
-		- [Use HttpClient to Make HTTP Calls](#use-httpclient-to-make-http-calls)
-		- [Use HttpClient to Make HTTP Calls and Return Unmarshalled Objects](#use-httpclient-to-make-http-calls-and-return-unmarshalled-objects)
-	- [Google API Sample](#google-api-sample-1)
-		- [Main (Message Based API)](#main-message-based-api)
+* [Service Discovery](#service-discovery-chain): Lets any service discovery mechanism to be plugged in and allows resolving endpoints by string identifiers, e.g., `paymentserv`.
+* [Per Client configuration](#per-client-configuration): Let's each client to individually override defaults in `application.conf`.
+* [Pipeline](#pipeline): Allows a `Bidi`Akka Streams flow to be registered globally or individually for clients.
```
anilgursel: 👍

akara:
While I agree that most should be in the Akka Http docs, I want to have adequate samples so people won't have to flip over. It turns out they have to compose the knowledge from the two places by themselves. It would be better to have some good amount of examples, not totally complete. Then point them to Akka Http for further detail. I'm not talking about the amount we had before. 😄

anilgursel:
I am pretty picky and concise when it comes to docs.  There will already be enough content on the page when we add docs for more features (e.g., Circuit Breaker, serialization/deserialization, passing different options to `ClientFlow`).  Earlier, it was needed to have samples `GET`, `POST`, etc because we had an API and we were making a different function call.  With Akka HTTP, it is all about setting the method in `HttpRequest` object and the rest of the stream looks exactly same.

I would like to be very careful on the message on the doc.  This is still Akka HTTP API with extra capabilities.  The project does not need any extra dependency, just an actor system.  So, you can use `squbs-httpclient` by itself.  The more content, the more complicated it would look.  I believe we will already have a long content when all features added.

akara:
I really feel even the Akka documentation seems to lack information on constructing a `HttpRequest` object. Of course, we can figure this out. It doesn't mean our users can figure it out.
My thought is just a section with examples of how to construct the HttpRequest object alone.
We can talk about this off-line.

akara:
LGTM - you want me to wait for Circuit Breaker to merge? Or do you want to take care of that as part of the Circuit Breaker PR.

anilgursel:
I wanna do it with the Circuit Breaker PR.  Currently, I just want to convey that it is on the way.

PR #339, opened 2016-11-28
#328: Added Java API support for Flow and Route.

akara:

- [X] Commits are squashed.
- [x] Tests added.

Also restructured the documentation to have a separate page discussing services and simplified the bootstrap documentation to reference the service documentation.


akara:
There are still some test coverage gaps. Will need to add a few more tests before merging. But please review the code and docs so far.

docs/http-services.md
```diff
@@ -0,0 +1,224 @@
...
+### Listener Binding
+
+A listener is declared in `application.conf` or `reference.conf` usually living in the project's `src/main/resources` directory. Listeners declare interfaces, ports, https security attributes, and name aliases, and are explained in [Configuration](configuration.md)
+
+A service handler attaches itself to one or more listeners. The `listeners` attribute is a list of listeners or aliases the handler should bind to. If listeners are not defined, it will default to the `default-listener`.
+
+The wildcard value `"*"` (note, it has to be quoted or will not be properly be interpreted) is a special case which translates to attaching this handler to all active listeners. By itself, it will however not activate any listener if it is not already activated by a concrete attachment of a handler. If the handler should activate the default listener and attach to any other listener activated by other handlers, the concrete attachment has to be specified separately as follows:
+
+```
+    listeners = [ default-listener, "*" ]
```
anilgursel: I think we need to restructure  this document.  It feels too mechanical and does not explain the motivation and also the ordering looks not so correct.  It other words, it is not flowing fluently.  I would suggest to have the structure as:

- Overview - this talks about the motivation.  Things like, auto-metrics enablement, pipeline, standardization.  Also, emphasizes sticking to Akka HTTP API.  
- Dependencies - what dependency needed to use these capabilities.
- Usage - show a sample Route API and immediately after show how to register it in squbs-meta.conf
- Then, talk about webcontext
- Talk about listener binding.
- Low level API sample
- High Level Java API
- Low level Java API

If we can make the examples shorter and make each as sub-sections, then all the different flavors of the APIs can be back to back as well (before talking about registering in squbs-meta.conf)..  But, I believe it is important to show the high level API first, because low level looks more complex, especially we need to specify webcontext etc..

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
```diff
@@ -317,10 +317,12 @@ object UnicomplexBoot extends LazyLogging {
...
       Try {
         (clazz asSubclass classOf[RouteDefinition], classOf[RouteActor])
-      } orElse {
-        Try {
-          (clazz asSubclass classOf[FlowDefinition], classOf[FlowActor])
-        }
+      } orElse Try {
+        (clazz asSubclass classOf[FlowDefinition], classOf[FlowActor])
+      } orElse Try {
+        (clazz asSubclass classOf[AbstractRouteDefinition], classOf[JavaRouteActor])
```
anilgursel: Looks something encrypted here :)

squbs-unicomplex/src/test/java/org/squbs/unicomplex/javaflowsvc/JavaFlowSvc.java
```diff
@@ -0,0 +1,67 @@
...
+import akka.http.javadsl.model.*;
+import akka.stream.javadsl.Flow;
+import akka.stream.javadsl.Source;
+import akka.util.ByteString;
+import org.squbs.unicomplex.AbstractFlowDefinition;
+import scala.Tuple2;
+
+public class JavaFlowSvc extends AbstractFlowDefinition {
+
+    private final String pingPath;
```
anilgursel: Does this line compile?

squbs-unicomplex/src/test/java/org/squbs/unicomplex/javaflowsvc/JavaFlowSvc.java
```diff
@@ -0,0 +1,67 @@
...
+    private final String chunkPath;
+    private final String exceptionPath;
+
+    public JavaFlowSvc() {
+        String prefix = "";
+        if (!webContext().isEmpty()) {
+            prefix = '/' + webContext();
+        }
+        pingPath = prefix + "/ping";
```
anilgursel: indentation?

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/JavaFlowSvcSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+import scala.concurrent.Await
+
+object JavaFlowSvcSpec {
+
+  val classPaths = Array(getClass.getClassLoader.getResource("classpaths/JavaFlowSvc").getPath)
+
+  val config = ConfigFactory.parseString(
+    s"""
+       |default-listener.bind-port = 0
+       |squbs {
```
anilgursel: indentation?  Also, shouldn't it be lower case ?

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/JavaFlowSvcSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  val portBindings = Await.result((Unicomplex(system).uniActor ? PortBindings).mapTo[Map[String, Int]], awaitMax)
+  val port = portBindings("default-listener")
+
+  override def afterAll() {
+    Unicomplex(system).uniActor ! GracefulStop
+  }
+
+  it should "handle a normal request" in {
+    Await.result(entityAsString(s"http://127.0.0.1:$port/javaflowsvc/ping"), awaitMax) shouldBe "pong"
```
anilgursel: Can use `AsyncFlatspec` to eliminate Await.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/JavaRouteSvcSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+import scala.concurrent.Await
+
+object JavaRouteSvcSpec {
+
+  val classPaths = Array(getClass.getClassLoader.getResource("classpaths/JavaRouteSvc").getPath)
+
+  val config = ConfigFactory.parseString(
+    s"""
+       |default-listener.bind-port = 0
+       |squbs {
```
anilgursel: indentation and squbs should be lower case

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/JavaRouteSvcSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+       |  actorsystem-name = JavaRouteSvcSpec
```
anilgursel: not sure "Actorsystem-name" is correct.

docs/http-services.md
```diff
@@ -0,0 +1,224 @@
...
+    
+    # Optional, only applies to actors.
+    init-required = false
+  }
+]
+```
+
+The class-name parameter identifies the service definition class which could use either the low-level or high-level API and could be implemented in either Java or Scala.
+
+The web-context is a string that uniquely identifies the web context of a request to be dispatched to this service. It **MUST NOT** start with a `/` character. It can have `/` characters inside as segment separators in case of multi-segment contexts. And it is allowed to be `""` for root context. If multiple services match the request, the longest context match takes precedence.
```
akara: That's what it is right now. File an issue :smile:

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
akara: Huh?

squbs-unicomplex/src/test/java/org/squbs/unicomplex/javaflowsvc/JavaFlowSvc.java
akara: Of course! Wouldn't have passed the build otherwise.

If you declare something as `final`, you can either assign a value here or in the constructor. In this case we assigned in the constructor.

squbs-unicomplex/src/test/java/org/squbs/unicomplex/javaflowsvc/JavaFlowSvc.java
akara: Sorry, I don't see anything wrong with this indentation.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/JavaRouteSvcSpec.scala
akara: Strange! It is all lower case. I have not changed anything.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/JavaRouteSvcSpec.scala
akara: Here too, I read `actorsystem-name`. Anything wrong with your browser rendering?

docs/http-services.md
```diff
@@ -0,0 +1,224 @@
...
+
+**Note:** Unlike programming to Akka Http directly, squbs provides all socket binding and connection management. Just provide the request/response handling through the defined interfaces and register those implementations to squbs.
+
+##The Web Context
+
+Each service entry point is bound to a unique web context which is the leading path segments separated by the `/` character. For instance, the url `http://mysite.com/my-context/index` would match the context `"my-context"`, if registered. It can also match the root context if `"my-context"` is not registered. Web contexts are not necessarily the first slash-separated segment of the path. Dependent on the context registration, it may match multiple such segments. A concrete example would be a URL with service versioning. The URL `http://mysite.com/my-context/v2/index` may have either `my-context` or `my-context/v2` as the web context, depending on what contexts are registered. If both `my-context` and `my-context/v2` are registered, the longest match - in this case `my-context/v2` will be used for routing the request.
```
akara: This was actually a feature request before the split. It allows v1 and v2 to have totally different resources or even different service cubes. I see that as a good feature. Now how it is written in the docs may be confusing. That's a different issue that we'll have to address.

docs/http-services.md
akara: Lets have a short discussion before I make this change. Agree with most comments. Only one point. What we have here is essentially the Usage section, which consists of short samples of the different implementations, then the registration. We don't go much into detail on any of these but point to the relevant Akka docs. I feel we should do these in one shot instead of a separate Usage section that only shows one of the Scala examples.

Since sections are interdependent, it is a bit hard to outline this without circular dependencies.

docs/http-services.md
```diff
@@ -0,0 +1,224 @@
...
+import akka.stream.scaladsl.Flow
+import org.squbs.unicomplex.FlowDefinition
+
+class SampleFlowSvc extends FlowDefinition {
+
+  def flow = Flow[HttpRequest].map {
+    case HttpRequest(_, Uri(_, _, Path("ping"), _, _), _, _, _) =>
+      HttpResponse(StatusCodes.OK, entity = "pong")
+    case _ =>
+      HttpResponse(StatusCodes.NotFound, entity = "Path not found!")
```
anilgursel: Hmm.  I recall it wrong then..  I thought an exception would be thrown and the default exception mapper would be mapping to 404.

docs/http-services.md
anilgursel: 👍 

Just created https://github.com/paypal/squbs/issues/340.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
anilgursel: Looks ok now.  Don't know what's going on here.  I remember seeing a class name which was not proper words.

squbs-unicomplex/src/test/java/org/squbs/unicomplex/javaflowsvc/JavaFlowSvc.java
anilgursel: Don't really know what's going on..  It used to be written as `Private Final String pingPath;`..  I mean, keywords were starting with capital letters.

squbs-unicomplex/src/test/java/org/squbs/unicomplex/javaflowsvc/JavaFlowSvc.java
anilgursel: Looks ok.  Next time I review, I will make sure copy pasting the code or taking a screen shot.  It is apparent that Github was playing with me while I was reviewing it 😄

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/JavaFlowSvcSpec.scala
anilgursel: I am not suggesting to change existing tests.  But, any new tests we add should use `AsyncFlatspec`.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/JavaRouteSvcSpec.scala
anilgursel: Same comments..  It looks perfectly ok now..  There was apparently some issue earlier in rendering 😡

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/JavaRouteSvcSpec.scala
anilgursel: Looks ok now.  But, during review time, it was showing as `Actorsystem-name`.  Strange..

anilgursel:
The indentation and capitalization issue seems to be looking fine now.  Apparently, there was a rendering issue during my review..

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/JavaFlowSvcSpec.scala
akara: OK. Will do - this afternoon.

akara:
I'm sure there will still be more doc modifications. But from code and test standpoint, I'm all good. Once docs are clear I'll do a final squash before merge.

docs/http-services.md
```diff
@@ -0,0 +1,252 @@
...
+
+Optionally, the listeners parameter declares a list of listeners to bind this service. Listener binding is discussed in the following section, below.
+
+The pipeline is a set of request pre-processors before the request gets processed by the request handler. The pipeline name can be specified by a `pipeline` parameter. A configured default pipeline will be used. To disable the default pipeline for this service, you can set `defaultPipelineOn = false` in META-INF/squbs-meta.conf. Please refer to [Streaming Request/Response Pipeline](streamingpipeline.md) for more information.
+
+Only actors can have another optional `init-required` parameter which allows the actor to feed back its state to the system. Please refer to the [Startup Hooks](lifecycle.md#startup-hooks) section of the [Runtime Lifecycles & API](lifecycle.md) documentation for a full discussion of startup/initialization hooks.
```
anilgursel: I believe this paragraph adds some confusion.  I'd rather delete this paragraph.  Also, since we have the low level API support, we can deprecate the http actors soon.  And the well known actors are already discussed in a different documentation.

docs/http-services.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Listener Binding
+
+A listener is declared in `application.conf` or `reference.conf` usually living in the project's `src/main/resources` directory. Listeners declare interfaces, ports, Https security attributes, and name aliases, and are explained in [Configuration](configuration.md)
```
anilgursel: Just thinking out loud..  It might be useful to add a sentence or two about standardization here..  I mean the motivation behind listeners.

docs/http-services.md
```diff
@@ -0,0 +1,252 @@
...
+                });
+    }
+}
+```
+
+Note: The `webContext()` method as well as the `context()` method for accessing the actor context are provided by the `AbstractFlowDefinition` class.
+
+This provides access to the `Flow` representation of the Akka Http low-level server-side API. Please refer to the [Akka HTTP low-level API](http://doc.akka.io/docs/akka-http/current/java/http/server-side/low-level-server-side-api.html#streams-and-http), [Akka Streams](http://doc.akka.io/docs/akka/current/java/stream/stream-quickstart.html), and the [Http model](http://doc.akka.io/docs/akka-http/current/java/http/http-model.html#http-model-java) documentation for further information on constructing more sophisticated `Flow`s.
+
+**Note:** Unlike programming to Akka HTTP directly, squbs provides all socket binding and connection management with its listeners. Just provide the request/response handling through one or more of the above interfaces and register those implementations to squbs.
```
anilgursel: Instead of having this paragraph, should we instead cover this under listeners section?

docs/http-services.md
```diff
@@ -0,0 +1,252 @@
...
+
+squbs supports Akka HTTP, both the low-level and high-level server-side APIs, for defining services. Both APIs enjoy the full productionalization support such as listeners, pipelines, logging, and monitoring. In addition, squbs supports both Scala and Java flavors of service definitions. These service handlers are declared in classes and registered to squbs in the metadata through the `META-INF/squbs-meta.conf` file through the `squbs-services` entry in this fil. Each style of service is registered in the same manner, just by providing the class name and configuration.
+
+All squbs service definitions have access to the field `context` which is an `akka.actor.ActorContext` useful for accessing the actor system, scheduler, and a variety of Akka facilities.
+
+##Dependencies
+
+squbs server side request handling is directly coupled to the Unicomplex. This is needed for starting the server as well as registering service definitions. The required dependency is:
```
anilgursel: This sentence might be perceived negatively.  We can instead just say:

"The following dependency is needed for starting the server as well as registering service definitions:"

docs/http-services.md
```diff
@@ -0,0 +1,252 @@
...
+    public Optional<ExceptionHandler> exceptionHandler() {
+        return Optional.of(ExceptionHandler.newBuilder()
+                .match(ServiceException.class, se ->
+                        complete("exception"))
+                .build());
+    }
+}
+```
+
+In addition to defining the `route`, you can also provide a `RejectionHandler` and an `ExceptionHandler` by overriding the `rejectionHandler` and `exceptionHandler` methods accordingly. These can be seen in the example above.
```
akara: Already there. Just in the paragraph below.

docs/http-services.md
```diff
@@ -0,0 +1,252 @@
...
+      HttpResponse(StatusCodes.OK, entity = "pong")
+    case _ =>
+      HttpResponse(StatusCodes.NotFound, entity = "Path not found!")
+  }
+}
+```
+
+##Concurrency on the High-Level Route API
+
+The provided `route` can be used by multiple connections, and therefore threads, concurrently. If the `route` accesses any state in the encapsulating `RouteDefinition` (Scala) or `AbstracRouteDefinition` (Java) class, it is important to note such access can be concurrent, both for reads and writes. It is not safe for such accesses to read or write mutable state inside the encapsulating class. The use of Akka `Actor`s or `Agent`s is highly encouraged in such situations.
```
akara: Yes we do. But that only covers the `flow` API. We need to also make the expectations clear for the `route` API.  Either we call it out that the bullet 3 also applies to route, or we have a separate section covering the route API.

docs/http-services.md
```diff
@@ -0,0 +1,252 @@
...
+
+##Rules and Behaviors of the Low-Level Flow API
+
+There are a few rules you have to keep in mind when implementing a `FlowDefinition` (Scala) or `AbstractFlowDefinition` (Java):
+
+1. **Exactly one response:** It is the responsibility of the application to generate exactly one response for every request.
+2. **Response ordering:** The ordering of responses matches the ordering of the associated requests (which is relevant if HTTP pipelining is enabled where processing of multiple incoming requests may overlap).
```
akara: How could we have done that? We can only do so when we have the whole `RequestContext`. But when `Flow`s give back  responses, there is not enough data - no request association - in the response, right?

docs/http-services.md
```diff
@@ -0,0 +1,252 @@
...
+3. **Concurrent state access:** The flow can be materialized multiple times, causing multiple instances of the `Flow` itself. If they access any state in the `FlowDefinition` or `AbstracFlowDefinition`, it is important to note such access can be concurrent, both for reads and writes. It is not safe for such accesses to read or write mutable state inside the encapsulating class. The use of Akka `Actor`s or `Agent`s is highly encouraged in such situations.
+4. **Access to actor context:** The `FlowDefinition`/`AbstractFlowDefinition` has access to the `ActorContext` with the `context` field (Scala) or `context()` method (Java) by default. This can be used to create new actors or access other actors.
```
akara: Yes.

docs/http-services.md
```diff
@@ -0,0 +1,252 @@
...
+5. **Access to web context:** For the Scala `FlowDefinition`, if the `WebContext` trait is mixed in, it will have access to the field `webContext`. The Java `AbstractFlowDefinition` provides the `webContext()` method in all cases. This field/method is used to determine the web context or path from the root where this `FlowDefinition`/`AbstractFlowDefinition` is handling requests.
```
akara: You're right. I better write up the rules for both in separate sections.

docs/http-services.md
```diff
@@ -0,0 +1,252 @@
...
+4. **Access to actor context:** The `FlowDefinition`/`AbstractFlowDefinition` has access to the `ActorContext` with the `context` field (Scala) or `context()` method (Java) by default. This can be used to create new actors or access other actors.
+5. **Access to web context:** For the Scala `FlowDefinition`, if the `WebContext` trait is mixed in, it will have access to the field `webContext`. The Java `AbstractFlowDefinition` provides the `webContext()` method in all cases. This field/method is used to determine the web context or path from the root where this `FlowDefinition`/`AbstractFlowDefinition` is handling requests. 
+6. **Request path:** The `HttpRequest` object is handed to this flow unmodified. The `webContext` is in the `Path` of the request. It is the users job (as seen above) to handle the request with the knowledge of the `webContext`. In other words, the low-level API handles the `HttpRequest` directly and needs to manually take the web context into consideration for any path matching.
+
+
+##Service Registration
```
akara: I could have guessed I'll get this comment 😄

akara: Moving it.

docs/http-services.md
akara: Moved.

docs/http-services.md
```diff
@@ -0,0 +1,252 @@
...
+    
+    # Optional, only applies to actors.
+    init-required = false
+  }
+]
+```
+
+The class-name parameter identifies the service definition class which could use either the low-level or high-level API and could be implemented in either Java or Scala.
+
+The web-context is a string that uniquely identifies the web context of a request to be dispatched to this service. It **MUST NOT** start with a `/` character. It can have `/` characters inside as segment separators in case of multi-segment contexts. And it is allowed to be `""` for root context. If multiple services match the request, the longest context match takes precedence.
```
akara: As I moved this section up, it will be before the web context section. So this would be a forward reference. We need to keep some short explanation here. Perhaps next round after moving it we can revisit this again.

docs/http-services.md
akara: Yes, remnants of copying the text over.

docs/http-services.md
```diff
@@ -0,0 +1,251 @@
...
+  override def exceptionHandler: Option[ExceptionHandler] =
+    Some(ExceptionHandler {
+      case _: ServiceException => complete("exception")
+    })
+}
+```
+
+In addition to defining the `route`, you can also provide a `RejectionHandler` and an `ExceptionHandler` by overriding the `rejectionHandler` and `exceptionHandler` functions accordingly. These can be seen in the example above.
+
+Please refer to the [Akka HTTP high-level API](http://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/index.html), [Routing DSL](http://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/overview.html), [Directives](http://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/directives/index.html), [Rejection](http://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/rejections.html), and [Exception Handling](http://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/exception-handling.html) documentation to fully utilize these API.
```
anilgursel: > ...  to fully utilize these API.

these APIs ?

docs/http-services.md
anilgursel: u're right.  I wish there was a way for us add something to `HttpRequest` that automatically gets propagated to `HttpResponse`.

docs/http-services.md
anilgursel: Sorry, my comment was not clear enough.  I meant we need to mention `RouteDefinition` as well.  This bullet was mentioning only flow based API.  But, it looks like you created a separate section for high level api.

docs/http-services.md
anilgursel: I would suggest moving most of this paragraph to the webcontext section and give a link to it here.  Very similar to what we do for listener binding.  For instance, this paragraph might contain:

The `web-context` is a string that uniquely identifies the web context of a request to be dispatched to this service.  `web-context` is discussed in [The Web Context](the-web-context) section, below.

Issue #331, opened 2016-11-15
System state should not be active with an invalid squbs-service class-name


anilgursel:
To reproduce:

- Create an Akka Http app
- In `squbs-meta.conf`, change class-name to an invalid one.  
- Start the service 

You see an exception in the logs; however, system state turns active and it starts.

Expectation: System state should be failed.

PR #352, opened 2016-12-13
#331: Ensure a ClassNotFoundException of actors or services results in a system startup failure.

akara:

- [X] Commits are squashed.
- [X] Tests added.

### DIFF
--- a/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/Unicomplex.scala
+++ b/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/Unicomplex.scala
@@ -23,6 +23,7 @@ import java.util.Date
 import java.util.concurrent.ConcurrentHashMap
 
 import akka.NotUsed
+import akka.actor.Actor.Receive
 import akka.actor.SupervisorStrategy._
 import akka.actor.{Extension => AkkaExtension, _}
 import akka.agent.Agent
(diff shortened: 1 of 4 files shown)


squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
```diff
@@ -475,8 +474,8 @@ object UnicomplexBoot extends LazyLogging {
     logger.info(s"Web Service started in $elapsed milliseconds")
   }
 
-  private[unicomplex] def getRootCause(e: Exception) = {
-    var t: Throwable = e
+  private[unicomplex] def getRootCause(e: Throwable) = {
+    var t = e
```
anilgursel: Just to eliminate to `var` usage, should we instead do something like 

```scala
def getRootCause(e: Throwable): Throwable = {
  if(e == null) null
  else Option(getRootCause(e.getCause)).getOrElse(e)
}
```

I know this is not tail recursive, but it would be highly unlikely to have stack over flow during startup because of the exception chain being long.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/Unicomplex.scala
```diff
@@ -741,3 +750,10 @@ class CubeSupervisor extends Actor with ActorLogging with GracefulStopHelper {
   }
 }
 
+/**
+  * A noop actor used when a stub ActorRef is needed to register failures.
+  */
+class NoopActor extends Actor {
```
akara: Yes. Good point. I'll make it private. Thanks!

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
akara: Let me try out.

anilgursel: Even better.

PR #360, opened 2016-12-17
#348: Added Java API to ClientFlow and updated resolver registration API

akara:

- [X] Commits are squashed.
- [X] Tests added.


docs/httpclient.md
```diff
@@ -29,18 +29,31 @@ implicit val system = ActorSystem()
...
 val poolClientFlow = ClientFlow[Int]("sample") // Only this line is specific to squbs
+
 val responseFuture: Future[(Try[HttpResponse], Int)] =
   Source.single(HttpRequest(uri = "/") -> 42)
     .via(poolClientFlow)
     .runWith(Sink.head)
 ```
 
-For Java usage of the above example, please see [Akka HTTP Host-Level Client-Side Java API](http://doc.akka.io/docs/akka-http/current/java/http/client-side/host-level.html#example).  The squbs specific line above would be written as:
+The Java usage of the above example is as followings:
```
anilgursel: Not sure why I hadn't done it earlier.  But, to align better with the rest of the docs and also with the rest of this doc itself, it would be good to add sub section with `##### Java`.

We should also do the same for `##### Scala` in line 24, just before the following sentence:

> Here is an example usage that looks very similar to [Akka HTTP Host-Level Client-Side API](http://doc.akka.io/docs/akka-http/current/scala/http/client-side/host-level.html#example):

docs/httpclient.md
```diff
@@ -29,18 +29,31 @@ implicit val system = ActorSystem()
...
 
 ```java
-// Java Usage
-ClientFlow.get(system).<Integer>create("sample", materializer);
+final ActorSystem system = ActorSystem.create();
+final ActorMaterializer mat = ActorMaterializer.create(system);
+
+final Flow<Pair<HttpRequest, Integer>, Pair<Try<HttpResponse>, Integer>, HostConnectionPool>
+    clientFlow = ClientFlow.create("sample", system, mat);
```
anilgursel: 👍

docs/httpclient.md
```diff
@@ -95,7 +125,44 @@ class SampleEndpointResolver extends EndpointResolver {
 // Register EndpointResolver
 EndpointResolverRegistry(system).register(new SampleEndpointResolver)
 ```
- 
+
```
anilgursel: Oh boy, the http client documentation is again getting longer 😟 But, no where near to the earlier versions.  I can clearly see that it is to the point and we need every one of these examples, so no action needed.  But, if it gets longer in the future, we can consider it to split it to scala vs java as Akka HTTP does.  But, we are not there yet.

docs/httpclient.md
```diff
@@ -79,7 +92,24 @@ HttpRequest complexRequest =
...
+`squbs-httpclient` does not require hostname/port combination to be provided during client pool creation.  Instead it allows a service discovery chain to be registered which allows resolving endpoints by a string identifier by running through the registered service discovery mechanisms.  For instance, in the above example, `"sample"` is the logical name of the service that client wants to connect, the configured service discovery chain will resolve it to an endpoint which includes host and port, e.g., `http://akka.io:80`.
+
+There are two variations of registering endpoint resolvers as shown below. The closure style allows more compact and readable code. However, the subclass has the power to keep state and make resolution decisions based on such state:
+
+##### Scala
+
+Register function type `(String, Env) => Option[Endpoint]`
```
anilgursel: `:` at the end of the line?

docs/httpclient.md
```diff
@@ -79,7 +92,24 @@ HttpRequest complexRequest =
...
+```scala
+EndpointResolverRegistry(system).register("SampleEndpointResolver", { (svcName, env) =>
+  svcName match {
+    case "sample" => Some(Endpoint("http://akka.io:80"))
+    case "google" => Some(Endpoint("http://www.google.com:80"))
+    case _ => None
+})
+```
+
+Register class extending `EndpointResolver`
```
anilgursel: `:` at the end of the line?

docs/httpclient.md
```diff
@@ -95,7 +125,44 @@ class SampleEndpointResolver extends EndpointResolver {
 // Register EndpointResolver
 EndpointResolverRegistry(system).register(new SampleEndpointResolver)
 ```
- 
+
+##### Java
+
+Register `BiFunction<String, Env, Optional<Endpoint>>`
```
anilgursel: `:` at the end of the line?

docs/httpclient.md
```diff
@@ -95,7 +125,44 @@ class SampleEndpointResolver extends EndpointResolver {
...
+    } else if ("google".equals(svcName))
+        return Optional.of(Endpoint.create("http://www.google.com:80"));
+    } else {
+        return Optional.empty();
+    }
+});
+
+```
+
+Register class extending `AbstractEndpointResolver`
```
anilgursel: `:` at the end of the line?

docs/httpclient.md
```diff
@@ -95,7 +125,44 @@ class SampleEndpointResolver extends EndpointResolver {
...
+
+```java
+class SampleEndpointResolver extends AbstractEndpointResolver {
+    String name = "SampleEndpointResolver"
```
anilgursel: In scala version, this is a method?  If we would like to keep it as a field, then we should do `private` (in that case it would not hide the one from parent though), or `protected`.

docs/httpclient.md
```diff
@@ -95,7 +125,44 @@ class SampleEndpointResolver extends EndpointResolver {
...
 EndpointResolverRegistry(system).register(new SampleEndpointResolver)
 ```
- 
+
+##### Java
+
+Register `BiFunction<String, Env, Optional<Endpoint>>`
+
+```java
+EndpointResolverRegistry.get(system).register("LocalhostHttpsEndpointResolver", (svcName, env) -> {
```
anilgursel: "SampleEndpointResolver"?

squbs-ext/src/main/scala/org/squbs/endpoint/Endpoint.scala
```diff
@@ -48,19 +49,79 @@ object Endpoint {
...
+  }
+}
+
 class EndpointResolverRegistryExtension(system: ExtendedActorSystem) extends Extension with LazyLogging {
-  var endpointResolvers = List[EndpointResolver]()
 
-  def register(resolver: EndpointResolver) {
+  private[endpoint] var endpointResolvers = List.empty[EndpointResolver]
+
+  /**
```
anilgursel: Great job!  It is a good reminder that we should start paying more attention on the ScalaDocs/JavaDocs.

PR #363, opened 2016-12-23
#362: Brought up code coverage for Java APIs and more

akara:

- [X] Commits are squashed.
- [X] Tests added.

### DIFF
--- a/squbs-ext/src/test/java/org/squbs/endpoint/EndpointTest.java
+++ b/squbs-ext/src/test/java/org/squbs/endpoint/EndpointTest.java
@@ -1,0 +1,155 @@
+/*
+ *  Copyright 2015 PayPal
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.squbs.endpoint;
+
+import akka.actor.ActorSystem;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Test;
+import org.squbs.env.Default;
+import org.squbs.env.EnvTestHelper;
+import org.squbs.env.Environment;
+import scala.collection.immutable.List;
+import scala.Option;
+
+import javax.net.ssl.SSLContext;
+import java.security.NoSuchAlgorithmException;
+import java.util.Optional;
+
+import static org.junit.Assert.*;
+
+public class EndpointTest {
+
+    private static final ActorSystem system = ActorSystem.create("EndpointTest");
+
+    @AfterClass
+    public static void afterAll() {
+        system.terminate();
+    }
+
+    @After
+    public void clear() {
+        EndpointTestHelper.clearRegistries(system);
+        EnvTestHelper.clearRegistries(system);
+    }
+
+    @Test
+    public void registerResolver() {
+        AbstractEndpointResolver resolver = new DummyLocalhostJavaResolver();
+        EndpointResolverRegistry.get(system).register(resolver);
+        assertEquals(1, EndpointResolverRegistry.get(system).endpointResolvers().size());
+        assertEquals("DummyLocalhostJavaResolver",
+                EndpointResolverRegistry.get(system).endpointResolvers().head().name());
+    }
+
+    @Test
+    public void registerLambdaResolver() {
+        EndpointResolverRegistry.get(system).register("RogueResolver",
+                (svcName, env) -> Optional.of(Endpoint.create("http://myrogueservice.com")));
+        assertEquals(1, EndpointResolverRegistry.get(system).endpointResolvers().size());
+        assertEquals("RogueResolver", EndpointResolverRegistry.get(system).endpointResolvers().head().name());
+    }
+
+    @Test
+    public void registerResolveSSLResolver() {
+        Optional<SSLContext> sslOption = Optional.empty();
+        try {
+            sslOption = Optional.of(SSLContext.getDefault());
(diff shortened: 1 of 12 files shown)


squbs-ext/src/test/java/org/squbs/endpoint/DummyLocalhost2Resolver.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.squbs.endpoint;
+
+import org.squbs.env.DEV;
+import org.squbs.env.Default;
+import org.squbs.env.Environment;
+
+import java.util.Optional;
+
+public class DummyLocalhost2Resolver extends AbstractEndpointResolver {
```
anilgursel: Can we rename this as `DummyLocalhostJavaResolver` instead?  Or move it under `org.squbs.endpoint.japi` and just name `DummyLocalhostResolver`?

squbs-ext/src/test/java/org/squbs/endpoint/EndpointTest.java
```diff
@@ -0,0 +1,110 @@
...
+public class EndpointTest {
+
+    private static final ActorSystem system = ActorSystem.create("EndpointTest");
+
+    @AfterClass
+    public static void afterAll() {
+        system.terminate();
+    }
+
+    private void clear() {
```
anilgursel: Instead of calling this in each test mode, we should annotate it with `@After` to run after each test.

squbs-ext/src/test/java/org/squbs/endpoint/EndpointTest.java
```diff
@@ -0,0 +1,110 @@
...
+    public void registerLambdaResolver() {
+        EndpointResolverRegistry.get(system).register("RogueResolver",
+                (svcName, env) -> Optional.of(Endpoint.create("http://myrogueservice.com")));
+        assertEquals(1, EndpointResolverRegistry.get(system).endpointResolvers().size());
+        assertEquals("RogueResolver", EndpointResolverRegistry.get(system).endpointResolvers().head().name());
+        clear();
+    }
+
+    @Test
+    public void registerResolveSSLResolver() throws NoSuchAlgorithmException {
```
anilgursel: Since you already catch `NoSuchAlgorithmException`, do we still need the `throws NoSuchAlgorithmException`?

squbs-ext/src/test/java/org/squbs/endpoint/EndpointTest.java
```diff
@@ -0,0 +1,110 @@
...
+        EndpointResolverRegistry.get(system).register("RogueResolver",
+                (svcName, env) -> Optional.of(Endpoint.create("http://myrogueservice.com")));
+        Option<EndpointResolver> resolver =
+                EndpointResolverRegistry.get(system).route("abcService", Default.value());
+        assertEquals("RogueResolver", resolver.get().name());
+        assertEquals("http://myrogueservice.com",
+                resolver.get().resolve("abcService", Default.value()).get().uri().toString());
+        clear();
+    }
+}
```
anilgursel: We should probably also have a tests that checks the ordering when two resolvers are registered.  Also, one that checks the ordering when one is registered with lambda API and the other with implementing the abstract class.

squbs-ext/src/test/java/org/squbs/env/EnvironmentTest.java
```diff
@@ -0,0 +1,58 @@
...
+    public static void afterAll() {
+        system.terminate();
+    }
+
+    @Test
+    public void resolveDefault() {
+        EnvironmentResolverRegistry.get(system).register(DummyProdEnvironmentResolver.get());
+        Environment env = EnvironmentResolverRegistry.get(system).resolve();
+        assertEquals(PROD.value(), env);
+        EnvTestHelper.clearRegistries(system);
```
anilgursel: Instead of calling this in each test, we should probably do it with `@After`?

squbs-ext/src/test/scala/org/squbs/endpoint/EndpointSpec.scala
```diff
@@ -63,13 +82,26 @@ class EndpointSpec extends TestKit(ActorSystem("EndpointSpec")) with FlatSpecLik
     resolver.value.resolve("abcService") should be (Some(Endpoint("http://localhost:8080")))
   }
 
+  it should "resolve the endpoint with resolver lambda" in {
+    EndpointResolverRegistry(system).register("RogueResolver", (_, _) => Some(Endpoint("http://myrogueservice.com")))
+    val resolver = EndpointResolverRegistry(system).route("abcService")
```
anilgursel: Just thinking out loud.  Should we rename `route` as `findResolver`?  Somehow `route` does not sound right to me. 

Moreover, AFAIR, it is used only in tests.  Should we consider making the method package private?  I feel like we adding it to the public API without an actual use case.

squbs-ext/src/test/scala/org/squbs/env/EnvironmentSpec.scala
```diff
@@ -57,4 +57,16 @@ with Matchers with BeforeAndAfterEach{
     EnvironmentResolverRegistry(system).resolve should be (PROD)
   }
 
+  it should "just warn and not throw exceptions when unregister non-existing resolver" in {
```
anilgursel: `it should "just warn and not throw exceptions when non-existing resolver is unregistered"` ?

Actually, since we do not assert "just warn" part, I would reword it as

`it should "just warn and not throw exceptions when unregister non-existing resolver"`

squbs-ext/src/test/scala/org/squbs/endpoint/EndpointSpec.scala
akara: Will do. It did not sound right to me, either.

squbs-ext/src/test/java/org/squbs/endpoint/EndpointTest.java
akara: Will do. I saw you have those in the Scala tests already so I was just intending to test the Java API. But sure, will do that.

squbs-ext/src/test/scala/org/squbs/env/EnvironmentSpec.scala
akara: Did you mean `it should "not throw exceptions when unregister non-existing resolver"`?

Otherwise the suggested text is the same as what's in there.

PR #365, opened 2016-12-29
#349: Refined JSON marshaller APIs and moved to squbs-ext

akara:

- [X] Commits are squashed.
- [X] Tests added.


akara:
Based on the previous comment on the ReflectHelper, I've checked and every serialization/deserialization indeed goes through the decision point whether it is a Java or Scala object. Also if needed, it goes through the `TypeTag`<->`Manifest` and `TypeTag`<->`ClassTag` conversion. I would not treat this as a red flag just yet. Even though the API name has "reflect" in it, we're actually reading the `TypeTag`/`Manifest` information generated at compile time and do not really use any Java reflection ourselves. Lets keep it this way until we have a profile pointing us otherwise.

Alternatively, we can also use `Manifest` instead of `TypeTag`. It will avoid one conversion for Json4s (which is using `Manifest`). There will still be `Manifest`<->`ClassTag` conversion for Jackson. That is not bad, though. `ClassTag` information is a small subset of `Manifest` or `TypeTag` information. But I'd do it as a separate issue and with more evidence this is becoming an issue. We can file an issue to investigate the overhead of this check (`isJavaClass`) and conversion, though. I won't do premature optimization of this as the cost is not yet clear to me. Generally, using `TypeTag` is more forward looking. Using `Manifest` incurs less conversions. So we have to choose between the two.

Issue #327, opened 2016-11-14
Circuit Breaker for Akka Streams


anilgursel:

anilgursel:
Please see docs at https://github.com/paypal/squbs/blob/master/docs/circuitbreaker.md.  This should address https://github.com/akka/akka/issues/15307 and https://github.com/akka/akka-http/issues/446.

PR #367, opened 2017-01-03
#327 Timeout BidiFlow

anilgursel:

- [x] Commits are squashed.
- [x] Tests added.
 
Changing Timeout flow from a flow that broadcasts (with a `delay` stage) to a single `Bidi` `GraphStage`.  This gives better control to us.


squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -0,0 +1,338 @@
...
+    new TimeoutBidiUnordered(timeout, LongIdGenerator().nextId())
+
+  def apply[In, Out, Id](timeout: FiniteDuration, idGenerator: () => Id): TimeoutBidiUnordered[In, Out, Id] =
+    new TimeoutBidiUnordered(timeout, idGenerator)
+}
+
+final class TimeoutBidiUnordered[In, Out, Id](timeout: FiniteDuration, idGenerator: () => Id) extends
+  TimeoutBidi[In, (Id, In), (Id, Out), Out](timeout) {
+
+  val timeouts = new mutable.LinkedHashMap[Id, (In, Long)]
```
anilgursel: Should we specify the size of the map to attributes.inputBuffer?  Not confident about that though.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -0,0 +1,338 @@
...
+  val readyToPush = mutable.Queue[(Out, Long)]()
+
+  override def mapFromInToWrapped(element: In): (Id, In) = {
+    val id = idGenerator()
+    timeouts.put(id, (element, System.nanoTime()))
```
anilgursel: Calling `System.nanoTime()` quite frequently. Is it costly to call System.nanoTime() too often? AFAIR, it does some caching internally.

akara: It should be OK to call `System.nanoTime` often, at least on Linux (and do we care about the rest right now?). This is a rather direct hardware call.

docs/flow-timeout.md
```diff
@@ -14,61 +14,109 @@ Add the following dependency to your `build.sbt` or scala build file:
...
-import org.squbs.streams.StreamTimeout._
 
-val flow = Flow[String].map(s => findTheNumberOfEnglishWordsThatStartWith(s))
-Source("a" :: "b" :: "c" :: Nil).
-  via(flow.withTimeout(20 milliseconds)).
-  runWith(Sink.seq)
-```
+If there is no downstream demand, then no timeout message will be pushed down.  Moreover, no timeout check will be done unless there is downstream demand.  Accordingly, based on the demand scenarios, some messages might be send as a `Success` even after the timeout duration.
+
+Timeout precision is 10ms to avoid unnecessary timer scheduling cycles.
```
akara: Just to protect yourself, you may want to say it is at best 10ms. This can be more in case of a lack of downstream demand. They can also be more in case of GC, context switches, etc.

docs/flow-timeout.md
```diff
@@ -14,61 +14,109 @@ Add the following dependency to your `build.sbt` or scala build file:
...
+    Flow.<String>create().map(s -> findAnEnglishWordThatStartWith(s));
+        
+Source.from(Arrays.asList("a", "b", "c"))
+    .via(timeoutBidiFlow.join(flow))
+    .runWith(Sink.seq(), mat);
+```   
+
+#### Flows without message order guarantee
+
+`TimeoutBidiFlowUnordered` is used to create a timeout `BidiFlow` to wrap flows that do not guarantee the order of messages.  To uniquely identify each element and its corresponding timing marks, an `id` of type `Long` is generated internally.  This `id` needs to be carried around by the wrapped flow.  Accordingly, the wrapped flow needs to be pulling/pushing a `Tuple2[Long, T]`.  The `id` type and the generator can be customized, please see  [Customizing Id type and Id generators](#customizing-id-type-and-id-generators) section.
```
akara: I know this is painful! But the convention for Java is using `Pair`. Converting back and forth is not that easy.
If needed, you may want to move the conversion package from squbs-httpclient to squbs-ext to help you with such conversions. Since squbs-httpclient depends on squbs-ext, it will have access to such facilities. I definitely have no problem with that.

docs/flow-timeout.md
```diff
@@ -14,61 +14,109 @@ Add the following dependency to your `build.sbt` or scala build file:
...
+Source("a" :: "b" :: "c" :: Nil)
+  .via(timeoutBidiFlow.join(flow))
+  .runWith(Sink.seq)
+```
+
+###### Java
+
+```java
+final BidiFlow<String, Tuple2<UUID, String>, Tuple2<UUID, String>, Try<String>, NotUsed> timeoutBidiFlow =
+    TimeoutBidiFlowUnordered.create(timeout, () -> UUID.randomUUID());
```
akara: It seems like we really don't care about the ordering of the id. It should work as long as it is unique. Looking at how Akka HTTP deals with this, it asks the user to provide the context, as in `request -> 42`. I'm not sure that's better than allowing users to provide an id generator. On one side, this API is easier to understand.

We need to think about the way to use it if the message already supplies the context. Either the message is already a `Tuple2` in the format of `payload -> context` as done in Akka HTTP. Or the message contains the context in some other form. If it is already a tuple, we need to be able to consume it. If the context needs to be extracted, the closure should have the signature `msg => id` or `msg => context`. This will work even if the context is hidden in the message (for instance, by mixing in a trait `MessageContext`). The `msg => id` knows about the message structure and can pull out the required information. Even if it is a `Tuple2`, the id can just be pulled out by providing the closure`{ case (_, ctx) => ctx }` or something of this nature.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -0,0 +1,338 @@
...
+  val in = Inlet[In]("TimeoutBidi.in")
+  val fromWrapped = Inlet[FromWrapped]("TimeoutBidi.fromWrapped")
+  val toWrapped = Outlet[ToWrapped]("TimeoutBidi.toWrapped")
+  val out = Outlet[Try[Out]]("TimeoutBidi.out")
+  val shape = BidiShape(in, toWrapped, fromWrapped, out)
+  val expireOffset = timeout.toNanos
+  private[this] def timerName = "TimeoutBidi"
+  val precision = 10.milliseconds.toNanos
+  val delayMillis = timeout.toMillis
+  var downstreamDemand = 0
```
anilgursel: This is for the following scenario:

- Downstream demand is provided by `push`ing in `onTimer`.
- Downstream demands again with `onPull`, but since wrapped flow has not provided anything yet, we cannot pull it. 

Every downstream demand needs to be passed to upstream as a demand.  When we address the downstream demand with a timeout message, we should not swallow the demand and need to make sure it needs to be passed upstream.

Good point, I should make it `private`.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -0,0 +1,338 @@
...
+import scala.concurrent.duration._
+import scala.util.{Failure, Success, Try}
+
+/**
+  *    +--------------------+
+  * ~> | in       toWrapped | ~>
+  *    |                    |
+  * <~ | out    fromWrapped | <~
+  *    +--------------------+
+  *
```
akara: Just a thought, still based on my comment on the doc. If the message already has a context, all we need is the function to extract the context, right? We don't need to explicitly wrap it.

To support the Akka HTTP convention of doing `payload -> context`, we can provide a default extractor `{ case (_, b) => b }`. But then, if the message does not have a context, we need to wrap it. We need to make a call here. Either we wrap it anyway and force the `Flow` to the right of the `TimeoutBidi` deal with this concrete format of wrapper tuple, or we can just say "Have it your way, just provide me an extraction function to get to the key."

docs/flow-timeout.md
anilgursel: 👍 

I actually thought about using `akka.scheduler.tick-duration` (please see http://doc.akka.io/docs/akka/current/general/configuration.html), but there is not access to the system config and I do not want to pass actor system or configuration around.

docs/flow-timeout.md
anilgursel: It was in my mind over the weekend actually, but since there were too many small items I was pushing it towards the end, and apparently forgot to do. 😄

I will change it to `Pair`.

docs/flow-timeout.md
anilgursel: We care about ordering only for the flows that keep the order, and in that scenario, we do not generate an id anyway.

I considered using Akka HTTP style.  However, in that scenario, Akka HTTP is the one that owns the flow that carries the context.  In Timeout use case, it is the app that owns the flow that carries the context.  So, the usage and documentation might get more complicated.  If the app flow already has a flow that carries a context, it can easily be adapted with a `map` stage between.

That's the reason I preferred making it a `Tuple`.  But, I am up for further discussion.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
anilgursel: It is not just extraction function, it is also about the setting the timeout context to the flow context.  So, potentially two functions needed.

docs/flow-timeout.md
akara: I agree with the complexity (also for documentation) point of view. Perhaps we need to build a few e2e apps before deciding this further. Just one rather simple ask. The key generation function should be `M=>K` instead if `()=>K`. You always pass the message in for this function. It is up to the provider to use the message or not. But it allows users to generate key from the message (such as extracting an existing context from the message).

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
akara: Yes, I see. I won't do the second one. Leave the wrapper in place. Just ensure the signature supports key extraction from the message. So `M=>K`.

docs/flow-timeout.md
anilgursel: Good point! I will change the signature to `M => K`.

squbs-ext/src/test/java/org/squbs/streams/TimeoutBidiFlowTest.java
```diff
@@ -0,0 +1,111 @@
...
+    @Test
+    public void testWithCustomIdGenerator() throws ExecutionException, InterruptedException {
+        final ActorRef delayActor = system.actorOf(Props.create(DelayActor.class));
+        final Flow<Tuple2<UUID, String>, Tuple2<UUID, String>, NotUsed> flow =
+                Flow.<Tuple2<UUID, String>>create()
+                        .mapAsyncUnordered(3, elem -> ask(delayActor, elem, 5000))
+                        .map(elem -> (Tuple2<UUID, String>)elem);
+
+        final BidiFlow<String, Tuple2<UUID, String>, Tuple2<UUID, String>, Try<String>, NotUsed> timeoutBidiFlow =
+                TimeoutBidiFlowUnordered.create(timeout, () -> UUID.randomUUID());
```
akara:
```java
TimeoutBidiFlowUnordered.create(timeout, UUID::randomUUID);
```

squbs-ext/src/test/scala/org/squbs/streams/TimeoutBidiFlowSpec.scala
```diff
@@ -0,0 +1,170 @@
...
+import org.scalatest.{AsyncFlatSpecLike, Matchers}
+
+import scala.concurrent.duration._
+import scala.util.{Failure, Success}
+
+class TimeoutBidiFlowSpec extends TestKit(ActorSystem("TimeoutBidiFlowSpec")) with AsyncFlatSpecLike with Matchers{
+
+  implicit val materializer = ActorMaterializer()
+
+  val timeout = 60 milliseconds
```
akara: No problem using postfixOps. But please do import the feature if you use it. This avoids compiler warnings.

```scala
import scala.language.postfixOps
```

docs/flow-timeout.md
```diff
@@ -14,61 +14,109 @@ Add the following dependency to your `build.sbt` or scala build file:
...
+        .mapAsyncUnordered(3, elem -> ask(ref, elem, 5000))
+        .map(elem -> (Tuple2<UUID, String>)elem);
 
-Source("a" :: "b" :: "c" :: Nil).
-  via(flow.withTimeout(20 milliseconds, (new BigIntIdGenerator).nextId())).
-  runWith(Sink.seq)
+Source.from(Arrays.asList("a", "b", "c"))
+    .via(timeoutBidiFlow.join(flow))
+    .runWith(Sink.seq(), mat);
 ```
```
anilgursel: Good point! 👍

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
anilgursel: 👍

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -0,0 +1,338 @@
...
+
+  def mapFromInToWrapped(elem: In): ToWrapped
+
+  def onPushFromWrapped(elem: FromWrapped, isOutAvailable: Boolean): Option[Try[Out]]
+
+  def onScheduledTimeout(): Option[Try[Out]]
+
+  def onPullOut(): Option[Try[Out]]
+
+  def isBuffersEmpty(): Boolean
```
anilgursel: 👍

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -0,0 +1,338 @@
...
+
+  def timeLeftForNextElemToTimeout(): Long = {
```
anilgursel: 👍

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -0,0 +1,338 @@
...
+    setHandler(toWrapped, new OutHandler {
+      override def onPull(): Unit = {
+        pull(in)
+      }
+      override def onDownstreamFinish(): Unit = completeStage()
+    })
+
+    setHandler(fromWrapped, new InHandler {
+      override def onPush(): Unit = {
+        onPushFromWrapped(grab(fromWrapped), isAvailable(out)) map { elem =>
```
anilgursel: 👍

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
anilgursel: Yes, caping the size and allocating from the beginning to prevent re-sizing operations.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -0,0 +1,338 @@
...
+
+  override def mapFromInToWrapped(element: In): (Id, In) = {
+    val id = idGenerator()
+    timeouts.put(id, (element, System.nanoTime()))
+    (id, element)
+  }
+
+  override def onPushFromWrapped(fromWrapped: (Id, Out), isOutAvailable: Boolean): Option[Try[Out]] = {
+    val (id, element) = fromWrapped
+    timeouts.remove(id) map { case(_, startTime) =>
```
anilgursel: 👍

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -0,0 +1,338 @@
...
+      readyToPush.enqueue((element, startTime))
+    }
+
+    if(isOutAvailable) pickNextElemToPush()
+    else None
+  }
+
+  override def firstElemStartTime = timeouts.headOption map { case (_, (_, startTime)) => startTime } getOrElse(0)
```
anilgursel: 👍

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -0,0 +1,338 @@
...
+
+  private def pickNextElemToPush(): Option[Try[Out]] = {
+    timeouts.headOption.filter { case(_, (_, firstElemStartTime)) =>
+      firstElemStartTime < expirationTime &&
+      readyToPush.headOption.filter { case(_, readyToPushStartTime) =>
+        readyToPushStartTime <= firstElemStartTime
+      }.isEmpty
+    } map { case(id, (_, _)) =>
+      timeouts.remove(id)
+      Failure(FlowTimeoutException())
```
anilgursel: Actually, this was something I had in mind to talk to you.  The question is how to do it.  Should it be part of the exception?  For instance `FlowTimeoutException` can be made typed? 

```scala
case class FlowTimeoutException[T](val t: T, msg: String = "Flow timed out!") extends TimeoutException(msg)
```

squbs-ext/src/test/java/org/squbs/streams/TimeoutBidiFlowTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+public class TimeoutBidiFlowTest {
+
+    final ActorSystem system = ActorSystem.create("TimeoutBidiFlowTest");
+    final Materializer mat = ActorMaterializer.create(system);
+    final FiniteDuration timeout = FiniteDuration.apply(60, TimeUnit.MILLISECONDS);
+    final Try<String> timeoutFailure = Failure.apply(new FlowTimeoutException("Flow timed out!"));
+
+    @Test
+    public void testFlowsWithMessageOrderGuarantee() throws ExecutionException, InterruptedException {
```
anilgursel: Old habits die hard 😄   I thought it is the convention.

squbs-ext/src/test/java/org/squbs/streams/TimeoutBidiFlowTest.java
anilgursel: 👍

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
akara: Strange, I don't see any operation on `LinkedHashMap` allowing you to do so. So you'll have to build the capping mechanism with back-pressure by yourself.

I'd refrain from that unless we see the need. Here we have a time cap already. A size cap while making the memory usage more predictable would also make it harder to use.

docs/flow-timeout.md
anilgursel: It is really painful 😕  The existing converters are Http specific, not sure if it is clean enough to create utilities under akka.http in squbs-ext..  😟

docs/flow-timeout.md
```diff
@@ -14,61 +14,139 @@ Add the following dependency to your `build.sbt` or scala build file:
...
+##### Scala
 
-val flow = Flow[TimerEnvelope[Long, String]].mapAsyncUnordered(2) { elem =>
-  (ref ? elem).mapTo[TimerEnvelope[Long, String]]
+`TimeoutBidiFlowUnordered[In, Out, Context](timeout: FiniteDuration)` is used to create an unordered timeout `BidiFlow`.  This `BidiFlow` can be joined with any flow that takes in a `(In, Context)` and outputs a `(Out, Context)`.
+
+```scala   
+val timeoutBidiFlow = TimeoutBidiFlowUnordered[String, String, UUID](timeout) 
+val flow = Flow[(String, UUID)].mapAsyncUnordered(10) { elem =>
+  (delayActor ? elem).mapTo[(String, UUID)]
```
anilgursel: rename `delayActor` to `ref`.

anilgursel:
The APIs that have *flow* in the name, e.g., `TimeoutBidiFlowOrdered` have the Scaladoc.  I believe your suggestion is to make sure we do the same for `TimeoutBidiOrdered` and `TimeoutBidiUnordered`.  Correct me if I am wrong.

Regarding `TParam`s, that makes sense.  I will tackle those with the circuit breaker PR (created https://github.com/paypal/squbs/issues/374).  I just do not want this to be delayed anymore to make sure I can file a PR over the long weekend for CB.

akara:
All good! Thanks for your patience!

Issue #369, opened 2017-01-10
CustomTestKit should detect resources like SimpleTestKit


akara:
Detecting resources from `src/main/resources` as well as its current `src/test/resources` would allow for full replacement of SimpleTestKit which auto-detects `squbs-meta.conf` from `src/main/resources`.

We should not make `withClassPath` true by default as not to have multiple test cases conflicting by over-scanning by default.

anilgursel:
If we automatically scan `src/main/resources`; we also need to make sure that we can override anything configuration in it.

akara:
Can you please elaborate more?

anilgursel:
For instance, if there is some resource defined in `src/main/rsource/squbs-meta.conf`; and in a test, we do not want that to be enabled or set the path to some other route.

anilgursel:
We should also assess the its cost to test execution time.  I would not expect much; but, something to keep in mind.

akara:
Fixed, with issue #380 created to deal with config.

PR #370, opened 2017-01-10
#369: Made CustomTestKit scan resources in src/main/resources, too.

akara:
We now need to avoid resource duplication. So built dedup logic into UnicomplexBoot resource scanning.

- [X] Commits are squashed.
- [X] Tests added.

### DIFF
--- a/squbs-testkit/src/main/scala/org/squbs/testkit/CustomTestKit.scala
+++ b/squbs-testkit/src/main/scala/org/squbs/testkit/CustomTestKit.scala
@@ -24,16 +24,19 @@ import akka.testkit.{ImplicitSender, TestKit}
 import com.typesafe.config.{Config, ConfigFactory}
 import org.scalatest.{BeforeAndAfterAll, Suite}
 import org.squbs.lifecycle.GracefulStop
-import org.squbs.unicomplex.{UnicomplexBoot, JMX, Unicomplex}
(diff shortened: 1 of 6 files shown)


akara:
Note: There is code provisioned for cases where the `ClassLoader` may return resource path strings other than expected or testable. So these edge cases (2 lines) are not covered. They cannot be covered unless we do something weird with the `ClassLoader` or come up with our own weird `ClassLoader` just for this test.

Harikiranvuyyuru:
LGTM

squbs-testkit/src/main/scala/org/squbs/testkit/CustomTestKit.scala
```diff
@@ -47,18 +50,41 @@ object CustomTestKit {
...
       .replace('.', '-')
       .replace('_', '-')
       .filter(_ != '$')
 
-  val defaultResource = Seq(getClass.getClassLoader.getResource("").getPath + "/META-INF/squbs-meta.conf")
+  /**
+    * Detects default resources for this test. These are usually at two locations:
+    * <ul>
+    *   <li>$project-path/target/scala-2.11/classes/META-INF/squbs-meta.conf</li>
+    *   <li>$project-path/target/scala-2.11/test-classes/META-INF/squbs-meta.conf</li>
```
anilgursel: Should we update the comment not to have `2.11`?  This will be invalid in the next release.

Issue #374, opened 2017-01-13
Scaladocs for `TimeoutBidiOrdered` and `TimeoutBidiUnordered` & `TParam`


anilgursel:
Scaladocs are there for `TimeoutBidiFlowOrdered` and `TimeoutBidiFlowUnordered`; however, we should also add for `TimeoutBidiOrdered` and `TimeoutBidiUnordered` as these can directly be used as well.  

We should also add `@tparam` for all APIs.

Issue #372, opened 2017-01-12
Akka Agent based CircuitBreakerState Implementation


anilgursel:

anilgursel:
The current implementation of Stream Circuit Breaker is based on `Atomic`, we should also try using an `Agent` based implementation.

anilgursel:
Akka 2.5 deprecated agents.  So, this issue is not valid anymore.

anilgursel:
Akka Agent is deprecated in 2.5.

Issue #361, opened 2016-12-17
Timed out element blocks subsequent element in Flow.withTimeout


akara:
```scala
  it should "work for flows that keep the order of messages" in {
    val delayActor = system.actorOf(Props[DelayActor])
    import akka.pattern.ask
    implicit val askTimeout = Timeout(5.seconds)
    val flow = Flow[String].mapAsyncUnordered(10)(elem => (delayActor ? elem).mapTo[String])

    val result = Source("a" :: "b" :: "c" :: Nil).via(flow.withTimeout(timeout)).runWith(Sink.seq)

    val expected = Success("a") :: Success("c") :: Failure(FlowTimeoutException()) :: Nil
    result map { _ should contain theSameElementsAs expected }
  }
```

Test above should be successful. At this time, it is not.

anilgursel:
While this is not a valid use case (`TimerEnvelope` API should be used if the flow may not be able to keep the ordering of message), we should at least need to figure out what's going on here.

anilgursel:
Here is a gist that show cases this problem: https://gist.github.com/anilgursel/aa70497a9431fc8b498c599d9d469e2b.  The first test passes while the second one fails.  If fusing is disabled (`implicit val materializer = ActorMaterializer(ActorMaterializerSettings(system).withAutoFusing(false))`), then both tests pass.

PR #375, opened 2017-01-17
#327 Akka Streams Circuit Breaker

anilgursel:

- [x] Commits are squashed.
- [x] Tests added.

Introduces Akka Streams BidiFlow to add Circuit Breaker functionality.  Also addresses the following issues:

- https://github.com/paypal/squbs/issues/374
- https://github.com/paypal/squbs/issues/359
- https://github.com/paypal/squbs/issues/361

Follow up issues:

- https://github.com/paypal/squbs/issues/373
- https://github.com/paypal/squbs/issues/372

-


squbs-ext/src/main/scala/org/squbs/circuitbreaker/impl/AtomicCircuitBreakerState.scala
```diff
@@ -0,0 +1,432 @@
...
+    * Holds reference to current state of CircuitBreaker - *access only via helper methods*
+    */
+  @volatile
+  private[this] var _currentStateDoNotCallMeDirectly: AtomicState = AtomicClosed
+
+  /**
+    * Holds reference to current resetTimeout of CircuitBreaker - *access only via helper methods*
+    */
+  @volatile
+  private[this] var _currentResetTimeoutDoNotCallMeDirectly: FiniteDuration = resetTimeout
```
akara: What did you want to signify with the underbar convention?

squbs-ext/src/test/java/org/squbs/circuitbreaker/CircuitBreakerBidiFlowTest.java
```diff
@@ -0,0 +1,291 @@
...
+            expectMsgEquals(Open.instance());
+            expectMsgEquals(HalfOpen.instance());
+            ref.tell("a", ActorRef.noSender());
+            expectMsgEquals(Closed.instance());
+        }};
+    }
+
+    @Test
+    public void testIncrementFailureCountBasedOnTheProvidedFunction() {
+        final ActorRef delayActor = system.actorOf(Props.create(org.squbs.streams.DelayActor.class));
```
akara: Might want to address this one from Codacy.

squbs-ext/src/main/scala/org/squbs/circuitbreaker/CircuitBreakerBidi.scala
```diff
@@ -0,0 +1,307 @@
...
+package org.squbs.circuitbreaker
+
+import java.util.Optional
+
+import akka.NotUsed
+import akka.http.org.squbs.util.JavaConverters
+import akka.http.org.squbs.util.JavaConverters.toJava
+import akka.japi.Pair
+import akka.stream._
+import akka.stream.scaladsl.BidiFlow
```
akara: Can be confusing, but use the IDE's optimize import option to clean this up.

docs/circuitbreaker.md
```diff
@@ -0,0 +1,291 @@
...
+```java
+CircuitBreakerBidiFlow.create(
+        circuitBreakerState,
+        Optional.empty(),
+        Optional.of(pair -> pair.first().get().equals("b")));
+```
+
+#### Creating CircuitBreakerState from configuration
+
+While `AtomicCircuitBreakerState` has programmatic API for configuration, it also allows the configuration to be provided in configuration:
```
akara: Sounds odd: "it also allows the configuration to be provided in configuration." Should it be "it also allows the configuration to be provided through the system configuration."?

docs/circuitbreaker.md
```diff
@@ -0,0 +1,291 @@
...
+
+It also keeps a `Gauge` for the circuit breaker state.
+
+To differentiate metrics across instance, `CircuitBreakerState` implementation require a name to passed in.
+
+`CircuitBreakerState` also allows a `MetricRegistry` instance to be passed in.  If no `MetricRegistry` is passed in, it creates one internally. 
+
+#### Potential effects of Circuit Open state on throughput
+
+If the upstream does not control the throughput, then the throughput of the stream might temporarily increase once the circuit is `Open`:  The downstream demand will be addressed with short circuit/fallback messages, which might (or might not) take less time than it takes the joined `Flow` to process an element.  To eliminate this problem, a throttle can be applied specifically for circuit breaker related messages.
```
akara: ... a throttle can be applied specifically for the circuit breaker open message: `(context, Failure(CircuitBreakerOpenException()))`.

squbs-ext/src/main/scala/org/squbs/circuitbreaker/impl/AtomicCircuitBreakerState.scala
anilgursel: This is just a copy from the Akka Circuit Breaker.  Not sure what's signified.  I can drop the underbar?

squbs-ext/src/test/java/org/squbs/circuitbreaker/CircuitBreakerBidiFlowTest.java
anilgursel: 👍

squbs-ext/src/main/scala/org/squbs/circuitbreaker/CircuitBreakerBidi.scala
anilgursel: 👍

docs/circuitbreaker.md
```diff
@@ -0,0 +1,291 @@
...
+```
+
+### Usage
+
+The circuit breaker functionality is provided as a `BidiFlow` that can be connected to a flow via the `join` operator.  `CircuitBreakerBidi` might potentially change the order of messages, so it requires a `Context` to be carried around.  The requirement is that either the `Context` itself or an attribute accessed via the `Context` should be able to uniquely identify an element.  Along with the `Context`, a `Try` is pushed downstream: 
+
+Circuit is `Closed`:
+
+   * If an output `msg` is provided by the wrapped flow within the timeout, then `Success(msg)` is passed down.
+   * Otherwise, a `Failure(FlowTimeoutException())` is pushed to downstream.
```
anilgursel: Updating it as `(Success(msg), context)` and `(Failure(FlowTimeoutException()), context)`.

docs/circuitbreaker.md
```diff
@@ -0,0 +1,291 @@
...
+
+
+Circuit is `Open`:
+
+   * The result of `fallback` function is pushed to downstream if one provided
+   * Otherwise, a `Failure(CircuitBreakerOpenException())` is pushed to downstream
```
anilgursel: 👍

docs/circuitbreaker.md
anilgursel: 👍

Also, chaning the title to "Creating CircuitBreakerState from system configuration".

docs/circuitbreaker.md
anilgursel: Modified it as "a throttle can be applied specifically for circuit breaker `Open` messages.".  Not using the code, because it might have to be done for fallback responses as well, and with that the example gets complicated.

squbs-ext/src/test/scala/org/squbs/circuitbreaker/CircuitBreakerBidiFlowSpec.scala
```diff
@@ -0,0 +1,375 @@
...
+class DelayActor extends Actor {
+
+  val delay = Map("a" -> 10.milliseconds, "b" -> 500.milliseconds, "c" -> 10.milliseconds)
+
+  def receive = {
+    case element: String =>
+      import context.dispatcher
+      context.system.scheduler.scheduleOnce(delay(element), sender(), element)
+    case element: (String, Long) =>
+      import context.dispatcher
```
akara: Just import `context.dispatcher` once inside this class. Makes code easier to read.

squbs-ext/src/test/scala/org/squbs/circuitbreaker/CircuitBreakerBidiFlowSpec.scala
```diff
@@ -0,0 +1,375 @@
...
+
+class DelayActor extends Actor {
+
+  val delay = Map("a" -> 10.milliseconds, "b" -> 500.milliseconds, "c" -> 10.milliseconds)
+
+  def receive = {
+    case element: String =>
+      import context.dispatcher
+      context.system.scheduler.scheduleOnce(delay(element), sender(), element)
+    case element: (String, Long) =>
```
akara:
```scala
case (element: String, _) => context.system.scheduler.scheduleOnce(delay(element), sender(), element)
```

You can also dismiss the third case match altogether.

squbs-ext/src/test/scala/org/squbs/circuitbreaker/CircuitBreakerBidiFlowSpec.scala
```diff
@@ -0,0 +1,375 @@
...
+    var context = 0L
+    val result = Source("a" :: "b" :: "b" :: "a" :: Nil)
+      .map { s => context += 1; (s, context) }
+      .via(circuitBreakerBidiFlow.join(flow))
+      .runWith(Sink.seq)
+
+    val failFastFailure = Failure(CircuitBreakerOpenException())
+    val expected = (Success("a"), 1) :: (flowFailure, 2) :: (flowFailure, 3) :: (failFastFailure, 4) :: Nil
+    whenReady(result) { r =>
+      r should contain theSameElementsInOrderAs(expected)
```
akara:
```scala
r should contain theSameElementsInOrderAs expected
```

squbs-ext/src/test/scala/org/squbs/circuitbreaker/CircuitBreakerBidiFlowSpec.scala
```diff
@@ -0,0 +1,375 @@
...
+    ref ! "a"
+    ref ! "b"
+    ref ! "b"
+    expectMsg(Open)
+    expectMsg(HalfOpen)
+    ref ! "a"
+    expectMsg(Closed)
+  }
+
+  it should "increment failure count based on the provided function" in {
```
akara: Hmm, the description may not be accurate. Are we really checking the failure count in this test?

squbs-ext/src/main/java/org/squbs/circuitbreaker/impl/AbstractAtomicCircuitBreakerLogic.java
```diff
@@ -0,0 +1,36 @@
...
+    protected final static long stateOffset;
+    protected final static long resetTimeoutOffset;
+
+    static {
+        try {
+            stateOffset = Unsafe.instance.objectFieldOffset(AtomicCircuitBreakerState.class.getDeclaredField("_currentStateDoNotCallMeDirectly"));
+            resetTimeoutOffset = Unsafe.instance.objectFieldOffset(AtomicCircuitBreakerState.class.getDeclaredField("_currentResetTimeoutDoNotCallMeDirectly"));
+        } catch(Throwable t){
+            throw new ExceptionInInitializerError(t);
+        }
```
anilgursel: Actually, this code is mostly taken from Akka Pattern.  Please see https://github.com/akka/akka/blob/master/akka-actor/src/main/java/akka/pattern/AbstractCircuitBreaker.java.  We can try different alternatives, but in the first cut, I did not want to invest the time on that and just rely on existing logic.

squbs-ext/src/main/scala/org/squbs/circuitbreaker/CircuitBreakerState.scala
```diff
@@ -0,0 +1,233 @@
...
+    */
+  final def failure(): Unit = {
+    metricRegistry.meter(FailureCount).mark()
+    fails()
+  }
+
+  /**
+    * Check if circuit should be short circuited.
+    */
+  final def shortCircuited(): Boolean = {
```
anilgursel: it actually updates metrics.  Doesn't it count?

anilgursel: > This applies to multiple defs in this class.

What other `def`?

squbs-ext/src/main/scala/org/squbs/circuitbreaker/CircuitBreakerState.scala
```diff
@@ -0,0 +1,233 @@
...
+    *
+    * @param fromState State we're coming from (`Closed` or `HalfOpen`)
+    */
+  protected final def tripBreaker(fromState: State): Unit = transition(fromState, Open)
+
+  /**
+    * Resets breaker to `Closed` state.  This is valid from `HalfOpen` state only.
+    *
+    */
+  protected final def resetBreaker(): Unit = transition(HalfOpen, Closed)
```
anilgursel: If it is already in the state that it wants to transition to, then the impl returns `false`, and `transition` just skips publishing state changed event.

```scala
protected final def transition(fromState: State, toState: State): Unit =
    if(transitionImpl(fromState, toState)) eventBus.publish(CircuitBreakerEvent(toState, toState))
```

squbs-ext/src/main/scala/org/squbs/circuitbreaker/CircuitBreakerBidi.scala
```diff
@@ -0,0 +1,305 @@
...
+                                           fallback: Option[((In, Context)) => (Try[Out], Context)],
+                                           failureDecider: Option[((Try[Out], Context)) => Boolean])
+  extends GraphStage[BidiShape[(In, Context), (In, Context), (Try[Out], Context), (Try[Out], Context)]] {
+
+  private val in = Inlet[(In, Context)]("CircuitBreakerBidi.in")
+  private val fromWrapped = Inlet[(Try[Out], Context)]("CircuitBreakerBidi.fromWrapped")
+  private val toWrapped = Outlet[(In, Context)]("CircuitBreakerBidi.toWrapped")
+  private val out = Outlet[(Try[Out], Context)]("CircuitBreakerBidi.out")
+  val shape = BidiShape(in, toWrapped, fromWrapped, out)
+  private var upstreamFinished = false
```
anilgursel: Thanks for catching!  The following three should actually move inside `createLogic`:

```scala
private var upstreamFinished = false
    private val readyToPush = mutable.Queue[(Try[Out], Context)]()

    private def onPushFromWrapped(elem: (Try[Out], Context), isOutAvailable: Boolean): Option[(Try[Out], Context)] = {
      readyToPush.enqueue(elem)
      if(isOutAvailable) Some(readyToPush.dequeue())
      else None
    }
```

squbs-ext/src/main/scala/org/squbs/circuitbreaker/CircuitBreakerBidi.scala
```diff
@@ -0,0 +1,305 @@
...
+  private val readyToPush = mutable.Queue[(Try[Out], Context)]()
+  private val isFailure = failureDecider.getOrElse((e: (Try[Out], Context)) => e._1.isFailure)
+
+  private def onPushFromWrapped(elem: (Try[Out], Context), isOutAvailable: Boolean): Option[(Try[Out], Context)] = {
+    readyToPush.enqueue(elem)
```
anilgursel: 👍

squbs-ext/src/main/scala/org/squbs/circuitbreaker/CircuitBreakerBidi.scala
```diff
@@ -0,0 +1,305 @@
...
+    })
+
+    setHandler(toWrapped, new OutHandler {
+      override def onPull(): Unit = if(!hasBeenPulled(in)) pull(in)
+      override def onDownstreamFinish(): Unit = completeStage()
+    })
+
+    setHandler(fromWrapped, new InHandler {
+      override def onPush(): Unit = {
+        val elemWithcontext = grab(fromWrapped)
```
anilgursel: IntelliJ also warns!  Somehow overlooked it.

squbs-ext/src/main/scala/org/squbs/circuitbreaker/CircuitBreakerBidi.scala
```diff
@@ -0,0 +1,305 @@
...
+        if(readyToPush.isEmpty) completeStage()
+        else upstreamFinished = true
+      }
+
+      override def onUpstreamFailure(ex: Throwable): Unit = fail(out, ex)
+    })
+
+    setHandler(out, new OutHandler {
+      override def onPull(): Unit =
+        if(!upstreamFinished || !readyToPush.isEmpty) readyToPush.dequeueFirst((_: (Try[Out], Context)) => true) match {
```
anilgursel: 👍

squbs-ext/src/main/scala/org/squbs/circuitbreaker/impl/AtomicCircuitBreakerState.scala
```diff
@@ -0,0 +1,432 @@
...
+  /**
+    * Helper method for access to underlying state via Unsafe
+    *
+    * @param oldState Previous state on transition
+    * @param newState Next state on transition
+    * @return Whether the previous state matched correctly
+    */
+  @inline
+  private[this] def swapAtomicState(oldState: AtomicState, newState: AtomicState): Boolean =
+  Unsafe.instance.compareAndSwapObject(this, AbstractAtomicCircuitBreakerLogic.stateOffset, oldState, newState)
```
anilgursel: I think in the second round, we should try to avoid `Unsafe`.  We can re-visit this at that point?

squbs-ext/src/main/scala/org/squbs/circuitbreaker/impl/AtomicCircuitBreakerState.scala
```diff
@@ -0,0 +1,432 @@
...
+
+  /**
+    * Mark a failed call through CircuitBreaker.
+    */
+  def fails(): Unit = currentAtomicState.fails()
+
+  /**
+    * Check if circuit should be short circuited.
+    */
+  def isShortCircuited(): Boolean = currentAtomicState.isShortCircuited
```
anilgursel: 👍

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -23,153 +23,151 @@ import akka.http.org.squbs.util.JavaConverters._
...
-        onPushFromWrapped(grab(fromWrapped), isAvailable(out)) map { elem =>
-          push(out, elem)
-        }
-        if(downstreamDemand > 0) {
-          pull(fromWrapped)
-          downstreamDemand -= 1
-        }
+  setHandler(fromWrapped, new InHandler {
+    override def onPush(): Unit = {
+      onPushFromWrapped(grab(fromWrapped), isAvailable(out)) map { elem =>
```
anilgursel: 👍

squbs-ext/src/test/scala/org/squbs/circuitbreaker/CircuitBreakerBidiFlowSpec.scala
anilgursel: how about `it should "decide on failures based on the provided function" in {`?

squbs-ext/src/test/scala/org/squbs/circuitbreaker/CircuitBreakerBidiFlowSpec.scala
anilgursel: 👍

squbs-ext/src/test/scala/org/squbs/circuitbreaker/CircuitBreakerBidiFlowSpec.scala
anilgursel: 👍

squbs-ext/src/test/scala/org/squbs/circuitbreaker/CircuitBreakerBidiFlowSpec.scala
anilgursel: 👍

squbs-ext/src/test/scala/org/squbs/circuitbreaker/CircuitBreakerBidiFlowSpec.scala
anilgursel: Actually, it needs to be `case element: (String, _) => context.system.scheduler.scheduleOnce(delay(element._1), sender(), element)`.

anilgursel:
Pushed a new commit that addresses the review comments.  I dropped some questions for some of the comments though, let's discuss those.  

Having some intermittent test failures, looking into those.

squbs-ext/src/main/scala/org/squbs/circuitbreaker/CircuitBreakerState.scala
akara: Hmm, the query into `shortCircuited` causes a `meter` to increase? This is on the state object, highly shared. Is it the right place to increment the `ShortCircuitCount`? I'm just thinking out loud whether the method name `shortCircuited` is right. It should imply you're asking for a decision and record that decision in the metrics instead of a casual name. Some name (for lack of a better name) like `checkAndRecordShortCircuit()` would be more meaningful in this case.

anilgursel: Metric wise, this is the right place if we would like to meter how many times it preferred the short route.

Naming wise, if we rename it as `checkAndRecordShortCircuit`, then we need to do it for  `success` and `failure`  methods as well, which kinda makes the API less clean.  And in case we decide to drop metrics or provide it in a different way in the future (just assuming), then the API would need to change again.

anilgursel: Copying my comment from Gitter https://gitter.im/paypal/squbs?at=589b36f96b2d8dd55219d3c5:

> How about the following method names?
>
> - markSuccess
> - markFailure
> - checkAndMarkIfShortCircuit

PR #378, opened 2017-01-20
#377: Generified resolver to resolve anything based on name and env

akara:

- [X] Commits are squashed.
- [X] Tests added.

We need to visit documentation before merging this PR.


squbs-ext/src/main/scala/org/squbs/resolver/Resolver.scala
```diff
@@ -0,0 +1,219 @@
...
+
+  private[resolver] final def toEndpointResolver = new Resolver[T] {
+    def name: String = AbstractResolver.this.name
+    def resolve(name: String, env: Environment = Default): Option[T] = AbstractResolver.this.resolve(name, env).asScala
+  }
+}
+
+class ResolverRegistryExtension(system: ExtendedActorSystem) extends Extension with LazyLogging {
+
+  private[resolver] var resolvers = List.empty[(Class[_], Resolver[_])]
```
Harikiranvuyyuru: Will the resolver behavior be predictable when we do register and unregister resolvers at the same time.If it is unpredicatable, should we add in documentation suggesting best practices on how the resolver should be registered and unregistered

akara: No. It won't be predictable for concurrent `register`/`unregister`. I've added full resolver docs.

Harikiranvuyyuru:
LGTM

docs/resolver.md
```diff
@@ -0,0 +1,180 @@
+#Resource Resolution
+
+Given that very few - if any - real life applications can work without external resources, environment aware resource resolution is becoming a crucial part of application infrastructure. squbs provides resource resolution through the `ResolverRegistry` and allows resources of any type to be resolved by name and environment which allows differentiation of the resource between production, qa, and dev environments.
```
anilgursel: should be "..  differentiation of a resource .." instead of "the".

docs/resolver.md
```diff
@@ -0,0 +1,180 @@
...
+
+The resolver sits in `squbs-ext`. Add the following dependency to your `build.sbt` or scala build file:
+
+```scala
+"org.squbs" %% "squbs-ext" % squbsVersion
+```
+
+###Usage
+
+The basic usage of the `Resolver` is for looking up resources. A type needs to be provided as the registry can hold resources of multiple types. We use the type `URI` in our samples in this documentation. A lookup call is shown in the followings:
```
anilgursel: "... can hold resources of multiple types.": It would be helpful to list a few examples here.

docs/resolver.md
```diff
@@ -0,0 +1,180 @@
...
+
+#### Discovery Chain
+
+The resource discovery follows a LIFO model. The most-recently registered resolver takes precedence over previously registered ones. The `ResolverRegistry` walks the chain one by one until there is a resolver compatible with the given type that provides the resource or the chain has been exhausted. In that case the registry will return a `None` for the Scala API and a `Optional.empty()` for the Java API.
+
+#### Type Compatibility
+
+The `ResolverRegistry` checks the requested type at the time of the `resolve` call. If the type of the registered resolver is the same type or a subtype of the requested type, that resolver will be used to try resolve the resource by name.
+
+Due to JVM type erasure, type parameters of the registered or requested types are not accounted for. For instance, a registration of type `java.util.ArrayList<String>` may be matched by a `resolve` call of type `java.util.List<Int>` as the type parameter `String` or `Int` is erased at runtime. Due to this limitation, using types with type parameters for registration and lookup is highly discouraged. The results are undefined - you may just get the wrong resource.
```
anilgursel: To keep the focus on generic parameter, should we use `java.util.List` in both instances?

If we would like to give example of type and subtype, it should be a separate example, IMO.

docs/resolver.md
anilgursel: Should we directly jump into environment resolution?  I believe we should first talk about resource resolution, then in addition to that we talk about environment resolution.  Those are two separate features that can be used together in harmony.  But, we need to make it clear that they are separate.

docs/resolver.md
```diff
@@ -0,0 +1,180 @@
...
+      case ("myservice2", QA) => Some(URI.create("http://myservice2.qa.mydomain.com"))
+      case ("myservice2", Default) => Some(URI.create("http://myservice2.mydomain.com"))
+      case _ => None
+    }
+  }
+}
+
+// Then just register the instance
+ResolverRegistry(system).register[URI](new MyResolver)
+```
```
anilgursel: It would be easier to read if we separate the two styles into two separate code blocks.

docs/resolver.md
```diff
@@ -0,0 +1,180 @@
...
+
+Due to JVM type erasure, type parameters of the registered or requested types are not accounted for. For instance, a registration of type `java.util.ArrayList<String>` may be matched by a `resolve` call of type `java.util.List<Int>` as the type parameter `String` or `Int` is erased at runtime. Due to this limitation, using types with type parameters for registration and lookup is highly discouraged. The results are undefined - you may just get the wrong resource.
+
+For simplicity, it is highly encouraged not to make use of type hierarchies. All registered types should be distinct types.
+
+#### Registering Resolvers
+
+There are two styles of APIs provided for resolver registration. One is a shortcut API allowing passing a closure/lambda as the resolver. The closure/lambda return type has to be `Option[T]` for Scala and `Optional<T>` for Java. The other full API takes a `Resolver[T]` in Scala or an `AbstractResolver<T>` in Java, `T` being the resource type. These can be seen in the followings:
```
anilgursel: Some character issue here:

screen shot 2017-02-06 at 3 29 44 pm

docs/resolver.md
```diff
@@ -0,0 +1,180 @@
...
+
+#### Type Compatibility
+
+The `ResolverRegistry` checks the requested type at the time of the `resolve` call. If the type of the registered resolver is the same type or a subtype of the requested type, that resolver will be used to try resolve the resource by name.
+
+Due to JVM type erasure, type parameters of the registered or requested types are not accounted for. For instance, a registration of type `java.util.ArrayList<String>` may be matched by a `resolve` call of type `java.util.List<Int>` as the type parameter `String` or `Int` is erased at runtime. Due to this limitation, using types with type parameters for registration and lookup is highly discouraged. The results are undefined - you may just get the wrong resource.
+
+For simplicity, it is highly encouraged not to make use of type hierarchies. All registered types should be distinct types.
+
+#### Registering Resolvers
```
anilgursel: This section should come before `Discovery Chain`.  Otherwise, it is difficult to grasp "Discovery Chain" and "Type Compatibility" sections.

docs/resolver.md
```diff
@@ -0,0 +1,180 @@
+#Resource Resolution
```
anilgursel: Overall, I feel like this doc emphasizes environment resolution (with wording and examples) more than resource resolution.  I believe environment resolution should be additional feature, probably in a separate section?

anilgursel: `Endpoint`  is also not talked much?  We should mention what an `Endpoint` is what examples are there, what kind of things it could resolve to, e.g., in case of http, a `URI` and `SSLContext`.

docs/httpclient.md
```diff
@@ -96,66 +96,67 @@ HttpRequest complexRequest =
...
+`squbs-httpclient` does not require hostname/port combination to be provided during client pool creation.  Instead it allows a service discovery chain to be registered which allows resolving `HttpEndpoint`s by a string identifier by running through the registered service discovery mechanisms.  For instance, in the above example, `"sample"` is the logical name of the service that client wants to connect, the configured service discovery chain will resolve it to an `HttpEndpoint` which includes host and port, e.g., `http://akka.io:80`.
 
-There are two variations of registering endpoint resolvers as shown below. The closure style allows more compact and readable code. However, the subclass has the power to keep state and make resolution decisions based on such state:
+There are two variations of registering resolvers as shown below. The closure style allows more compact and readable code. However, the subclass has the power to keep state and make resolution decisions based on such state:
 
 ##### Scala
 
-Register function type `(String, Env) => Option[Endpoint]`:
+Register function type `(String, Env) => Option[HttpEndpoint]`:
```
akara: There is already a link to `resolver.md`, just below. Please let me know if you think I should add more.

docs/resolver.md
akara: IMHO, I think we generally address resource resolution, not environment resolution. All we're saying is that these resource may not be the same in different environments. What part of this suggest we're discussing environment resolution?

I'm not sure how I should change it to ensure the right perception. Please let me know your ideas.

docs/resolver.md
akara: 3 code blocks. One for each style and one for the registration. We can discuss by I feel it is harder to do.

anilgursel: If it is together, it gives the impression that the entire code block is needed (unless each item is a one/two liner)..  

With long code examples, it looks related and the code sometimes look intimidating..

But, I leave it up to you..

Issue #388, opened 2017-02-04
refactor readConfigs(file) in UnicomplexBoot


thefourtheye:
1. The `readConfigs` function which reads from jar file or a directory, defines two functions, based on the type of file passed as input (jar or directory). That can be refactored into two private functions.

2. The current file being processed is modified in both the functions mentioned in the previous point. We could reduce the footprint of that `var` variable by modifying it only before calling the refactored functions.

3. The code actually is written in a little imperative style. We could refactor it with the functional goodies of Scala.

anilgursel:
Just marking it as "fixed".  We'll close it at the end of the sprint.

PR #389, opened 2017-02-04
#388 Refactor readConfigs(file)

thefourtheye:
Fixes: https://github.com/paypal/squbs/issues/388

1. The `readConfigs` function which reads from jar file or a directory, defines two functions, based on the type of file passed as input (jar or directory). That can be refactored into two private functions.

2. The current file being processed is modified in both the functions mentioned in the previous point. We could reduce the footprint of that `var` variable by modifying it only before calling the refactored functions.

3. The code actually is written in a little imperative style. We could refactor it with the functional goodies of Scala.

- [x] Commits are squashed.
- [ ] Tests added.

### DIFF
--- a/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
+++ b/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
@@ -30,8 +30,8 @@ import com.typesafe.config._
 import com.typesafe.scalalogging.LazyLogging
 import org.squbs.lifecycle.ExtensionLifecycle
 import org.squbs.pipeline.PipelineSetting
-import org.squbs.util.ConfigUtil._
 import org.squbs.unicomplex.UnicomplexBoot.CubeInit
+import org.squbs.util.ConfigUtil._
 
 import scala.annotation.tailrec
 import scala.collection.concurrent.TrieMap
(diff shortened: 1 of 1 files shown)


squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
```diff
@@ -133,54 +134,49 @@ object UnicomplexBoot extends LazyLogging {
...
 
-  private[this] def readConfigs(jarName: String): Option[Config] = {
+  private def createReaderFromFS(directory: File): String => Option[Reader] = {
+    (filePath: String) => Option(new File(directory, filePath)) filter (_.isFile) map {
+      file => new InputStreamReader(new FileInputStream(file), "UTF-8")
+    }
+  }
+
+  private def createReaderFromJarFile(file: File): String => Option[Reader] = {
+    lazy val jarFile = new JarFile(file)
```
akara: Need to be `lazy`?

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
```diff
@@ -133,54 +134,49 @@ object UnicomplexBoot extends LazyLogging {
...
+    (filePath: String) => Option(jarFile.getEntry(filePath)) filterNot (_.isDirectory) map {
```
akara: Ditto here. Use `collect` is cleaner and more efficient.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
```diff
@@ -133,54 +134,49 @@ object UnicomplexBoot extends LazyLogging {
     boot.copy(cubes = cubeList, jarConfigs = jarConfigs, listeners = activeListeners, listenerAliases = activeAliases)
   }
 
-  private[this] def readConfigs(jarName: String): Option[Config] = {
+  private def createReaderFromFS(directory: File): String => Option[Reader] = {
+    (filePath: String) => Option(new File(directory, filePath)) filter (_.isFile) map {
```
thefourtheye: I am updating with `collect`. Regarding the `(filePath: String)`, I generally include the types because when I read/review code, it would be easier if I didn't have to look around much to know the type of the variable. After all, code is written once and read many times :-) If you feel strongly about it, I can remove.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
thefourtheye: Yes, when `createReaderFromJarFile` is invoked by `getConfigReader`, `new JarFile` could throw an Exception. We are not catching it anywhere. When the returned function is invoked, the exceptions are caught there and processed properly.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
thefourtheye: Ack!

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
akara: I understand your point now. But depending on the lazy exception throwing of `lazy val` seems very obscure. Would such code be clearer (albeit a bit longer)?

```scala
  private def createReaderFromJarFile(file: File): String => Option[Reader] = {
    val jarFileTry = Try { new JarFile(file) }
    (filePath: String) => jarFileTry match {
      case Success(jarFile) =>
        Option(jarFile.getEntry(filePath)) collect {
          case entry if !entry.isDirectory => new InputStreamReader(jarFile.getInputStream(entry), "UTF-8")
        }
      case Failure(e) => throw e
    }
  }
```

thefourtheye: I love it :-) Updated now :-) PTAL.

akara:
LGTM. Will merge as soon as build is green (but may not be tonight - running out of juice). Thanks a ton!

Issue #386, opened 2017-02-03
ClientFlow should allow for environment-specific configuration


akara:
Such configuration should be recognized by `ClientFlow`:

```
svc1 {
  type = squbs.httpclient
  akka.http.host-connection-pool {
    max-connections = 10
  }

  QA {
    // Any standard configuration
    akka.http.host-connection-pool {
      max-connections = 2
    }
  }
}
```

anilgursel:
I tend to disagree with this issue.  We have `EnvironmentResolver` which could be used to customize configuration per environment.  Baking this functionality directly into `ClientFlow` does not sound right to me.

PR #390, opened 2017-02-06
#386: Added environment-specific configurations.

akara:

- [X] Commits are squashed.
- [X] Tests added.


anilgursel:
If you already have an environment resolution mechanism (by using `EnvironmentResolver`), then this feature just adds more confusion.  IMO, we should not muddy http client by hardcoding any environment specific logic in it.  A squbs user can define an environment specific configuration by just reading environment specific sections of `application.conf`.

akara:
I tripped on this when updating internal stuff (please see internal PR) and found it as a feature regression. That's why the issue and the relatively easy fix.

The `EnvironmentResolver` gives you the environment, but no configuration with it. So to achieve the same, the user will need to resolve the environment manually, then read the config for the desired client and environment from a user-specified location in the config, and finally pass the config to the `ClientFlow` upon creation. Don't you think that is harder to manage and support than having a well known location for environment-specific client configuration?

We better discuss face-to-face. Thanks!

anilgursel:
I agree, let's discuss.

docs/httpclient.md
```diff
@@ -192,6 +192,26 @@ sample {
...
+    akka.http.host-connection-pool {
+      max-connections = 4
+    }    
+  }
+}
+```
+
+In this case if the `ClientFlow` is created for the `QA` environment, the settings `max-connections = 4` would apply instead of `max-connections = 10` as specified for all other environments.
+
 ### Pipeline
```
DaJoker78: Ithoughtyouknew

DaJoker78: Tre4

DaJoker78: Shaggy34

DaJoker78: chadmichaellheureux

DaJoker78: World Inc

DaJoker78: Me

DaJoker78: HappyNes

Issue #376, opened 2017-01-17
DummyBidiFlow and DummyAbortableBidiFlow  example in the streamingpipeline documentation needs to be updated


Harikiranvuyyuru:
currently in the documentation we override create method with signature override def create(implicit system: ActorSystem) but such a signature doesn't exist and we need to use this signature instead : override def create(context: Context)(implicit system: ActorSystem)

Issue #345, opened 2016-12-12
Allow setting default server and client pipelines separately


anilgursel:
Currently, we only have `squbs.pipeline.streaming.defaults`.  But, sometimes the default on the server side and client side might need to be different.  So, consider having three options: 

- `squbs.pipeline.streaming.defaults`
- `squbs.pipeline.streaming.server.defaults`
- `squbs.pipeline.streaming.client.defaults`

Also, do we still need `streaming` in the config name?

anilgursel:
For simplicity, we decided to drop the common default key.  We are also dropping `streaming` and "s" in `defaults`.  So, we should have the following two keys:

`squbs.pipeline.server.default`
`squbs.pipeline.client.default`

Please see the corresponding discussion on gitter: https://gitter.im/paypal/squbs?at=5882995cd43728124e835b1d.

Issue #315, opened 2016-11-11
The setting `defaultPipelineOn` should be renamed to `defaultPipeline`


anilgursel:
Instead of `defaultPipelineOn = true`, we should use `defaultPipeline = on`.

akara:
Please also make sure we change the documentation and clearly show `defaultPipeline = on` is part of the `squbs-meta.conf`. Just show it in the example.

PR #393, opened 2017-02-07
#345 #376 #315 #392 Pipeline changes

anilgursel:

- [x] Commits are squashed.
- [x] Tests added.

- Seperate default configuration for client and server pipeline.
- Rename `defaultPipelineOn` to `defaultPipeline`.
- Move pipeline out of `streaming` package.
- Rename `squbs-streamingpipeline` to `squbs-pipeline`.
- Fix doc issue about `PipelineFactory` usage.


squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
```diff
@@ -86,9 +100,15 @@ object PipelineExtension extends ExtensionId[PipelineExtensionImpl] with Extensi
       flowMap = flowMap + (name -> flowFactory)
     }
 
-    val pre = Try(system.settings.config.getString("squbs.pipeline.streaming.defaults.pre-flow")).toOption
-    val post = Try(system.settings.config.getString("squbs.pipeline.streaming.defaults.post-flow")).toOption
-    new PipelineExtensionImpl(flowMap, pre, post)(system)
+    val serverDefaultPreFlow = Try(config.getString("squbs.pipeline.server.default.pre-flow")).toOption
```
akara: Shall we use `config.getOption[String](...)` instead? Its error handling is more concise than `Try(...).toOption`

anilgursel: We cannot!  Because `ConfigUtil` is in `squbs-ext` and there is no dependency from `squbs-pipeline` to `squbs-ext`.  And since `squbs-ext` already depends on `squbs-pipelnie` (for metric pipeline, etc), I do not want to cause a circular dependency.

akara: Ah, got it. Of course, the question is then whether that dependency is correct. However, that is a discussion we should have separately. Merging now.

Issue #342, opened 2016-12-11
EndpointResolver chain does not resolve URL


akara:
When using `ClientFlow` not passing a name, but a whole base URL, the `ClientFlow` should try to resolve it as name, and last check whether it is a URL and resolve that instead of giving up.

anilgursel:
Yes.  But, not sure if it should be enabled by default, or that we should register it in the templates by default.

anilgursel:
How should we register it?

- Should it be pre-registered by default in the chain? Being the default last one?
- Or it should not have any difference than any other resolver and should be registered at the template level in the extensions?

anilgursel:
Also, consider adding DefaultEndpointResolver that resolves baseUrl from application.conf and ssl from there as well, or even merge it with SimpleEndpointResolver.

PR #396, opened 2017-02-08
#342 Default HttpEndpointResolver

anilgursel:

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/squbs-httpclient/src/main/scala/org/squbs/httpclient/ClientFlow.scala
+++ b/squbs-httpclient/src/main/scala/org/squbs/httpclient/ClientFlow.scala
@@ -18,6 +18,7 @@ package org.squbs.httpclient
 
 import java.lang.management.ManagementFactory
 import java.util.Optional
+import java.util.concurrent.ConcurrentHashMap
 import javax.management.ObjectName
 
 import akka.actor.ActorSystem
@@ -44,6 +45,7 @@ object ClientFlow {
 
   val AkkaHttpClientCustomContext = "akka-http-client-custom-context"
   type ClientConnectionFlow[T] = Flow[(HttpRequest, T), (Try[HttpResponse], T), HostConnectionPool]
+  private[httpclient] val defaultResolverRegistrationRecord = new ConcurrentHashMap[String, Unit]
 
   def create[T](name: String, system: ActorSystem, mat: Materializer):
   js.Flow[Pair[jm.HttpRequest, T], Pair[Try[jm.HttpResponse], T], jd.HostConnectionPool] =
@@ -73,6 +75,12 @@ object ClientFlow {
               settings: Option[ConnectionPoolSettings] = None,
               env: Environment = Default)(implicit system: ActorSystem, fm: Materializer): ClientConnectionFlow[T] = {
 
+    defaultResolverRegistrationRecord.computeIfAbsent(system.name,
+      new java.util.function.Function[String, Unit] {
+        override def apply(t: String): Unit =
+          ResolverRegistry(system).register[HttpEndpoint](new DefaultHttpEndpointResolver)
+      })
+
     val environment = env match {
       case Default => EnvironmentResolverRegistry(system).resolve
       case _ => env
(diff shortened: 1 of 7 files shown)


squbs-httpclient/src/main/scala/org/squbs/httpclient/DefaultHttpEndpointResolver.scala
```diff
@@ -0,0 +1,42 @@
...
+  */
+class DefaultHttpEndpointResolver extends Resolver[HttpEndpoint] with LazyLogging {
+
+  override def name: String = getClass.getName
+
+  override def resolve(name: String, env: Environment): Option[HttpEndpoint] = {
+    Try(URI.create(name)).toOption match {
+      case Some(uri) if uri.getScheme == "http" || uri.getScheme == "https" => Some(HttpEndpoint(uri))
+      case _ =>
+        logger.debug(s"Could not resolve to an HttpEndpoint.  Invalid http URI: $name")
```
thefourtheye: Wouldn't it be better if we logged the actual error here?

squbs-httpclient/src/main/scala/org/squbs/httpclient/DefaultHttpEndpointResolver.scala
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2015 PayPal
```
thefourtheye: Why the Copyright is still 2015?

squbs-httpclient/src/main/scala/org/squbs/httpclient/DefaultHttpEndpointResolver.scala
anilgursel: This is not an error, and this resolve not being able to resolve to an endpoint is a valid use case.  Accordingly, I do not prefer logging the exception or the exception message.  For a happy path, I do not prefer to see an exception (even at debug level) in the logs.  It misleads during troubleshooting issues.

Also, there are two problems here:

1. Parsin error by `URI.create` which may throw an exception.
2. `uri.getScheme` might be valid but not `http` or `https`.

Providing detailed log for each use case is not worth the complexity (branching) of the code IMO.  That was my first implementation, then refactored and simplified it.

IMO, the log message is clear enough and also prints the `URI`, which makes troubleshooting easy enough.

squbs-httpclient/src/main/scala/org/squbs/httpclient/DefaultHttpEndpointResolver.scala
anilgursel: That's automatically added by my IntelliJ 😄  That's related to legal, which I will let @akara to comment on.

akara: Need to file an issue. Update all Copyright to 2017 before release. Will do.

squbs-httpclient/src/main/scala/org/squbs/httpclient/ClientFlow.scala
```diff
@@ -73,6 +75,12 @@ object ClientFlow {
               settings: Option[ConnectionPoolSettings] = None,
               env: Environment = Default)(implicit system: ActorSystem, fm: Materializer): ClientConnectionFlow[T] = {
 
+    import scala.compat.java8.FunctionConverters.asJavaFunction
+    defaultResolverRegistrationRecord.computeIfAbsent(system.name,
+      asJavaFunction((_: String) => {
```
akara: Will this line be a problem when we move to Scala 2.12? In some previous code, I went through the tedious work of keeping this an anonymous inner class instead of using closures and adapters thereof. It is not that hard, here.

squbs-httpclient/src/main/scala/org/squbs/httpclient/ClientFlow.scala
```diff
@@ -44,6 +45,7 @@ object ClientFlow {
 
   val AkkaHttpClientCustomContext = "akka-http-client-custom-context"
   type ClientConnectionFlow[T] = Flow[(HttpRequest, T), (Try[HttpResponse], T), HostConnectionPool]
+  val defaultResolverRegistrationRecord = new ConcurrentHashMap[String, Unit]
```
akara: Since this is mutable, should we at least do a `private[httpclient] defaultResolverRegistrationRecord`?

anilgursel: :+1:

squbs-httpclient/src/main/scala/org/squbs/httpclient/ClientFlow.scala
anilgursel: let me see.

PR #405, opened 2017-02-19
#401 Time window based auto cleanup for Persistent Buffer

### LINKED ISSUE #401: Issue with Persistent Buffer's time window based auto cleanup
- queue's roll cycles based on `filename` cleanup causes issues with time window shifts back 0th hour.

- For example, when queue files with name series 20,21,..23 validated for cleanup and outport's min complete on 23rd hour then files prior to 23, that is, 20..22 are taken for deletion.  

- But (Issue scenario) - when 1st hour data is processed, the data files prior to it 23rd in case of hourly cycle `does not` get deleted.

- The logic of using file name's numbers to delete the file is the problem. https://github.com/paypal/squbs/blob/master/squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala#L193

- Proposed Solution:
Instead of filename comprision, file's lastModifiedTime can be taken for comparision, which will allow deletion of old files without being caught due to naming issues.

queuedata-cleanup

SarathChandran:

- [X] Commits are squashed.
- [X] Tests added.

### DIFF
--- a/squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
+++ b/squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
@@ -16,14 +16,15 @@
 package org.squbs.pattern.stream
 
 import java.io.{File, FileNotFoundException}
+import java.util.function.{Function => JFunc}
 
 import com.typesafe.config.Config
 import com.typesafe.scalalogging.Logger
 import net.openhft.chronicle.bytes.MappedBytesStore
 import net.openhft.chronicle.core.OS
 import net.openhft.chronicle.queue.ChronicleQueueBuilder
-import net.openhft.chronicle.queue.impl.StoreFileListener
 import net.openhft.chronicle.queue.impl.single.SingleChronicleQueue
+import net.openhft.chronicle.queue.impl.{RollingResourcesCache, StoreFileListener}
 import net.openhft.chronicle.wire.{ReadMarshallable, WireIn, WireOut, WriteMarshallable}
 import org.slf4j.LoggerFactory
 
@@ -33,7 +34,7 @@ case class Event[T](outputPortId: Int, index: Long, entry: T)
   *
   * @tparam T The type of elements to be stored in the queue.
   */
-class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x => {}))(implicit val serializer: QueueSerializer[T]) {
+class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = _ => {})(implicit val serializer: QueueSerializer[T]) {
 
   def this(config: Config)(implicit serializer: QueueSerializer[T]) = this(QueueConfig.from(config))
 
@@ -47,15 +48,18 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
 
   val logger = Logger(LoggerFactory.getLogger(this.getClass))
 
-  private val queue = ChronicleQueueBuilder
+  private[stream] val resourceManager = new ResourceManager
+
+  private val builder = ChronicleQueueBuilder
     .single(persistDir.getAbsolutePath)
     .wireType(wireType)
     .rollCycle(rollCycle)
     .blockSize(blockSize.toInt)
     .indexSpacing(indexSpacing)
     .indexCount(indexCount)
-    .storeFileListener(new ResourceManager)
-    .build()
+    .storeFileListener(resourceManager)
+
+  private val queue = builder.build()
 
   private val appender = queue.acquireAppender
   private val reader = Vector.tabulate(outputPorts)(_ => queue.createTailer)
@@ -70,7 +74,21 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
   private val cycle = Array.ofDim[Int](outputPorts)
   private val lastCommitIndex = Array.ofDim[Long](outputPorts)
 
-  val totalOutputPorts = outputPorts
+  val totalOutputPorts: Int = outputPorts
+
+  import SingleChronicleQueue.SUFFIX
+
+  // `fileIdParser` will parse a given filename to its long value.
+  // The value is based on epoch time and grows incrementally.
+  // https://github.com/OpenHFT/Chronicle-Queue/blob/chronicle-queue-4.5.13/src/main/java/net/openhft/chronicle/queue/RollCycle.java#L35
+  private[stream] val fileIdParser = new RollingResourcesCache(queue.rollCycle(), queue.epoch(),
+    new JFunc[String, File] {
+      def apply(name: String) = new File(builder.path(), name + SUFFIX)
+    },
+    new JFunc[File, String] {
+      def apply(file: File): String = file.getName.stripSuffix(SUFFIX)
+    }
+  )
 
   private def mountIndexFile(): Unit = {
     indexFile = IndexFile.of(path, OS.pageSize())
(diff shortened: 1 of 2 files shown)


squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -186,17 +200,20 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
...
 
     private def minCycle = reader.iterator.map(_.cycle).min
 
-    private def fileNameToLong(file: File): Long = file.getName
-      .stripSuffix(SingleChronicleQueue.SUFFIX).split("-").last.toLong
-
-    private def deleteFiles(cycle: Int, fileNameInLong: Long): Unit = Option(persistDir.listFiles) foreach {
-      f => f.filterNot(f => f.getName.equals(Tailer) || fileNameToLong(f) >= fileNameInLong).foreach {
-        file => logger.info("File released {} - {}", cycle.toString, file.getPath)
+    private def deleteThisAndOlderFiles(cycle: Int, releasedFile: File): Unit = {
```
akara: This function should be named `deleteOlderFiles` as we no longer delete this file.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -186,17 +200,20 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
     }
 
     override def onReleased(cycle: Int, file: File): Unit =
-      if (minCycle >= cycle) deleteFiles(cycle, fileNameToLong(file))
+      if (cycle > minCycle && minCycle > 0) deleteThisAndOlderFiles(cycle, file)
```
akara: Odd. Did I miss anything here? Lets say current cycle is 10, min cycle is 5, you want to delete anything below 5. The release may not be on the min cycle, right?

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentQueueSpec.scala
```diff
@@ -145,4 +146,38 @@ class PersistentQueueSpec extends FlatSpec with Matchers {
...
+        queue.enqueue(element)
+        addToQueue(dataFiles.size, i + 1)
+      }
+    }
+
+    dataFiles should not be empty
+    val deleteThisAndOlderFiles = PrivateMethod[(Int, File)]('deleteThisAndOlderFiles)
+
+    // case 1: it should not cleanup unprocessed queue file
+    val oldestFile = dataFiles.head
```
akara: Do we have any guarantees that the oldest file is the first one in the list?

akara: It may just be the current OS. But I would not rely on that fact for a test.

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentQueueSpec.scala
```diff
@@ -145,4 +146,38 @@ class PersistentQueueSpec extends FlatSpec with Matchers {
...
+
+  it should "clean up buffer resources automatically" in {
+    val tempPath = Files.createTempDirectory("persistent_queue")
+    val queue = new PersistentQueue[ByteString](QueueConfig(tempPath.toFile, rollCycle = RollCycles.TEST_SECONDLY))
+
+    def dataFiles = tempPath.toFile.listFiles().toList.filterNot(_.getName == "tailer.idx")
+
+    addToQueue(0, 0)
+
+    def addToQueue(size: Int, i: Int): Unit = {
```
akara: Just for clarity, please have `@tailrec` above `addToQueue`

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
SarathChandran: had the change for testing, shouldn't have the logic changed. will revert it back.

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentQueueSpec.scala
```diff
@@ -145,4 +148,40 @@ class PersistentQueueSpec extends FlatSpec with Matchers {
...
+        addToQueue(dataFiles.size, i + 1)
+      }
+    }
+
+    dataFiles should not be empty
+    val deleteOlderFiles = PrivateMethod[(Int, File)]('deleteOlderFiles)
+
+    // case 1: it should not cleanup unprocessed queue file
+    val parser = queue.fileIdParser
+    val (oldestFile, fileLong) = dataFiles.map(f => f -> parser.toLong(f)).min
```
akara: Looks like we're just trying to find the oldest file. I don't think the `.min` at the end will correctly get you the min `_._2` for the tuple in all cases. Or at least I wont rely on that. Too obscure.

How about `dataFiles.minBy(parser.toLong(_))`?

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentQueueSpec.scala
```diff
@@ -145,4 +148,40 @@ class PersistentQueueSpec extends FlatSpec with Matchers {
...
+    queue.resourceManager invokePrivate deleteOlderFiles(0, dataFiles.filter(_ != oldestFile).head)
+    dataFiles.filter(f => f == oldestFile) shouldBe empty
+    dataFiles should not be empty
+
+    // case 2: it should cleanup processed queue file, except last file
+    val lastFile = dataFiles.last
```
akara: Is this supposed to be the newest file? We should probably name the variable `newestFile` and use similar semantics as `oldestFile` to establish it, something like:

```scala
val lastFile = dataFile.maxBy(parser.toLong(_))
```

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -186,17 +200,20 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = (x
...
-
-    private def deleteFiles(cycle: Int, fileNameInLong: Long): Unit = Option(persistDir.listFiles) foreach {
-      f => f.filterNot(f => f.getName.equals(Tailer) || fileNameToLong(f) >= fileNameInLong).foreach {
-        file => logger.info("File released {} - {}", cycle.toString, file.getPath)
+    private def deleteOlderFiles(cycle: Int, releasedFile: File): Unit = {
+      val files = Option(persistDir.listFiles).toArray.flatten.filterNot(_.getName == Tailer).toSeq
+      if (files.nonEmpty) {
+        val releasedFileLong = fileIdParser.toLong(releasedFile)
+        files.filter { file =>
+          file.getName.endsWith(SUFFIX) && (fileIdParser.toLong(file) < releasedFileLong)
```
anilgursel: So, even if the cycle roll overs the, `fileIdParser.toLong(file)` will always keep growing?  If so, can you please add a comment for this line and referencing to the corresponding Chronicle documentation or code

anilgursel:
Also, it would be better if the title gets updated as follows

> #401 Time window based auto cleanup for Persistent Buffer

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
SarathChandran: should i add comment at this line, guessing it will be better place https://github.com/paypal/squbs/pull/405/files#diff-717b79f6434ac986220216a2e1c86d40R79

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentQueueSpec.scala
```diff
@@ -145,4 +148,40 @@ class PersistentQueueSpec extends FlatSpec with Matchers {
...
+    val queue = new PersistentQueue[ByteString](QueueConfig(tempPath.toFile, rollCycle = RollCycles.TEST_SECONDLY))
+
+    def dataFiles = tempPath.toFile.listFiles().toList.filterNot(_.getName == "tailer.idx")
+
+    addToQueue(0, 0)
+
+    @tailrec
+    def addToQueue(size: Int, i: Int): Unit = {
+      if (size < 4) {
+        Thread.sleep(1)
```
anilgursel: `Thread.sleep` has caused too much trouble in the tests.  Can you please explain why it is needed?

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
anilgursel: I would prefer it to be closer to this line.  But, if you think otherwise, that's ok as well.

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentQueueSpec.scala
SarathChandran: used it to make sure that some queue files are generated and deletion of the same can be tested. On a regular testcase, we might hardly have 1 or 2 files generated.

anilgursel: Firstly, using `Thread.sleep` introduces undeterministic results in tests.  Secondly, the precision is around `10ms` so, sleeping just for `1 ms` should be same as not sleeping at all.

squbs-pattern/src/test/scala/org/squbs/pattern/stream/PersistentQueueSpec.scala
```diff
@@ -145,4 +148,40 @@ class PersistentQueueSpec extends FlatSpec with Matchers {
...
+        addToQueue(dataFiles.size, i + 1)
+      }
+    }
+
+    dataFiles should not be empty
+    val deleteOlderFiles = PrivateMethod[(Int, File)]('deleteOlderFiles)
+
+    // case 1: it should not cleanup unprocessed queue file
+    val parser = queue.fileIdParser
+    val oldestFile = dataFiles.minBy(parser.toLong(_))
```
akara: Just one more thing. It was actually my bad and IntelliJ told me we could write it as `dataFiles.minBy(parser.toLong)`. Basically passing method reference instead of closure.

Ditto for `maxBy` below.

Issue #418, opened 2017-03-08
Refactor Circuit Breaker API to accept a Envelope style id extractor


anilgursel:
Something similar to ConsistentHashMappable and envelop.

PR #420, opened 2017-03-11
#418 Refactor Circuit Breaker API to accept a Envelope style id extra…

anilgursel:
…ctor

- [x] Commits are squashed.
- [x] Tests added.


anilgursel:
Overall, I wonder if we are making it more complicated than it is supposed to be.  Instead of all this uniqueId extraction, we could also rely on `hashCode` and `equals`, can't we?

Just require that the `Context` has a unique id of some sort that app can guarantee and that gets used in `hashCode` and `equals`.  Isn't it already what's happening with `Int`, `Long` or `UUID`?

I actually tried using `MyContext(s: String, uuid: UUID)` without extending `UniqueId.Mapper`, without wrapping with `UniqueId.Envelope` and without using any `UniqueId.Mapper`, functionality just worked as expected.

The only problem would be if `Context` object is not in the control of the app itself.  In such scenarios, we can suggest using `UniqueId.Envelope`.

Am I missing something?

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -282,27 +252,32 @@ object TimeoutBidiUnordered {
...
   *             This then gets wrapped with a [[Try]] and pushed downstream with a [[Context]]
-  * @tparam Context the type of the context that is carried around along with the elements.  The context may be of any
-  *                 type and may contain any information given that it also contains information that is retrieved
-  *                 via `uniqueId` function to uniquely identify each element.
-  * @tparam Id Id the type that is used to uniquely identify each element.
+  * @tparam Context the type of the context that is carried around along with the elements.
   */
-final class TimeoutBidiUnordered[In, Out, Context, Id](timeout: FiniteDuration, uniqueId: Context => Id) extends
- GraphStage[BidiShape[(In, Context), (In, Context), (Out, Context), (Try[Out], Context)]]{
+final class TimeoutBidiUnordered[In, Out, Context](timeout: FiniteDuration, uniqueIdMapping: Mapping)
```
anilgursel: This is not a `case` class.. And the only reason is just to align with standard Akka Streams operators.  Otherwise, I would make this a `case` class and remove the companion.

Do you know why they are not `case class`?

squbs-ext/src/main/scala/org/squbs/circuitbreaker/CircuitBreakerBidi.scala
```diff
@@ -189,27 +189,13 @@ object CircuitBreakerBidi {
...
-    * @param failureDecider the function that gets called to determine if an element passed by the joined [[Flow]] is
-    *                       actually a failure or not
-    * @tparam In the type of the elements pulled from the upstream along with the [[Context]] and pushed down to joined
-    *            flow
-    * @tparam Out the type that's contained in a [[Try]] and pushed downstream along with the [[Context]]
-    * @tparam Context the type of the context that is carried around along with the elements.  The context may be of any
-    *                 type that can be used to uniquely identify each element
-    * @return a [[CircuitBreakerBidiFlow]] that can be joined with a [[Flow]] with corresponding types to add timeout
-    *         functionality.
+    * @see [[apply]] that takes a [[UniqueId.Mapper]] parameter for details.
```
anilgursel: Should `CircuitBreakerBidi` move under package `org.squbs.streams` instead of `org.squbs.circuitbreaker`?    It kind looks odd that `TimeoutBidi` is under `org.squbs.streams`.

The problem with circuit breaker though is, it has implementation class that are in `impl` package.. But, we can also do `org.squbs.streams.circuitbreaker`.

docs/flow-timeout.md
```diff
@@ -92,54 +92,88 @@ Source.from(Arrays.asList("a", "b", "c"))
...
-   
+
     public UUID uuid() {
         return uuid;
     }
 }
 
-final BidiFlow<Pair<String, MyContext>, Pair<String, MyContext>, Pair<String, MyContext>, Pair<Try<String>, MyContext>, NotUsed> timeoutBidiFlow =
-        TimeoutBidiFlowUnordered.create(timeout, MyContext::uuid);
+final UniqueId.Mapper myUniqueIdMapper = message -> {
```
akara: Does this work? I don't see a `FunctionalInterface`. In that case it should not be assignable from a Java closure. What did I miss?

docs/flow-timeout.md
```diff
@@ -92,54 +92,88 @@ Source.from(Arrays.asList("a", "b", "c"))
...
+
+    if(message instanceof MyContext) return ((MyContext) message).uuid;
+    return null;
```
akara: For some odd reason, I'm a bit allergic to returning a null 😆

squbs-ext/src/main/scala/org/squbs/circuitbreaker/CircuitBreakerBidi.scala
akara: I'd go for the latter.

squbs-ext/src/main/scala/org/squbs/circuitbreaker/CircuitBreakerBidi.scala
```diff
@@ -235,45 +221,56 @@ object CircuitBreakerBidiFlow {
     * Creates a [[BidiFlow]] that can be joined with a [[Flow]] to add Circuit Breaker functionality.
     *
     * Circuit Breaker functionality requires each element to be uniquely identified, so it requires a [[Context]], of
-    * any type defined by the application, to be carried along with the [[Flow]]'s input and output as a tuple.
+    * any type defined by the application, to be carried along with the flow's input and output as a tuple.  The
```
akara: as a `akka.japi.Pair`. This is the Java API.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
akara: It is not intended for case matching or extractions.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -282,27 +252,32 @@ object TimeoutBidiUnordered {
...
+    }
+  }
+
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new TimeoutGraphStageLogic(shape) {
 
-    val timeouts = mutable.LinkedHashMap.empty[Id, (Context, Long)]
+    val timeouts = mutable.LinkedHashMap.empty[Any, (Context, Long)]
     val readyToPush = mutable.Queue[((Try[Out], Context), Long)]()
 
     override protected def timeoutDuration: FiniteDuration = timeout
```
akara: Marker - finished reviewing through here. To be continued.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -282,27 +252,32 @@ object TimeoutBidiUnordered {
...
 
   private val in = Inlet[(In, Context)]("TimeoutBidiUnordered.in")
   private val fromWrapped = Inlet[(Out, Context)]("TimeoutBidiUnordered.fromWrapped")
   private val toWrapped = Outlet[(In, Context)]("TimeoutBidiUnordered.toWrapped")
   private val out = Outlet[(Try[Out], Context)]("TimeoutBidiUnordered.out")
   val shape = BidiShape(in, toWrapped, fromWrapped, out)
 
+  private[streams] def uniqueId(context: Context) = {
+    context match {
+      case _ if uniqueIdMapping.isDefinedAt(context) ⇒ uniqueIdMapping(context)
```
akara: Don't quite like our use of `PartialFunction` here. It causes double evaluation which is sometimes hard to avoid. A similar design of `T => Option[T]` would not cause the same. It requires you to make 2 layer matches, but may be better than double-evaluation.

squbs-ext/src/main/scala/org/squbs/streams/UniqueId.scala
```diff
@@ -0,0 +1,59 @@
...
+  /**
+    * Mapping from message to a unique id
+    */
+  trait Mapper {
+    def uniqueId(message: Any): Any
+  }
+
+  private[squbs] def mappingAdapter(mapper: Mapper): Mapping = {
+    case message if mapper.uniqueId(message).asInstanceOf[AnyRef] ne null ⇒
+      mapper.uniqueId(message)
```
akara: This is also related to my comment above. Instead of doing `PartialFunction[Any, Any]`, we could do `type Mapping = Any => Option[Any]`. Then the `mappingAdapter` can just be `message => Option(mapper.uniqueId(message))`

This will ensure the `mapper.uniqueId(message)` is being executed once.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
anilgursel: Fair enough.  It would save us implementing our own companions though 😄

squbs-ext/src/main/scala/org/squbs/circuitbreaker/CircuitBreakerBidi.scala
anilgursel: So, `org.squbs.streams.circuitbreaker` it is.

docs/circuitbreaker.md
```diff
@@ -165,31 +165,49 @@ final CircuitBreakerState circuitBreakerState =
...
+  failureDecider: Option[((Try[Out], Context)) => Boolean],
+  uniqueIdMapper: Mapper)
 ```
-Here is an example with a custom unique id retriever:
+Here is an example with a `UniqueId.Mapper`:
 
 ```scala
 case class MyContext(s: String, uuid: UUID)
 
+val myUniqueIdMapper = new Mapper {
```
anilgursel: 👍

docs/flow-timeout.md
```diff
@@ -92,54 +92,88 @@ Source.from(Arrays.asList("a", "b", "c"))
...
 
 ###### Java
 
+The following API is used to pass a `UniqueId.Mapper`:
+
 ```java
-TimeoutBidiFlowUnordered
-    .create[In, Out, Context, Id]
-    (timeout: FiniteDuration, uniqueId: java.util.function.Function[Context, Id])
+TimeoutBidiFlowUnordered.create[In, Out, Context](timeout: FiniteDuration, uniqueIdMapper: Mapper)
```
anilgursel: I agree on the signature change.

Not sure on the `FiniteDuration` vs `ms`.  Check `delay` usage in Java DSL.  It takes `FiniteDuration`.  While I agree it is Scala artifact and might be unnatural for Java, I would prefer to better align with other Akka Streams stages.

docs/flow-timeout.md
anilgursel: All examples are from unit-tests.  So, yes it works.

anilgursel: http://download.java.net/lambda/b78/docs/api/java/lang/FunctionalInterface.html

> However, the compiler will treat any interface meeting the definition of a functional interface as a functional interface regardless of whether or not a FunctionalInterface annotation is present on the interface declaration.

docs/flow-timeout.md
anilgursel: Me too.  

It is for this logic here https://github.com/anilgursel/squbs/blob/231f2d05f1c3227e8e51778f5cc7ebf5f4c562b2/squbs-ext/src/main/scala/org/squbs/streams/UniqueId.scala#L56.  

Let me try to eliminate.

squbs-ext/src/main/scala/org/squbs/streams/UniqueId.scala
```diff
@@ -0,0 +1,59 @@
...
+
+    def apply(x: Any) = throw new UnsupportedOperationException("Empty ConsistentHashMapping apply()")
+  }
+
+  /**
+    * Mapping from message to a unique id
+    */
+  trait Mapper {
+    def uniqueId(message: Any): Any
+  }
```
anilgursel: Not necessarily.  Both can use both.  That was of my comments during our discussion.  `Mapping` is needed when the class in hand is from a third party and cannot extend `Mapper`.

squbs-ext/src/main/scala/org/squbs/streams/UniqueId.scala
anilgursel: 👍

squbs-ext/src/test/java/org/squbs/circuitbreaker/CircuitBreakerBidiFlowTest.java
```diff
@@ -269,8 +276,14 @@ public long next() {
                         FiniteDuration.apply(10, TimeUnit.MILLISECONDS),
                         system.dispatcher());
 
+        final UniqueId.Mapper myMapper = message -> {
```
anilgursel: Answered in one of the previous comments.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
anilgursel: I pretty much followed the pattern from Akka source code.  Will change the design.

anilgursel:
Any comments on https://github.com/paypal/squbs/pull/420#issuecomment-285882554 ?

docs/circuitbreaker.md
```diff
@@ -165,31 +165,47 @@ final CircuitBreakerState circuitBreakerState =
...
 ```scala
-CircuitBreakerBidiFlow[In, Out, Context, Id](circuitBreakerState: CircuitBreakerState,
-                                             fallback: Option[((In, Context)) => (Try[Out], Context)],
-                                             failureDecider: Option[((Try[Out], Context)) => Boolean],
-                                             uniqueId: Context => Id)
+CircuitBreakerBidiFlow[In, Out, Context](
+  circuitBreakerState: CircuitBreakerState,
+  fallback: Option[((In, Context)) => (Try[Out], Context)],
+  failureDecider: Option[((Try[Out], Context)) => Boolean],
+  uniqueIdMapper: Any => Option[Any])
```
anilgursel: Do you think it looks odd?  `fallback` and `failureDecider` are `Option` but `uniqueIdMapper` is not..  Just trying to remember why I did it this ways; however, I do not remember.

akara: You don't need to provide a fallback, so it is an `Option`
There is a default behavior to decide for a failure, so providing your own decider is an `Option`, too.
But you do need to be able to get a unique id out of the context. That's why the last is not an `Option`.

The only thought coming out if this is why uniqueIdMapper is an `Any => Option[Any]`? I need to go through the code again, but would `Context => Option[Any]` make more sense?

Don't worry, perhaps I'm totally wrong. Let me go read the code once again.

akara: Yes, I did go read the code. It is only used against the `Context` type. The resulting unique id type is certainly an `Any`. But the logic is trying whether this mapper works, otherwise it falls back to using the context object directly. Just wondering what complications it will create if we scope it down to `Context => Option[Any]`, or any other reason we leave the input type totally open.

akara:
Other than just this last one comment to be clarified, the PR looks good to me.

docs/circuitbreaker.md
anilgursel: You're right.  I do not see a reason for it to be `Any`..  That's probably an oversight I did while refactoring etc..  I will change it to `Context => Option[Any]`.

It actually makes it more confusing to read the API with `Any => Option[Any]`.

akara:
LGTM. Please squash and I'll merge.

squbs-ext/src/test/java/org/squbs/streams/TimeoutBidiFlowTest.java
```diff
@@ -38,10 +38,13 @@
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
```
akara: @anilgursel Can you please check this Codacy warning real quick? Was about to merge and somehow just saw it.

squbs-ext/src/test/java/org/squbs/streams/circuitbreaker/CircuitBreakerBidiFlowTest.java
```diff
@@ -47,6 +47,8 @@
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
```
akara: Here too.

PR #430, opened 2017-03-22
#429 `HttpEndpoint` should take an optional `Config`

anilgursel:

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpEndpoint.scala
+++ b/squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpEndpoint.scala
@@ -19,6 +19,8 @@ import java.net.URI
 import java.util.Optional
 import javax.net.ssl.SSLContext
 
+import com.typesafe.config.Config
+
 import scala.compat.java8.OptionConverters._
 
 /**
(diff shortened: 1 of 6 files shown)


akara:
LGTM. Just awaiting build to pass and I'll merge.

squbs-httpclient/src/main/scala/org/squbs/httpclient/HttpEndpoint.scala
```diff
@@ -19,22 +19,25 @@ import java.net.URI
...
+case class HttpEndpoint(uri: URI, sslContext: Option[SSLContext] = None, config: Option[Config] = None)
 
 /**
   * Creators for HttpEndpoint.
   */
 object HttpEndpoint {
   def apply(s: String): HttpEndpoint = HttpEndpoint(new URI(s))
 
-  def apply(s: String, sslContext: Option[SSLContext]): HttpEndpoint = HttpEndpoint(new URI(s), sslContext)
+  def apply(s: String, sslContext: Option[SSLContext], config: Option[Config]): HttpEndpoint =
```
akara: @anilgursel In retrospect, I'm wondering whether we should have default values with `apply` and provide backwards compatible signatures of `create`. It would have avoided the broken internal code today (recompile was a must).

It also slipped my mind when doing code reviews. But we can still add these signatures.

anilgursel: I thought about it actually.  But, requires more combinations of `create`.  Given it is still in snapshot, it is ok.  If we were doing this after the release, we would definitiely provide backward compatibility.

PR #439, opened 2017-03-27
#424 Java API: JUnit & TestNG versions of RouteTest and CustomRouteTestKit

akara:
* Added JUnit & TestNG versions of RouteTest and CustomRouteTestKit
* Refactored AbstractCustomRouteTestKit to CustomRouteTestKit.
* Removed old delegation model CustomRouteTestKit
* Refactored all tests using CustomRouteTestKit with delegation model.
* Ensured all shutdown happen properly.

- [X] Commits are squashed.
- [X] Tests added.


squbs-testkit/src/main/scala/org/squbs/testkit/japi/CustomRouteTestKit.scala
```diff
@@ -0,0 +1,63 @@
...
+import akka.stream.ActorMaterializer
+import akka.testkit.TestKitBase
+import com.typesafe.config.Config
+import org.junit.Rule
+import org.squbs.testkit.{DebugTiming, PortGetter, CustomTestKit => SCustomTestKit}
+import org.squbs.unicomplex.UnicomplexBoot
+
+import scala.collection.JavaConversions._
+
+abstract class CustomRouteTestKit(val boot: UnicomplexBoot) extends {
```
anilgursel: We have `AbstractCustomTestKit`, so should we rename this as `AbstractCustomRouteTestKit` as well?

squbs-testkit/src/test/java/org/squbs/testkit/japi/CustomRouteTestKitConfigResourcesTest.java
```diff
@@ -0,0 +1,70 @@
...
+    static List<String> resources = Arrays.asList(CustomRouteTestKitConfigResourcesTest.class.getClassLoader()
+            .getResource("").getPath() + "/CustomTestKitDefaultSpec/META-INF/squbs-meta.conf");
+    static Config customConfig = ConfigFactory.parseString("CustomRouteTestKitConfigResourcesTest.foobar = baz");
+
+    public CustomRouteTestKitConfigResourcesTest() {
+        super(customConfig, resources, false);
+    }
+
+    @Test
+    public void testSimpleRoute() {
```
anilgursel: I believe we also should have test which is completely outside of the context of `CustomRouteTestKit`..  Without unicomplex being started etc..

squbs-testkit/src/main/scala/org/squbs/testkit/japi/CustomRouteTestKit.scala
sebady: There is actually a bit in common with AbstractCustomTestKit  here. Should we consider having this class extend `AbstractCustomTestKit`?

akara: Thanks for both of your comments.
@anilgursel I can definitely do so, but doesn't this get a bit long?
@sebady I'm not sure it will make a lot of sense. If you look at the content of `AbstractCustomTestKit`, it basically delegates to the Scala `CustomTestKit` and adds the `PortGetter` trait. It does not have any functionality by itself (lots of constructors, that's it). We need to consider what a type relationship will give us in such cases. When in doubt, I generally tend to stay away from subclasses. Or, subclass if and only if you're 100% sure there is a "is a" relationship.

anilgursel: Haha, you're reminding me the Spring classes with long names 😄 

I think it is not that long.  I'd rather prefer the consistency in the naming across.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
```diff
@@ -464,7 +464,7 @@ object UnicomplexBoot extends LazyLogging {
   def startServiceInfra(boot: UnicomplexBoot)(implicit actorSystem: ActorSystem) {
     import actorSystem.dispatcher
     val startTime = System.nanoTime
-    implicit val timeout = Timeout((boot.listeners.size * 5) seconds)
+    implicit val timeout = Timeout((boot.listeners.size * 10) seconds)
```
akara: With all these tests and JUnit starting a Unicomplex for each test, I'm running into sporadic timeouts testing on my mac on this line. Hence increasing the timeout.

squbs-testkit/build.sbt
```diff
@@ -11,10 +11,10 @@ libraryDependencies ++= Seq(
   "com.typesafe.akka" %% "akka-testkit" % akkaV,
   "com.typesafe.akka" %% "akka-http-testkit" % akkaHttpV % "provided",
   "org.scala-lang.modules"     %% "scala-java8-compat"  % "0.7.0",
-  "junit" % "junit" % "4.12" % "test",
+  "junit" % "junit" % "4.12" % "provided",
```
anilgursel: Should we move junit and testng versions to https://github.com/paypal/squbs/blob/master/project/Versions.scala#L24 ?

squbs-testkit/src/main/scala/org/squbs/testkit/japi/TestNGRouteTest.scala
```diff
@@ -0,0 +1,111 @@
...
+  * and their lifetime is automatically managed.
+  */
+trait TestNGRouteTestBase extends RouteTest with RouteDefinitionTest with TestNGSuiteLike  {
+  protected def systemResource: SystemResource
+  implicit def system: ActorSystem = systemResource.system
+  implicit def materializer: Materializer = systemResource.materializer
+
+  protected def createTestRouteResult(request: HttpRequest, result: RouteResult): TestRouteResult =
+    new TestRouteResult(result, awaitDuration)(system.dispatcher, materializer) {
+      protected def assertEquals(expected: AnyRef, actual: AnyRef, message: String): Unit =
```
anilgursel: Not clear, why we need `assertEquals` wrapping `Assert.assertEquals`?  Also, the position of `expected` and `actual` are swapped between the two methods.

akara: This needed to be done for every test framework and enables this API:

```java
simpleRoute.run(HttpRequest.GET("/ping"))
    .assertStatusCode(200)
    .assertEntity("pong");
```

These assertions again call the assertion entries provided here.

Now, field positions in TestNG and JUnit assertions are swapped. 

* In TestNG, it is `assertEquals(actual, expected, message)`
* In JUnit, it is `assertEquals(message, expected, actual)`

This is why you see the swapping. While having the wrong position logically works (swapping expected and actual), it will have issues with the generated error message and confuse you when tests fail.

It is wise to follow the framework's signature.

docs/testing.md
```diff
@@ -163,25 +169,53 @@ class SampleSpec extends CustomTestKit(SampleSpec.boot) with FunSpecLike with Ma
...
+
+```java
+    @After
+    public void tearDown() {
+        shutdown();
+    }
 ```
 
-For Akka-Http:
+Note: If you construct the `CustomTestKit` passing a `UnicomplexBoot` object on the `super(boot)` call, use caution how and when to shutdown. If the `UnicomplexBoot` instance is created per-class, meaning one single instance is used for all tests, the shutdown also needs to happen only once. Use JUnit's `@AfterClass` annotation to annotate a static shutdown method. But if the `UnicomplexBoot` instance is created per-test instance, the `@After` annotation should be used similar to default construction of `CustomTestKit` (the default).
```
anilgursel: Is "per-test " the right terminology?  If so, I am ok.  I feel "per test method" is more clear.  Same comment for "... one single instance is used for all tests ..."

docs/testing.md
```diff
@@ -163,25 +169,53 @@ class SampleSpec extends CustomTestKit(SampleSpec.boot) with FunSpecLike with Ma
...
+
+## Testing Akka Http Routes using Akka Http TestKit
+
+The `akka-http-testkit` needs to be added to the dependencies in order to test routes.  Please add the followings to your dependencies:
 
 ```
 "com.typesafe.akka" %% "akka-http-testkit" % akkaV % "test"
```
anilgursel: Shouldn't this be `akkaHttpV`?

docs/testing.md
```diff
@@ -163,25 +169,53 @@ class SampleSpec extends CustomTestKit(SampleSpec.boot) with FunSpecLike with Ma
...
 ```
 
 ### Usage
-As you specify routes in squbs by extending the `RouteDefinition` trait which squbs will compose with actors behind
-the scenes, it can be difficult to construct routes for use with the Spray/Akka Http TestKit test DSL. `TestRoute` is provided
+As you specify routes in squbs by extending the `RouteDefinition` trait *(Scala)* or the `AbstractRouteDefinition` class *(Java)* which squbs will compose with actors behind
```
anilgursel: Do we really need to specify the detail of "which squbs will compose with actors behind the scenes" ?  It was more important Spray world; however, with the recent changes in the design, it might be confusing?

docs/testing.md
```diff
@@ -163,25 +169,53 @@ class SampleSpec extends CustomTestKit(SampleSpec.boot) with FunSpecLike with Ma
...
+the scenes, it can be difficult to construct routes for use with the Akka Http TestKit test DSL alone. The squbs testkit provides utilities to deal with such routes in both standalone and full system mode where the infrastructure and cubes can be brought up as part of the test.
```
anilgursel: > it can be difficult to construct routes for use with the Akka Http TestKit test DSL alone. 

Do we need to give a negative sense here?  I do not think it is because it is difficult, it is just a utility..  I would just explain how to do it, and mention no comments regarding difficulty.

docs/testing.md
```diff
@@ -212,9 +245,9 @@ Implementing the test, obtaining route from `TestRoute[MyRoute]`:
 ```scala
 package com.myorg.mypkg
```
anilgursel: The package and the empty line is unnecessary.  Can we delete those?

docs/testing.md
```diff
@@ -192,7 +226,6 @@ Specifying the `RouteDefinition`
...
```
anilgursel: The package and the empty line is unnecessary.  Can we delete those?

docs/testing.md
```diff
@@ -163,25 +169,53 @@ class SampleSpec extends CustomTestKit(SampleSpec.boot) with FunSpecLike with Ma
...
 "com.typesafe.akka" %% "akka-http-testkit" % akkaV % "test"
 ```
 
 ### Usage
-As you specify routes in squbs by extending the `RouteDefinition` trait which squbs will compose with actors behind
-the scenes, it can be difficult to construct routes for use with the Spray/Akka Http TestKit test DSL. `TestRoute` is provided
+As you specify routes in squbs by extending the `RouteDefinition` trait *(Scala)* or the `AbstractRouteDefinition` class *(Java)* which squbs will compose with actors behind
+the scenes, it can be difficult to construct routes for use with the Akka Http TestKit test DSL alone. The squbs testkit provides utilities to deal with such routes in both standalone and full system mode where the infrastructure and cubes can be brought up as part of the test.
+
+##### Scala
```
anilgursel: Not strong about this..  However, Scala and Java looks small on the page.  Since the parent has "###", we could do this "####"?  Applies to the other example in the page.

docs/testing.md
```diff
@@ -236,14 +269,76 @@ Alternatively, you may also want to pass a web context to your route. This can b
 
 or just pass `"mycontext"` without the parameter name. The `TestRoute` signature without parameters is equivalent to passing the root context `""`.
 
+##### Java
+
+In order to test the route, the test needs to extend the proper `RouteTest` abstract class for the test framework of your choice. Choose to extend `org.squbs.testkit.japi.TestNGRouteTest` if you are using TestNG, or `org.squbs.testkit.japi.JunitRouteTest` if you're using Junit. The usage for both are the same.
```
anilgursel: I assume the class name is actually `JUnitRouteTest`, upper case 'U' after 'J'.

docs/testing.md
```diff
@@ -236,14 +269,76 @@ Alternatively, you may also want to pass a web context to your route. This can b
...
+                )
+        );
+    }
+}
+```
+
+Implementing the test, obtaining route from `testRoute(MyRoute.class)`:
+
+```java
+package com.myorg.mypkg;
```
anilgursel: Ditto for `package` comment above..

docs/testing.md
```diff
@@ -236,14 +269,76 @@ Alternatively, you may also want to pass a web context to your route. This can b
...
+
 ### Using TestRoute with CustomTestKit
 
 A need may arise to bootstrap `Unicomplex` while testing with `TestRoute`, such as when:
 
    * a squbs well-known actor is involved in request handling.
-   * [The Actor Registry](registry) is used during request handling.
+   * [The Actor Registry](registry.md) is used during request handling.
+
+Using Akka's `TestKit` together with `ScalatestRouteTest` can be tricky as they have conflicting initialization.  squbs provides a test utilities named `CustomRouteTestKit` *(Scala)*, `TestNGCustomRouteTestKit`, and `JUnitCustomRouteTestKit` *(Java, for each test framework)* to solve this problem.  `CustomRouteTestKit` supports all the APIs provided by `CustomTestKit`.  Here are example usages of `TestRoute` with `CustomRouteTestKit`:
```
anilgursel: Typo "s a test utilities"..

docs/testing.md
```diff
@@ -163,25 +169,53 @@ class SampleSpec extends CustomTestKit(SampleSpec.boot) with FunSpecLike with Ma
...
-The `spray-testkit` or `akka-http-testkit` needs to be added to the dependencies in order to test routes.  Please add the followings to your dependencies:
+The `CustomTestKit` shuts down automatically, unless you override the `afterAll()` method. No further actions are needed.
 
-For Spray:
+#### TestNG
 
+Use the `@AfterClass` annotation to annotate a method calling `shutdown()` as follows:
+
+```java
+    @AfterClass
```
akara: Not sure how to handle this comment. IMO, this behavior makes TestNG much easier to use, more like ScalaTest. It instantiates one instance of the test class, runs all tests, and calls `@AfterClass` with the instance still intact and available. Allows you to cleanup more easily. The  only quirk is that the annotation should rather be named `@AfterAll` similar to ScalaTest's `BeforeAndAfterAll` trait.

sebady: JUnit's "unexpected" behavior in this are is one of the reason testng was "conceived".  See

docs/testing.md
akara: Good catch!

docs/testing.md
akara: No we don't. But when you work without brains like sometimes I do, you'd miss all this. I'm lucky we have a great reviewer 😄

docs/testing.md
akara: Another brainless left-behind. Thanks!

docs/testing.md
akara: I always use 5 hashes for the Scala and Java label in recent documentation to keep it consistent. Not feeling too strong about it, but just wanted to keep the consistency. If you poke me on this again I'll make the change 😆

docs/testing.md
anilgursel: Yes, I agree on TestNG execution making more sense.  My comment was specifically about annotation being names as `AfterClass` and not requiring the class to be `static`, `@AfterAll` would make even more sense.

PR #486, opened 2017-07-31
Retry stage 417

sebady:

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/docs/flow-retry.md
+++ b/docs/flow-retry.md
@@ -1,0 +1,169 @@
+# Retry stage
+
+### Overview
+
+Some stream use cases may require retrying of requests after a failure response.  squbs provides a `RetryBidi` Akka Streams stage to add a retry capability to
+ streams that need to add retries for any failing requests.
+
+### Dependency
+
+Add the following dependency to your `build.sbt` or scala build file:
+
+```
+"org.squbs" %% "squbs-ext" % squbsVersion
+```
+
+### Usage
+
+The retry stage functionality is provided via a `BidiFlow` that can be connected to flows via the `join` operator.  The retry `BidiFlow` will perform some
+ specified maximum number of retries of any failures from downstream. A failure is determined by either a passed in failure decider function or a Failure,
+ if a failure decider function is not supplied.
+
+If all the retried attempts fail then the last failure for that request is emitted.  The retry stage requires a `Context` to be carried around for each
+ element.  This is needed to uniquely identify each element (see [Context to Unique Id Mapping](#context-to-unique-id-mapping) section for more details).
+
+##### Scala
+
+```scala
+val retryBidi = RetryBidi[String, Long](maxRetries = 10)
+val flow = Flow[Try[String], Long].map(s => findAnEnglishWordThatStartWith(s))
+
+Source("a" :: "b" :: "c" :: Nil)
+  .via(retryBidi.join(flow))
+  .runWith(Sink.seq)
+```
+
+##### Java
+
+```java
+final BidiFlow<Pair[String, Context], Pair[String, Context],
+    Pair[Try[String], Context], Pair[Try[String], Context], NotUsed> retryBidi =
+    RetryBidi.create(2);
+
+final Flow<String, String, NotUsed> flow =
+    Flow.<String>create().map(s -> findAnEnglishWordThatStartWith(s));
+
+Source.from(Arrays.asList("a", "b", "c"))
+    .via(retryBidi.join(flow))
+    .runWith(Sink.seq(), mat);
+```
+
+#### Context to Unique Id Mapping
+
+The `Context` type itself might be used as a unique id.  However, in many scenarios, `Context` contains more than the unique id itself or the unique id might
+ be retrieved as a mapping from the `Context`.  squbs allows different options to provide a unique id:
+
+   * `Context` itself is a type that can be used as a unique id, e.g., `Int`, `Long`, `java.util.UUID`
+   * `Context` extends `UniqueId.Provider` and implements `def uniqueId`
+   * `Context` is wrapped with `UniqueId.Envelope`
+   * `Context` is mapped to a unique id by calling a function
+
+With the first three options, a unique id can be retrieved directly through the context.
+
+For the last option, `RetryBidi` allows a function to be passed in as a parameter.
+
+###### Scala
+
+The following API can be used to pass a uniqueId mapper:
+
+```scala
+RetryBidi[In, Out, Context](maxRetries: Int, uniqueIdMapper: Context => Option[Any])
+```
+
+This `BidiFlow` can be joined with any flow that takes in a `(In, Context)` and outputs a `(Out, Context)`.
+
+```scala
+case class MyContext(s: String, uuid: UUID)
+
+val retryBidi = RetryBidi[String, String, MyContext](maxRetries = 2, (context: MyContext) => Some(context.uuid))
+val flow = Flow[(String, MyContext)].mapAsyncUnordered(10) { elem =>
+  (ref ? elem).mapTo[(String, MyContext)]
+}
+
+Source("a" :: "b" :: "c" :: Nil)
+  .map( _ -> MyContext("dummy", UUID.randomUUID))
+  .via(retryBidi.join(flow))
+  .runWith(Sink.seq)
+```
+
+###### Java
+
+The following API is used to pass a uniqueId mapper:
+
+```java
+public class RetryBidi {
+    public static <In, Out, Context> BidiFlow<Pair<In, Context>,
+                                              Pair<In, Context>,
+                                              Pair<Try<Out>, Context>,
+                                              Pair<Try<Out>, Context>,
+                                              NotUsed>
+    create(Long maxRetries, Function<Context, Optional<Object>> uniqueIdMapper);
+}
+```
+
+This `BidiFlow` can be joined with any flow that takes in a `akka.japi.Pair<In, Context>` and outputs a `akka.japi.Pair<Try<Out>, Context>`.
+
+```java
+class MyContext {
+    private String s;
+    private UUID uuid;
+
+    public MyContext(String s, UUID uuid) {
+        this.s = s;
+        this.uuid = uuid;
+    }
+
+    public UUID uuid() {
+        return uuid;
+    }
+};
+
+final BidiFlow<Pair<String, MyContext>, Pair<String, MyContext>,
+               Pair<Try<String>, MyContext>, Pair<Try<String>, MyContext>, NotUsed>
+                retryBidi = RetryBidi.create(maxRetries, context -> Optional.of(context.uuid));
+
+final Flow<Pair<String, MyContext>, Pair<String, MyContext>, NotUsed> flow =
+        Flow.<Pair<String, MyContext>>create()
+                .mapAsyncUnordered(10, elem -> ask(ref, elem, 5000))
+                .map(elem -> (Pair<String, MyContext>)elem);
+
+Source.from(Arrays.asList("a", "b", "c"))
+        .map(s -> new Pair<>(s, new MyContext("dummy", UUID.randomUUID())))
+        .via(retryBidi.join(flow))
+        .runWith(Sink.seq(), mat);
+
+```
+
+#### Failure decider
+
+By default, any `Failure` from the joined `Flow` is considered a failure for retry purposes.  However, the `Retry` stage also accepts an optional
+ `failureDecider` parameter to more finely control what elements from the joined `Flow` should actually be treated as failures that should be retried.
+
+##### Scala
+
+A function of type `Try[Out] => Boolean` can be provided via the `failureDecider` parameter. Below is an example where, along with any `Failure` message,
+Any response with failing http status code is also considered a failure:
+
+```scala
+
+val failureDecider = (tryResponse: Try[HttpResponse]) => tryResponse.isFailure || tryResponse.get().status().isFailure()
+
+val retryBidi = RetryBidi[Request, Response, MyContext](maxRetries = 3, (context: MyContext) => Some(context.uuid), failureDecider = Option(failureDecider), OverflowStrategy.backpressure())
+
+```
+
+##### Java
+
+A `Function<Try<Out>, Boolean>` can be provided via Retry Optional `failureDecider` parameter to create method.  Below is an example where,
+along with any `Failure` message, a `Success` of `HttpResponse` with status code `400` and above is also considered a failure:
+
+```java
+
+final Function<Try<HttpResponse>, Optional<Object>> failureDecider =
+tryResponse -> tryResponse.isFailure() || tryHttpResponse.get().status().isFailure());
+
+final BidiFlow<Pair<HttpResponse, MyContext>, Pair<HttpResponse, MyContext>, Pair<Try<HttpResponse>, MyContext>,
+    Pair<Try<HttpResponse>, MyContext>, NotUsed> retryFlow =
+    RetryBidi.create(3L, Optional.of(failureDecider), OverflowStrategy.backpressure());
+
+```
--- a/squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
+++ b/squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
@@ -1,0 +1,307 @@
+/*
+ * Copyright 2017 PayPal
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.squbs.streams
+
+import java.util.Optional
+
+import akka.NotUsed
+import akka.http.org.squbs.util.JavaConverters
+import akka.japi.Pair
+import akka.stream.Attributes.InputBuffer
+import akka.stream.scaladsl.{BidiFlow, Flow}
+import akka.stream.stage._
+import akka.stream._
+import akka.stream.OverflowStrategy._
+
+import scala.collection.mutable
+import scala.util.{Failure, Try}
+
+object RetryBidi {
+  /**
+    * Creates a [[BidiFlow]] that can be used to provide Retry functionality.
+    * This API is specifically for flows that are using [[Try]]'s for elements that may occasionally fail.  By default,
+    * any [[Failure]] is considered a failure that should be retried. However, a failureDecider function can be
+    * specified to control what constitutes a failure.
+    *
+    * Retry functionality requires each element passing through this flow to be uniquely identifiable for retrying, so
+    * it requires a [[Context]], of any type carried along with the flow's input and output element as a
+    * [[Tuple2]] (Scala) or [[Pair]] (Java).  The requirement is that either the [[Context]] type itself or a mapping
+    * from [[Context]] should be able to uniquely identify each element passing through flow.
+    *
+    * Here are the ways a unique id can be provided:
+    *
+    *   - [[Context]] itself is a type that can be used as a unique id, e.g., [[Int]], [[Long]], [[java.util.UUID]]
+    *   - [[Context]] extends [[UniqueId.Provider]] and implements [[UniqueId.Provider.uniqueId]] method
+    *   - [[Context]] is of type [[UniqueId.Envelope]]
+    *   - [[Context]] can be mapped to a unique id by calling {{{uniqueIdMapper}}}
+    *
+    * This stage supports a default in-flight maximum number of elements based on stage Attribute InputBuffer max.
+    * This maximum buffer size includes all elements currently failing (and being re-tried) as well as any elements
+    * in-flight.  To increase this size you can update the stage attribute max value for InputBuffer.
+    *
+    * @param maxRetries     the maximum number of retry attempts on any failures before giving up.
+    * @param uniqueIdMapper the function that maps [[Context]] to a unique id
+    * @param failureDecider function to determine if an element passed by the joined [[Flow]] is
+    *                       actually a failure or not
+    * @param overflowStrategy the overflowStrategy to use on Retry buffer filling
+    * @tparam In      the type of elements pulled from upstream along with the [[Context]]
+    * @tparam Out     the type of the elements that are pushed to downstream along with the [[Context]]
+    * @tparam Context the type of the context that is carried along with the elements.
+    * @return a [[BidiFlow]] with Retry functionality
+    */
+  def apply[In, Out, Context](maxRetries: Long, uniqueIdMapper: Context => Option[Any] = (_: Any) => None,
+                              failureDecider: Option[Try[Out] => Boolean] = None,
+                              overflowStrategy: OverflowStrategy = OverflowStrategy.backpressure):
+  BidiFlow[(In, Context), (In, Context), (Try[Out], Context), (Try[Out], Context), NotUsed] =
+    BidiFlow.fromGraph(new RetryBidi(maxRetries, uniqueIdMapper, failureDecider, overflowStrategy))
+
+  import scala.compat.java8.OptionConverters._
+  /**
+    * Java API
+    * Creates a [[akka.stream.javadsl.BidiFlow]] that can be joined with a [[akka.stream.javadsl.Flow]] to add
+    * Retry functionality with uniqueIdMapper, custom failure decider and OverflowStrategy.
+    */
+  def create[In, Out, Context](maxRetries: Long, uniqueIdMapper: java.util.function.Function[Context, Optional[Any]],
+                               failureDecider: Optional[java.util.function.Function[Try[Out], Boolean]],
+                               overflowStrategy: OverflowStrategy):
+  javadsl.BidiFlow[Pair[In, Context], Pair[In, Context], Pair[Try[Out], Context], Pair[Try[Out], Context], NotUsed] =
+    JavaConverters.toJava(apply[In, Out, Context](maxRetries,
+      uniqueIdMapper = UniqueId.javaUniqueIdMapperAsScala(uniqueIdMapper),
+      failureDecider = failureDecider.asScala.map(f => (out: Try[Out]) => f(out)),
+      overflowStrategy))
+
+  /**
+    * Java API
+    * @see above for details about each parameter
+    */
+  def create[In, Out, Context](maxRetries: Long,
+                               failureDecider: Optional[java.util.function.Function[Try[Out], Boolean]],
+                               overflowStrategy: OverflowStrategy):
+  javadsl.BidiFlow[Pair[In, Context], Pair[In, Context], Pair[Try[Out], Context], Pair[Try[Out], Context], NotUsed] =
+    JavaConverters.toJava(apply[In, Out, Context](maxRetries,
+      failureDecider = failureDecider.asScala.map(f => (out: Try[Out]) => f(out)),
+      overflowStrategy = overflowStrategy))
+
+  /**
+    * Java API
+    * @see above for details about each parameter.
+    */
+  def create[In, Out, Context](maxRetries: Long, uniqueIdMapper: java.util.function.Function[Context, Optional[Any]],
+                               overflowStrategy: OverflowStrategy):
+  javadsl.BidiFlow[Pair[In, Context], Pair[In, Context], Pair[Try[Out], Context], Pair[Try[Out], Context], NotUsed] =
+    JavaConverters.toJava(apply[In, Out, Context](maxRetries,
+      uniqueIdMapper = UniqueId.javaUniqueIdMapperAsScala(uniqueIdMapper),
+      overflowStrategy = overflowStrategy))
+
+  /**
+    * Java API
+    * @see above for details about each parameter.
+    */
+  def create[In, Out, Context](maxRetries: Long):
+  javadsl.BidiFlow[Pair[In, Context], Pair[In, Context], Pair[Try[Out], Context], Pair[Try[Out], Context], NotUsed] =
+    JavaConverters.toJava(apply[In, Out, Context](maxRetries))
+}
+
+/**
+  * A bidi [[GraphStage]] that can be joined with flows that produce [[Try]]'s to add Retry functionality
+  * when there are any failures.  When the joined [[Flow]] has a failure then based on the provided
+  * max retries count, it will retry the failures.
+  *
+  * '''Emits when''' a Success is available from joined flow or a failure has been retried the maximum number of retries
+  *
+  * '''Backpressures when''' the element is not a failure and downstream backpressures or the retry buffer is full
+  *
+  * '''Completes when''' upstream completes
+  *
+  * '''Cancels when''' downstream cancels
+  *
+  * {{{
+  *          upstream      +------+      downstream
+  *       (In, Context) ~> |      | ~> (In, Context)
+  *            In1         | bidi |        Out1
+  * (Try[Out], Context) <~ |      | <~ (Try[Out], Context)
+  *           Out2         +------+        In2
+  * }}}
+  *
+  * @param maxRetries maximum number of retry attempts on any failing [[Try]]'s
+  * @param uniqueIdMapper function that maps a [[Context]] to a unique value per element
+  * @param failureDecider function that gets called to determine if an element passed by the joined [[Flow]] is a
+  *                       failure
+  * @tparam In the type of elements pulled from the upstream along with the [[Context]]
+  * @tparam Out the type of the elements that are pushed by the joined [[Flow]] along with the [[Context]].
+  *             This then gets wrapped with a [[Try]] and pushed downstream with a [[Context]]
+  * @tparam Context the type of the context that is carried around along with the elements.
+  */
+final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqueIdMapper: Context => Option[Any],
+                                                         failureDecider: Option[Try[Out] => Boolean] = None,
+                                                         strategy: OverflowStrategy = OverflowStrategy.backpressure)
+  extends GraphStage[BidiShape[(In, Context), (In, Context), (Try[Out], Context), (Try[Out], Context)]] {
+
+  require(maxRetries > 0)
+
+  private val in1 = Inlet[(In, Context)]("RetryBidi.in1")
+  private val out1 = Outlet[(In, Context)]("RetryBidi.out1")
+  private val in2 = Inlet[(Try[Out], Context)]("RetryBidi.in2")
+  private val out2 = Outlet[(Try[Out], Context)]("RetryBidi.out2")
+  override val shape = BidiShape(in1, out1, in2, out2)
+
+  private[streams] def uniqueId(context: Context) =
+    uniqueIdMapper(context).getOrElse {
+      context match {
+        case uniqueIdProvider: UniqueId.Provider ⇒ uniqueIdProvider.uniqueId
+        case `context` ⇒ `context`
+      }
+    }
+
+  private[streams] val isFailure = failureDecider.getOrElse((e: Try[Out]) => e.isFailure)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape)
+    with StageLogging {
+
+    val internalBufferSize: Int =
+      inheritedAttributes.get[InputBuffer] match {
+        case None ⇒ throw new IllegalStateException(s"Couldn't find InputBuffer Attribute for $this")
+        case Some(InputBuffer(_, max)) ⇒ max
+      }
+
+    // A map of all the in flight (including failed) in elements with their retry count (bounded by size)
+    private val retryRegistry = mutable.LinkedHashMap.empty[Any, (In, Context, Long)]
+    private val readyToRetry = mutable.Queue.empty[(In, Context)] // Queue of elements ready to be emitted on out1
+    private var upstreamFinished = false
+
+    private def incrementOrRemoveFailure(context: Context): Boolean =
+      retryRegistry.get(uniqueId(context)) match {
+        case None =>
+          log.debug("Entry for context [{}] dropped", context)
+          true
+        case Some((_, ctx, retry)) if retry >= maxRetries =>
+          retryRegistry -= uniqueId(ctx)
+          log.debug("All retries exhausted for context [{}]", context)
+          true
+        case Some((in, ctx, retry)) =>
+          retryRegistry += ((uniqueId(context), (in, ctx, retry + 1)))
+          readyToRetry.enqueue((in, context))
+          log.debug("Queueing retry {} for context [{}]", retry + 1, context)
+          false
+      }
+
+    // Some useful hidden types for pattern match
+    private val backPressure = OverflowStrategy.backpressure
+    private val dropBuffer = OverflowStrategy.dropBuffer
+    private val dropHead = OverflowStrategy.dropHead
+    private val dropNew = OverflowStrategy.dropNew
+    private val dropTail = OverflowStrategy.dropTail
+    private val fail = OverflowStrategy.fail
+
+    private def handleBufferFull(): Unit = strategy match {
+      case `dropHead` =>
+        retryRegistry -= retryRegistry.head._1 // build a Buffer for squbs
+        log.debug("Buffer full dropping head")
+        grabAndPush()
+      case `dropNew` =>
+        grab(in1)
+        log.debug("Buffer full dropping newest")
+      case `dropTail` =>
+        retryRegistry -= retryRegistry.last._1
+        log.debug("Buffer full dropping last")
+        grabAndPush()
+      case `dropBuffer` =>
+        retryRegistry.clear()
+        log.debug("Buffer full dropping buffer")
+        grabAndPush()
+      case `fail` =>
+        failStage(BufferOverflowException(s"Retry buffer overflow for retry stage (max capacity was: $internalBufferSize)!"))
+      case `backPressure` =>
+        // NOP.  Don't grab any elements from upstream in backpressure mode when full
+      case _ ⇒
+        throw new IllegalStateException("Retry buffer overflow mode not supported")
+    }
+
+    private def pullCondition: Boolean = strategy != backpressure || retryRegistry.size < internalBufferSize
+    private def isBufferFull: Boolean = retryRegistry.size >= internalBufferSize
+
+    def grabAndPush(): Unit = {
+      val (elem, context) = grab(in1)
+      retryRegistry.put(uniqueId(context), (elem, context, 0))
+      push(out1, (elem, context))
+    }
+
+    setHandler(in1, new InHandler {
+      override def onPush(): Unit = {
+        if (isAvailable(out1)) {
+          if (isBufferFull) handleBufferFull()
+          else grabAndPush()
+        }
+      }
+
+      override def onUpstreamFinish(): Unit = {
+        if (retryRegistry.isEmpty) completeStage()
+        upstreamFinished = true
+      }
+
+      override def onUpstreamFailure(ex: Throwable): Unit = if (retryRegistry.isEmpty) fail(out1, ex) else failStage(ex)
+    })
+
+    setHandler(out1, new OutHandler {
+      override def onPull(): Unit = {
+        if (readyToRetry.nonEmpty) push(out1, readyToRetry.dequeue())
+        else if (isAvailable(in1)) {
+          if (isBufferFull) handleBufferFull()
+          else grabAndPush()
+        } else if (pullCondition && !upstreamFinished && !hasBeenPulled(in1)) pull(in1)
+      }
+
+      override def onDownstreamFinish(): Unit =
+        if (retryRegistry.isEmpty) {
+          completeStage()
+          log.debug("completed Out1")
+        } else cancel(in1)
+    })
+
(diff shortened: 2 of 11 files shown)


squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,211 @@
...
+    setHandler(in1, new InHandler {
+      override def onPush(): Unit = {
+        val (elem, context) = grab(in1)
+        if (isAvailable(out1)) {
+          inBuffer.put(uniqueId(context), (elem, context, 0))
+          push(out1, (elem, context))
+        } else readyToPush.enqueue((elem, context))
+      }
+
+      override def onUpstreamFinish(): Unit = if (inBuffer.isEmpty && readyToPush.isEmpty) complete(out1)
```
anilgursel: `else upstreamFinished = true`?  

Should we also consider checking `upstreamFinished` in `onPush` of `in2` or in `onPull` or `out2`?  In any one of them, if the registry is empty, that would mean there is no inflight messages and if `upstreamFinish` is `true`, then it may be ok to complete or schedule an event to complete (to let downstream to process messages in case downstream is not draining in-flight properly).

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,211 @@
...
+      override def onUpstreamFailure(ex: Throwable): Unit = fail(out1, ex)
```
anilgursel: Should we let the in-flight messages processed?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,211 @@
...
+    })
+
+    setHandler(out1, new OutHandler {
+      override def onPull(): Unit = {
+        if (readyToPush.nonEmpty) {
+          val (elem, context) = readyToPush.dequeue()
+          inBuffer.put(uniqueId(context), (elem, context, 0))
```
anilgursel: I would expect `inBuffer` already having this element.  As soon a we grab it from `in1`, we should be adding it to `inBuffer`, right?  Any benefit not doing that ?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,211 @@
...
+      inBuffer.put(uniqueId(ctx), (elem, ctx, retry + 1))
+    }
+
+    setHandler(in1, new InHandler {
+      override def onPush(): Unit = {
+        val (elem, context) = grab(in1)
+        if (isAvailable(out1)) {
+          inBuffer.put(uniqueId(context), (elem, context, 0))
+          push(out1, (elem, context))
+        } else readyToPush.enqueue((elem, context))
```
anilgursel: Shouldn't we do `inBuffer.put(uniqueId(context), (elem, context, 0))` no matter what ?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,211 @@
...
+    })
+
+    setHandler(out1, new OutHandler {
+      override def onPull(): Unit = {
+        if (readyToPush.nonEmpty) {
+          val (elem, context) = readyToPush.dequeue()
+          inBuffer.put(uniqueId(context), (elem, context, 0))
+          push(out1, (elem, context))
+        } else if (inBuffer.nonEmpty) {
+          val (elem, context, retry) = inBuffer.head._2 //TODO
```
anilgursel: The order of `inBuffer` should not matter at all.  We should always treat equally to a re-tried vs first time elements.  Otherwise, we may end up with a scenario of possible starvation.  Also, I do not think the re-tried element is less important than a first time try (of course this assume the retry delay has already passed).

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,211 @@
...
+          incrementRetry(context)
+          push(out1, (elem, context))
+        } else if (!hasBeenPulled(in1)) pull(in1)
+      }
+
+      override def onDownstreamFinish(): Unit =
+        if (readyToPush.isEmpty) {
```
anilgursel: I think we should instead check the registry (`inBuffer`).  `readyToPush` might be empty but there may still be in-flight elements in the joined flow.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,211 @@
...
+          completeStage()
+          log.debug("completed Out1")
+        } else cancel(in1)
```
anilgursel: This stops receiving any new elements from `in1`, right?   👍 

Not sure, but once all messages are drained, we may need to check if we need to complete the stage again.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,211 @@
...
+
+    setHandler(in2, new InHandler {
+      override def onPush(): Unit = {
+        val (elem, context) = grab(in2)
+        if (elem.isFailure) {// have client determine what constitutes "failure" via func
+          if (retriesExhausted(context)) push(out2, (elem, context))
+          else if (!hasBeenPulled(in2)) pull(in2)
+        } else {
+          inBuffer.remove(uniqueId(context))
+          push(out2, (elem, context))
```
anilgursel: I believe this should always be safe.  But, need to think about the scenario where downstream finishes.  Would that cause an exception.  Need to check.  There may not be much to do, but it may just allow us to handle things without abrupt failures.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,211 @@
...
+        }
+      }
+
+      override def onUpstreamFinish(): Unit = {
+        if (inBuffer.isEmpty) completeStage()
+        upstreamFinished = true
+      }
+      override def onUpstreamFailure(ex: Throwable): Unit = fail(out2, ex)
```
anilgursel: Not sure if we should immediately fail.  In-flight messages may still be successfully drained?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,211 @@
...
+    })
+
+    setHandler(out2, new OutHandler {
+      override def onPull(): Unit = {
+        if (!hasBeenPulled(in2)) pull(in2)
+     }
+
+      override def onDownstreamFinish(): Unit =
+        if (readyToPush.isEmpty && inBuffer.isEmpty) {
```
anilgursel: I think `if (readyToPush.isEmpty && inBuffer.isEmpty)` should be enough?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,211 @@
...
+          completeStage()
+          log.debug("completed out2")
```
anilgursel: When you do `completeStage`, doesn't it complete the entire stage?

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+      case (elem, ctx) => (Success(elem), ctx)
+    }
+    var context = 0L
+    val result = Source("a" :: "b" :: "c" :: Nil)
+      .map { e => context += 1; (e, context) }
+      .via(RetryBidi[String, String, Long](1).join(flow))
+      .runWith(Sink.seq)
+
+    val expected = (Success("b"), 2) :: (Success("c"), 3) :: (failure, 1) :: Nil
+    result map { _ should contain theSameElementsAs expected }
```
anilgursel: I would suggest having a retried element in the middle as well and it should not affect any following elements.  Of course, the order cannot be guaranteed in that scenario.

In addition, a test where downstream does not guarantee the order of messages would be useful (just to be on the safe side).

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  it should "return failure if last failure exhausts retries" in {
```
anilgursel: We should also assert that the element is actually tried max-retry count times,

anilgursel: Also, a scenario where the element succeeds after two retries.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new TimerGraphStageLogic(shape) with StageLogging {
+
+    // A map of the in flight (possibly failed) in elements with their retry count
+    private val inBuffer = mutable.HashMap.empty[Any, (In, Context, Long)]
+    private val readyToPush = mutable.Queue.empty[(In, Context)] // Element ready to be emitted on out1
+    private var upstreamFinished = false
+
+    private def retriesExhausted(context: Context): Boolean = {
```
thefourtheye: This doesn't sound like a function which has side-effects

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,211 @@
...
+      context match {
+        case uniqueIdProvider: UniqueId.Provider ⇒ uniqueIdProvider.uniqueId
+        case context ⇒ context
+      }
+    }
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new TimerGraphStageLogic(shape) with StageLogging {
+
+    // A map of the in flight (possibly failed) in elements with their retry count
+    private val inBuffer = mutable.HashMap.empty[Any, (In, Context, Long)]
```
sebady: I'll use `retryRegistry`

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
sebady: The element is added into inBuffer only after the initial push on out1.  If it's queue for push on out1 then it would be in readyToPush.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
sebady: The benefit to not putting it directly into inBuffer when initially grabbed is to easily distinguish elements that were able to be initially pushed on `out1` from those that could not be pushed (because out1 was not available).

anilgursel: For what purpose do we use that? If for stats, then we can infer that by looking into `readyToPush` and `inBuffer`.

sebady: How would out1 onPull determine which elements have never been pushed on out1?  Without causing starvation due to too many retries in inBuffer?

anilgursel: Why does it need to determine which elements have never been pushed?  It just looks at `readyToPush` to determine what's next.

> The current onPull favors elements that never have been tried ahead of elements that have been failing.

I disagree.  It should not favor.  What's the reason for favoring?  If there are too many re-tries, you will most likely be back pressuring (which is expected).

At the end `readyToPush` it a queue, whatever gets first in the queue, goes out first whether it is the first time or a retry.

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
```diff
@@ -0,0 +1,206 @@
...
+import akka.NotUsed
+import akka.actor.{ActorSystem, Props}
+import akka.stream.{ActorMaterializer, OverflowStrategy}
+import akka.stream.scaladsl.{BidiFlow, Flow, Sink, Source}
+import akka.testkit.TestKit
+import akka.util.Timeout
+import org.scalatest.{AsyncFlatSpecLike, Matchers}
+import org.squbs.streams.UniqueId.Provider
+
+import scala.concurrent.duration._
```
akara: `import scala.language.postfixOps` to alleviate compiler warnings around postfix operations.

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
```diff
@@ -0,0 +1,206 @@
...
+      r should contain theSameElementsAs (failure, 42)::Nil
+      count.get shouldEqual maxRetry + 1
+    }
+    //    succeed
+  }
+
+  ignore should "return Success when a failure is retried successfully" in {
+    var first = true
+    val flow = Flow[(String, Long)].map {
+      case ("y", ctx) if (first == true) => {
```
akara: `case ("y", ctx) if first =>`

Also, the curly brace at the end of line is not needed following the `=>`

IntelliJ should have given you this.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,243 @@
...
+  def create[In, Out, Context](maxRetries: Long, failureDecider: Optional[Function[Try[Out], Boolean]],
+                               uniqueIdMapper: java.util.function.Function[Context, Optional[Any]]):
+  akka.stream.javadsl.BidiFlow[Pair[In, Context], Pair[In, Context], Pair[Try[Out], Context], Pair[Try[Out], Context], NotUsed] =
+    JavaConverters.toJava(apply(maxRetries, failureDeciderOptionAsScala[Out](failureDecider), UniqueId.javaUniqueIdMapperAsScala(uniqueIdMapper)))
+
+  import scala.compat.java8.FunctionConverters._
+  import scala.compat.java8.OptionConverters._
+
+  private def failureDeciderAsScala[Out](failureDecider: Function[Try[Out], Boolean]): Try[Out] => Boolean =
+    (out: Try[Out]) => failureDecider.asScala.apply(out)
```
anilgursel: We should actually be able to simplifyg this once 2.12 PR is merged.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,243 @@
...
+      }
+    }
+
+    setHandler(in1, new InHandler {
+      override def onPush(): Unit = {
+        val (elem, context) = grab(in1)
+        retryRegistry.put(uniqueId(context), (elem, context, 0))
+        if (isAvailable(out1)) {
+          push(out1, (elem, context))
+        } else readyToPush.enqueue((elem, context))
```
anilgursel: I agree with you.  We can grab the element only if`out1` is available, and in `onPull` of `out1`, we can try to grab it if element is already available.

But, the queue will no grow indefinitely though.  `pull(in1)` is only done if there is downstream demand and `readyToPush` is empty.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,243 @@
...
+        upstreamFinished = true
+      }
+
+      override def onUpstreamFailure(ex: Throwable): Unit = if (retryRegistry.isEmpty) fail(out1, ex) else failStage(ex)
+    })
+
+    setHandler(out1, new OutHandler {
+      override def onPull(): Unit = {
+        if (readyToPush.nonEmpty) {
+          push(out1, readyToPush.dequeue())
```
anilgursel: There are two sources for `out1`: `in1` and retried elements from `in2`.  There is no point of delaying `grab(in2)` until `out1` is available as the element may actually be a success or retries may have maxed out (in these cases we send the element downstream).  But, once we get an element that should be retried and if `out1` is not available, then we need to put it in a queue, don't we?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,243 @@
...
+          else pull(in2)
+        } else {
+          retryRegistry.remove(uniqueId(context))
+          push(out2, (elem, context))
+        }
+      }
+
+      override def onUpstreamFinish(): Unit = {
+        if (retryRegistry.isEmpty) complete(out2)
+        upstreamFinished = true
```
anilgursel: Shouldn't this be `if (retryRegistry.isEmpty) completeStage()`?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
sebady: So it sounds preferable to not grab the element unless out1 is available?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
akara: I'm OK with buffering some retries, which should not be a problem in the normal situation. The question is how much before we stop grabbing from `in2`. If the retried part of the stream can hold very many elements (say there is a `PersistentBuffer` in there), and you have a huge amount of elements requiring a retry, you may run into the situation your buffer just blows up. So there is a choice between a small, limited buffer (and how to configure the limit), and no buffer at all (which I prefer in case there is a buffer in the retry - as the chance of back-pressure on `out1` would be small.

If not having a buffer there does not cause a deadlock, I still prefer that.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
akara: Yes you're right, we should not complete `out2` separate from `completeStage()`. But is there any thought behind keeping it separate?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
anilgursel: Yes.  In `onPull` of `out1`, we should first check the buffer, if empty than you can `grab` if an element is available.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
anilgursel: Actually, I do not think we should override `onUpstreamFinish` for `in2` (we should do it only for `in1`).  If it is finished, we have no option but to complete the stage.  Delaying the completion and not doing anything would just hang the stream and prevent it from completing (hopefully this scenario should never happen though).

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
anilgursel: If there is a `PersistentBuffer` kind of component in the joined flow, then there will be a continuous demand from downstream (`out1`), so `readyToPush` should not be growing.

Just to get more clarity, your suggestion is not to do `grab(in2)` unless `out1` is available?  If so, this may potentially cause a deadlock.  We may even have to `pull(in2)`.  But, given that `grab(in1)` will be done only if `out1` is available, we may be able to set the size of `readyToPush` to 1?

squbs-pattern/src/test/scala/org/squbs/pattern/stream/BroadcastBufferAtLeastOnceSpec.scala
```diff
@@ -45,8 +45,11 @@ abstract class BroadcastBufferAtLeastOnceSpec[T: ClassTag, Q <: QueueSerializer[
 
   val transform = Flow[Int] map createElement
 
+  // awaitMax value is too low on slow systems so increase timeout here.
+  val timeout = awaitMax * 8
```
sebady: I will change the awaitMax value in Timeouts.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+    private val Backpressure = OverflowStrategy.backpressure
+    private val DropBuffer = OverflowStrategy.dropBuffer
+    private val DropHead = OverflowStrategy.dropHead
+    private val DropNew = OverflowStrategy.dropNew
+    private val DropTail = OverflowStrategy.dropTail
+    private val Fail = OverflowStrategy.fail
+
+    private def onBufferFull: Unit = strategy match {
+      case DropHead ⇒
+        retryRegistry.remove(retryRegistry.head._1)  // build a Buffer for squbs
```
akara: You can also use the `-=` operator if you wish to.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+
+    // Some useful hidden types for pattern match
+    private val Backpressure = OverflowStrategy.backpressure
+    private val DropBuffer = OverflowStrategy.dropBuffer
+    private val DropHead = OverflowStrategy.dropHead
+    private val DropNew = OverflowStrategy.dropNew
+    private val DropTail = OverflowStrategy.dropTail
+    private val Fail = OverflowStrategy.fail
+
+    private def onBufferFull: Unit = strategy match {
```
akara: Since `onBufferFull` possibly mutates the state of the buffer, we should follow convention to provide a `()` in the sigrature.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+    def grabAndPush(): Unit = {
+      val (elem, context) = grab(in1)
+      retryRegistry.put(uniqueId(context), (elem, context, 0))
+      push(out1, (elem, context))
+    }
+
+    setHandler(in1, new InHandler {
+      override def onPush(): Unit = {
+        if (isAvailable(out1)) {
+          if (isBufferFull) onBufferFull
```
akara: So here we should call `onBufferFull()`. We can debate the naming and this is not a requirement, but perhaps we should call it `handleBufferFull()` instead.

squbs-ext/src/test/java/org/squbs/streams/RetryBidiTest.java
```diff
@@ -0,0 +1,122 @@
...
+                Source.from(Arrays.asList("a", "b", "c"))
+                        .map(s -> new Pair<>(s, UUID.randomUUID()))
+                        .via(retry.join(flow))
+                        .map(Pair::first)
+                        .runWith(Sink.seq(), mat);
+
+        final List<Try<String>> expected = Arrays.asList(Success.apply("a"), Success.apply("b"), Success.apply("c"));
+        List<Try<String>> actual = result.toCompletableFuture().get();
+        assertEquals(actual.size(), 3);
+        assertTrue(actual.containsAll(expected));
```
akara: These `assertTrue(...)` calls are hard to trace when things fail. You may want to add an explanatory string in the `assertTrue`.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+      }
+    }
+
+    // Some useful hidden types for pattern match
+    private val Backpressure = OverflowStrategy.backpressure
+    private val DropBuffer = OverflowStrategy.dropBuffer
+    private val DropHead = OverflowStrategy.dropHead
+    private val DropNew = OverflowStrategy.dropNew
+    private val DropTail = OverflowStrategy.dropTail
+    private val Fail = OverflowStrategy.fail
```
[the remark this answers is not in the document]
sebady: We need the case class here (`DropHead`, `DropNew` etc) for the pattern match below.  These are ```private[akka]``` though.
  
 `dropHead`,  `dropNew` etc values didn't work. Unless I'm not understand your comment here. I initially tried to use
```
private def handleBufferFull(): Unit = strategy match {
      case dropHead =>
```
but encountered  `patterns after a variable pattern cannot match (SLS 8.1.1)`

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+          retryRegistry.remove(uniqueId(context))
+          log.debug("All retries exhausted for context [{}]", context)
+          true
+        } else {
+          retryRegistry.put(uniqueId(context), (in, ctx, retry + 1))
+          readyToPush.enqueue((in, context))
+          log.debug("Queueing retry {} for context [{}]", retry + 1, context)
+          false
+        }
+      }
```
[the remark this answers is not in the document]
sebady: Thanks. I'll make this update as well as the other code style recommendations here.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+import scala.util.{Failure, Try}
+
+object RetryBidi {
+  /**
+    * Creates a [[BidiFlow]] that can be used to provide Retry functionality.
+    * This API is specifically for flows that are using [[Try]]'s for elements that may occasionally fail.  By default, any
+    * [[Failure]] is considered a failure that should be retried. However a failureDecider function can be specified to control
+    * what constitutes a failure.
+    *
+    * Retry functionality requires each element passing through this flow to be uniquely identifiable for retrying, so
```
anilgursel: > ... uniquely identifiable for retrying ...

I think we may have to talk about registry little bit instead of just saying "retrying".

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+    *   - [[Context]] itself is a type that can be used as a unique id, e.g., [[Int]], [[Long]], [[java.util.UUID]]
+    *   - [[Context]] extends [[UniqueId.Provider]] and implements [[UniqueId.Provider.uniqueId]] method
+    *   - [[Context]] is of type [[UniqueId.Envelope]]
+    *   - [[Context]] can be mapped to a unique id by calling {{{uniqueIdMapper}}}
+    *
+    * This stage supports a default in-flight maximum number of elements based on stage Attribute InputBuffer max.
+    * This maximum buffer size includes all elements currently failing (and being re-tried) as well as any elements in-flight.
+    * To increase this size you can update the stage attribute max value for InputBuffer.
+    *
+    * @param maxRetries     the maximum number of retry attempts on any failuures before giving up.
```
anilgursel: typo on "failuures".

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+
+  /**
+    * Java API
+    * @see above for details about each parameter.
+    */
+  def create[In, Out, Context](maxRetries: Long):
+  akka.stream.javadsl.BidiFlow[Pair[In, Context], Pair[In, Context], Pair[Try[Out], Context], Pair[Try[Out], Context], NotUsed] =
+    JavaConverters.toJava(apply[In, Out, Context](maxRetries))
+
+  import scala.compat.java8.FunctionConverters._
```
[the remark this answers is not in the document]
anilgursel: `master` is on Scala 2.12 now, you should not need it.  Please see https://github.com/paypal/squbs/pull/501 on examples of how to eliminate some of the converters.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+                               overflowStrategy: OverflowStrategy):
+  akka.stream.javadsl.BidiFlow[Pair[In, Context], Pair[In, Context], Pair[Try[Out], Context], Pair[Try[Out], Context], NotUsed] =
+    JavaConverters.toJava(apply[In, Out, Context](maxRetries, failureDecider = failureDeciderOptionAsScala(failureDecider),
+      overflowStrategy = overflowStrategy))
+
+  /**
+    * Java API
+    * @see above for details about each parameter.
+    */
+  def create[In, Out, Context](maxRetries: Long, uniqueIdMapper: java.util.function.Function[Context, Optional[Any]],
```
anilgursel: with too many settings, should we consider having the API as `withMaxRetries`, `withUniqueIdMapper`, etc ?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+  private def failureDeciderAsScala[Out](failureDecider: java.util.function.Function[Try[Out], Boolean]): Try[Out] => Boolean =
+    (out: Try[Out]) => failureDecider.asScala.apply(out)
+
+  private def failureDeciderOptionAsScala[Out](failureDecider: Optional[java.util.function.Function[Try[Out], Boolean]]) =
+    failureDecider.asScala.map(failureDeciderAsScala)
+}
+
+/**
+  * A bidi [[GraphStage]] that can be joined with flows that produce [[Try]]'s to add Retry functionality
+  * when any [[Try]] is a `Failure`.  When the joined [[Flow]] has a failure then based on the provided
```
anilgursel: > ... when any [[Try]] is a `Failure`

This is not necessarily true.  Some `Success`es can be retried as well.  All depends on failure decider.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+  * max retries count (and delay) setting, it will retry the failed [[Try]]'s.
```
anilgursel: Instead of "delay", should we instead say "backoff" ?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+  *
+  * '''Emits when''' a Success is available from joined flow or Failure has already been retried max retry times already
```
anilgursel: I think the second part of this sentence is not complete.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+  *
+  * '''Backpressures when''' the element is not a Failure and downstream backpressures
```
anilgursel: Backpressures when registry is full, isn't it ?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) with StageLogging {
+    val size: Int =
+      inheritedAttributes.get[InputBuffer] match {
+        case None                        ⇒ throw new IllegalStateException(s"Couldn't find InputBuffer Attribute for $this")
+        case Some(InputBuffer(_, max)) ⇒ max
+      }
+
+    // A map of all the in flight (including failed) in elements with their retry count (bounded by size)
+    private val retryRegistry = mutable.LinkedHashMap.empty[Any, (In, Context, Long)]
```
anilgursel: Because of the overflow strategies (drop head, drop tail, etc), we may have to keep a "Linked" one; otherwise, I think it would not be the most efficient one.  I'd rather use Java map's merge for look and update in one shot.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+    private val readyToPush = mutable.Queue.empty[(In, Context)] // Queue of elements ready to be emitted on out1
+    private var upstreamFinished = false
+
+    private def incrementOrRemoveFailure(context: Context): Boolean = {
+      val element = retryRegistry.get(uniqueId(context))
+      if (element.isEmpty) {
```
anilgursel: I think it is more readable with a `match` and `case Some(elem)` instead of `isEmpty` check and `element.get`.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+        log.debug("Entry for context [{}] dropped", context)
+        true
```
anilgursel: It returns a boolean; however, the function name does not reflect any boolean convention.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+      case DropBuffer ⇒
+        retryRegistry.clear()
+        log.debug("Buffer full dropping buffer")
+        grabAndPush()
+      case Fail ⇒
+        failStage(BufferOverflowException(s"Retry buffer overflow for retry stage (max capacity was: $size)!"))
+      case Backpressure ⇒
+        throw new IllegalStateException("Retry buffer must never overflow in backpressure mode")
+      case _ ⇒
+        throw new IllegalStateException("Retry buffer overflow mode not supported")
```
anilgursel: Should the log statement reflect the name of the strategy being tried to be used ?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    private def pullCondition: Boolean = strategy != Backpressure || retryRegistry.size < size
```
anilgursel: Rename `size` as `internalBufferSize` ?  Just for better readability ?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
anilgursel: When the overflow strategy is backpressure, this would throw an exception.  Shouldn't we instead just not grab the element, so that it back pressures?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+      }
+
+      override def onUpstreamFailure(ex: Throwable): Unit = if (retryRegistry.isEmpty) fail(out1, ex) else failStage(ex)
+    })
+
+    setHandler(out1, new OutHandler {
+      override def onPull(): Unit = {
+        if (readyToPush.nonEmpty) push(out1, readyToPush.dequeue())
+        else if (isAvailable(in1)) {
+          if (isBufferFull) onBufferFull // TODO can this throw in backpressure mode
```
anilgursel: Same comment as above.  This would throw an exception in back pressure mode, instead of actually back pressuring.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+          else grabAndPush()
+        } else if (pullCondition) pull(in1)
```
anilgursel: Should we also check `!hasBeenPulled(in1)`?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+      }
+
+      override def onDownstreamFinish(): Unit =
+        if (retryRegistry.isEmpty) {
```
anilgursel: If downstream finishes, no matter registry is full or not, we should finish, IMO.  Because, there is no where we can send the elements in registry down to.  Is it just to wait for inflight elements reach out to in2 ?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+          completeStage()
+          log.debug("completed Out1")
+        } else cancel(in1)
+    })
+
+    setHandler(in2, new InHandler {
+      override def onPush(): Unit = {
+        val (elem, context) = grab(in2)
+        if (isFailure(elem)) {
+          if (incrementOrRemoveFailure(context)) push(out2, (elem, context))
```
anilgursel: I think the function name makes the logic confusing.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+      override def onUpstreamFailure(ex: Throwable): Unit = if (readyToPush.isEmpty) fail(out2, ex)
+    })
+
+    setHandler(out2, new OutHandler {
+      override def onPull(): Unit =
+        if (retryRegistry.isEmpty && upstreamFinished) completeStage()
+        else if (!hasBeenPulled(in2)) pull(in2)
+
+      override def onDownstreamFinish(): Unit =
+        if (retryRegistry.isEmpty) {
```
anilgursel: If downstream finishes, no matter what we need to complete the stage as there is no downstream to send the messages down to anymore.

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
```diff
@@ -0,0 +1,319 @@
...
+
+    val retryBidi = RetryBidi[String, String, Long](1)
+
+    var context = 0L
+    val result = Source("d" :: "e" :: "f" :: Nil)
+      .map { s => context += 1; (s, context) }
+      .via(retryBidi.join(flow))
+      .runWith(Sink.seq)
+
+    val expected = (Success("d"), 1) :: (Success("e"), 2) :: (failure, 3) :: Nil
```
anilgursel: I'd try some of these tests with larger number of messages.  Currently, you are below the internal buffer size I believe.  Maybe you already have such scenario below, haven't gone through all yet.

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
```diff
@@ -0,0 +1,319 @@
...
+    result map {
+      _ should contain theSameElementsAs expected
+    }
+  }
+
+  it should "return exhausted failures in expected FIFO order" in {
```
anilgursel: Why FIFO order ?  I would think there is no guarantee in ordering at all.  Of course, if the joined flow keeps the ordering that's a different story.

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
```diff
@@ -0,0 +1,319 @@
...
+      .map { s => context += 1; (s, context) }
+      .via(retry.join(bottom))
+      .toMat(TestSink.probe)(Keep.both).run()
+
+    source.sendNext("a").sendNext("b")
+    val nextA = sink.request(1).requestNext()
+    assert((Success("a"), 1) == nextA)
+  }
+
+  ignore should "drop tail elements when dropTail buffer mode" in {
```
anilgursel: ignore ?

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
```diff
@@ -0,0 +1,319 @@
...
+      .via(retry.join(bottom))
+      .toMat(TestSink.probe)(Keep.both).run()
+
+    source.sendNext("a")
+    source.sendNext("b")
+    val next = sink.requestNext()
+    succeed
+  }
+
+  ignore should "fail when upstream fails" in {
```
anilgursel: ignore?

squbs-testkit/src/main/scala/org/squbs/testkit/Timeouts.scala
```diff
@@ -24,5 +24,5 @@ object Timeouts {
 
   implicit val askTimeout = Timeout(30 seconds)
 
-  val awaitMax = 60 seconds
+  val awaitMax = 240 seconds
```
anilgursel: Why is this needed?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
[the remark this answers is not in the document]
akara: Sorry, my bad. I forgot to tell you to back-quote. Try:

```scala
private def handleBufferFull(): Unit = strategy match {
  case `dropHead` =>
```

Reason is, without the quote it will try to treat `dropHead` as a local variable for this case, which conflicts with the imported one. To tell the pattern matcher you meant it should be the value of the outer variable, you back-quote it. Read more here: https://alvinalexander.com/scala/scala-unreachable-code-due-to-variable-pattern-message

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -0,0 +1,309 @@
...
+import akka.stream._
+
+import scala.collection.mutable
+import scala.util.{Failure, Try}
+
+object RetryBidi {
+  /**
+    * Creates a [[BidiFlow]] that can be used to provide Retry functionality.
+    * This API is specifically for flows that are using [[Try]]'s for elements that may occasionally fail.  By default, any
+    * [[Failure]] is considered a failure that should be retried. However a failureDecider function can be specified to control
```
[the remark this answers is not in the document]
sebady: hmm. I thought our code style was to limit at 160?  Are we using 120 as the limit?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
sebady: Are we doing this with other squbs stages?  Some API consistency between stage implementation would be nice.  If there are too many API settings here I could revisit the RetryFlowSettings case class (see https://github.com/paypal/squbs/pull/486/commits/8f96801e127a86272dd583b543a3e8f3d9771d45)

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
sebady: I'm going to remove the reference to `delay setting` for now.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
sebady: Yes, I am changing this.

docs/flow-retry.md
```diff
@@ -0,0 +1,133 @@
+# Retry Flow
```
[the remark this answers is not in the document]
sebady: I will fix this and the TImeout as well.

docs/flow-retry.md
```diff
@@ -0,0 +1,133 @@
...
+
+If all the retried attempts fail then the last `Failure` for that request is emitted.  The retry stage requires a `Context` to be carried around for each
+ element.  This is needed to uniquely identify each element.
+
+(see [Context to Unique Id Mapping](#context-to-unique-id-mapping) section for more details)
+
+##### Scala
+
+```scala
+val retryBidi = RetryBidi[String, Long](maxRetries = 2)
```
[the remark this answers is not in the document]
sebady: I favor `RetryBidi` to just `Retry`.  That's more precise in my opinion than just `Retry`. This namine is also consistent with timeout and circuit breaker.  I think some consistency is good here.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
sebady: Isn't this a bit of an implementation detail?  I've tries to not expose the registry as part of the API (cause I hope to change this in the future). I will elaborate here though on why `uniquely identifiable` is a requirement.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
sebady: not sure how I missed this?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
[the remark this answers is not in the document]
anilgursel: Yes, the limit is 120.  That's the default one on IntelliJ as well.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
anilgursel: You are right.  No need to talk about the registry; however, should explain why it needs to be uniquely identified.  Basically, what we get from joined flow is a message (success or failure), to retry we need the original message.  I think this needs to be talked.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
anilgursel: Yes, consistency is very important.  We already do that in CircuitBreakerSettings.

docs/flow-retry.md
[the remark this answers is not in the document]
anilgursel: For this PR, we should probably go for `RetryBidi`; however, I feel like without `Bidi` is a better naming for all the stages.

anilgursel:
Can you please add an entry in TOC at https://github.com/paypal/squbs/blob/master/mkdocs.yml#L31

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
anilgursel: Comments ?

Issue #487, opened 2017-08-03
Drain response entity when a timeout happens


anilgursel:
When a timeout happens, the `TimeoutBidi` stage sends a `Failure` to downstream and drops the actual response when received (after timeout period).  However, it should also allow a function to be passed for what to do when the response is received after the timeout.  For instance, for HTTP use cases, the response entity needs to be drained.  

Thanks @talst for reporting it https://gitter.im/paypal/squbs?at=597c5fc5329651f46ed1dc5a.

The timeout stage can be found at https://squbs.readthedocs.io/en/latest/flow-timeout/

talst:
I've created a suggestion to add this functionality:  https://github.com/paypal/squbs/pull/488

PR #488, opened 2017-08-04
#487 Add clean up callback for TimeOutBidi

talst:


squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -331,18 +347,28 @@ object TimeoutBidiFlowOrdered {
...
-  def create[In, Out](timeout: FiniteDuration): akka.stream.javadsl.BidiFlow[In, In, Out, Try[Out], NotUsed] =
-    apply(timeout).asJava
+  def create[In, Out](timeout: FiniteDuration,
+                      cleanUp: Optional[java.util.function.Function[Out, Void]]):
+  akka.stream.javadsl.BidiFlow[In, In, Out, Try[Out], NotUsed] = {
+    val cleanUpAsScala = scala.compat.java8.OptionConverters.toScala(cleanUp).map(_.asScala.andThen(_ => ()))
+    apply(timeout, cleanUpAsScala).asJava
+  }
+
+  def create[In, Out](timeout: FiniteDuration):
```
anilgursel: Ditto.. Java API comment.

squbs-ext/src/test/scala/org/squbs/streams/TimeoutBidiFlowSpec.scala
```diff
@@ -97,13 +101,19 @@ class TimeoutBidiFlowSpec extends TestKit(ActorSystem("TimeoutBidiFlowSpec")) wi
       FlowShape(partition.in, merge.out)
     })
 
-    val timeoutFlow = TimeoutBidiFlowOrdered[String, String](timeout, cleanUp = Some(cleanUpFunction))
+    val timeoutFlow = TimeoutBidiFlowOrdered[String, String](timeout, cleanUp = cleanUpFunction)
 
-    val result = Source("a" :: "b" :: "c" :: Nil).via(timeoutFlow.join(flow)).runWith(Sink.seq)
+    val result = Source("a" :: "b" :: "c" :: Nil)
+      .map{ s =>
+        system.scheduler.scheduleOnce(3.1 seconds)(notCleanedUpFunction(s))
```
anilgursel: Travis sometimes do not provide enough resources and timings go crazy and we end up re-structuring the timeouts.  Accordingly, I'd suggest specifying the timeout relative to `timeout` instead of hardcoding as `3.1 seconds`.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -190,23 +179,56 @@ object TimeoutBidiFlowUnordered {
...
   BidiFlow[(In, Context), (In, Context), (Out, Context), (Try[Out], Context), NotUsed] =
-    BidiFlow.fromGraph(TimeoutBidiUnordered(timeout, uniqueIdMapper))
+    BidiFlow.fromGraph(TimeoutBidiUnordered(timeout, uniqueIdMapper, cleanUp))
+
+  /**
+    * Java API
+    */
+  def create[In, Out, Context](timeout: FiniteDuration,
+                               uniqueIdMapper: java.util.function.Function[Context, Optional[Any]],
+                               cleanUp: java.util.function.Function[Out, Void]):
```
akara: Don't we rather want a `java.util.function.Consumer[Out]`?

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -190,23 +179,56 @@ object TimeoutBidiFlowUnordered {
...
+  akka.stream.javadsl.BidiFlow[Pair[In, Context], Pair[In, Context], Pair[Out, Context], Pair[Try[Out], Context], NotUsed] = {
+
+    val cleanUpAsScala = cleanUp.asScala.andThen(_ => ())
```
akara: Then we can specify cleanUpAsScala as follows:

```scala
def cleanUpAsScala(out: Out) = cleanUp.accept(out)
```

This can then be passed in the same way. Do `cleanUpAsScala _` if the compiler cannot recognize this as a function. But unlikely. This case is too simple.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -190,23 +179,56 @@ object TimeoutBidiFlowUnordered {
...
                                uniqueIdMapper: java.util.function.Function[Context, Optional[Any]]):
   akka.stream.javadsl.BidiFlow[Pair[In, Context], Pair[In, Context], Pair[Out, Context], Pair[Try[Out], Context], NotUsed] = {
+    toJava(apply[In, Out, Context](timeout, UniqueId.javaUniqueIdMapperAsScala(uniqueIdMapper)))
+  }
 
-    toJava(apply(timeout, UniqueId.javaUniqueIdMapperAsScala(uniqueIdMapper)))
+  /**
+    * Java API
+    */
+  def create[In, Out, Context](cleanUp: java.util.function.Function[Out, Void],
```
akara: Ditto for this function.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -331,18 +359,29 @@ object TimeoutBidiFlowOrdered {
...
 
   /**
     * Java API
     */
-  def create[In, Out](timeout: FiniteDuration): akka.stream.javadsl.BidiFlow[In, In, Out, Try[Out], NotUsed] =
-    apply(timeout).asJava
+  def create[In, Out](timeout: FiniteDuration,
+                      cleanUp: java.util.function.Function[Out, Void]):
+  akka.stream.javadsl.BidiFlow[In, In, Out, Try[Out], NotUsed] = {
+    val cleanUpAsScala = cleanUp.asScala.andThen(_ => ())
```
akara: Ditto

squbs-ext/src/main/scala/org/squbs/streams/circuitbreaker/japi/CircuitBreakerSettings.scala
```diff
@@ -31,6 +31,9 @@ import scala.util.Try
 case class CircuitBreakerSettings[In, Out, Context] private[japi] (
   circuitBreakerState: CircuitBreakerState,
   fallback: Optional[Function[In, Try[Out]]] = Optional.empty[Function[In, Try[Out]]],
+  cleanUp: Function[Out, Void] = new  Function[Out, Void] {
+    override def apply(t: Out): Void = null
```
akara: If I understand correctly, this is `java.util.function.Function`, right? If so, using `Consumer` would be better.

```scala
cleanUp: Consumer[Out] = new Consumer[Out] {
  override def accept(t:Out): Unit = {}
}
```

akara: You'll avoid using `null`, too! 😄

squbs-ext/src/main/scala/org/squbs/streams/circuitbreaker/japi/CircuitBreakerSettings.scala
```diff
@@ -62,6 +69,9 @@ case class CircuitBreakerSettings[In, Out, Context] private[japi] (
   private[japi] def fallbackOptionAsScala(fallback: Optional[Function[In, Try[Out]]]) =
     fallback.asScala.map(fallbackAsScala)
 
+  private[japi] def cleanUpAsScala(cleanUp: Function[Out, Void]) =
+    cleanUp.asScala.andThen(_ => ())
```
akara: Same issue here.

docs/flow-timeout.md
```diff
@@ -194,4 +194,15 @@ Source.from(Arrays.asList("a", "b", "c"))
...
+
+##### Clean up callback
+
+The `TimeoutBidi` also provide a callback clean up function to be passed. This function will be called for emitted elements were already considered timed out.
+
+Example for a use case for this functionality is Akka Http client: As described in [Implications of the streaming nature of Request/Response Entities](http://doc.akka.io/docs/akka-http/current/scala/http/implications-of-streaming-http-entity.html), all http responses must be consumed or discarded. By passing a clean up callback to discard the timed out requests when they complete we avoid clogging down the stream.
+```scala
+val akkaHttpDiscard = (response: HttpResponse) => response.discardEntityBytes()
+
+TimeoutBidiFlowOrdered[HttpRequest, HttpResponse](20 milliseconds, cleanUp = akkaHttpDiscard)
```
anilgursel: For HTTP client use cases,`TimeoutBidi` will mainly be used as part of `CircuitBreaker`.  So, if we would like to make it `implicit`, we would do it in `CircuitBreaker`, IMO.  But, then, we need to do the same for `failureDecider`, etc.  It may cause more API changes.

akara: Isn't `CircuitBreaker` separate from `HttpClient`? If so, it won't make sense to have `HttpClient` cleanup function in `CircuitBreaker`, right?

anilgursel: Not sure if I understood your comment.

What I am suggesting is that if we make `cleanup` `implicit`, then we may have to do it for some other parameters as well, e.g., `failureDecider`.

squbs-ext/src/main/scala/org/squbs/streams/circuitbreaker/japi/CircuitBreakerSettings.scala
talst: Good point! It's been a while since I've written anything in Java and I never actually used Java 8

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -194,29 +196,39 @@ object TimeoutBidiFlowUnordered {
...
+    toJava(apply[In, Out, Context](timeout, UniqueId.javaUniqueIdMapperAsScala(uniqueIdMapper)))
   }
 
-  def create[In, Out, Context](timeout: FiniteDuration,
-                               uniqueIdMapper: java.util.function.Function[Context, Optional[Any]],
-                               cleanUp: java.util.function.Function[Out, Void]):
+  /**
+    * Java API
+    */
+  def create[In, Out, Context](cleanUp: java.util.function.Function[Out, Void],
```
talst: `uniqueIdMapper: java.util.function.Function[Context, Optional[Any]])` and `cleanUp: java.util.function.Function[Out, Void])` have the same type after erasure.
I changed the order to differentiate the two functions.
But now I'm gonna use `Consumer` which should solve this problem

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -465,3 +478,12 @@ final class TimeoutBidiOrdered[In, Out](timeout: FiniteDuration, cleanUp: Option
   * @param msg Defaults to "Flow timed out!"
   */
 case class FlowTimeoutException(msg: String = "Flow timed out!") extends TimeoutException(msg)
+
+object TimeoutBidi {
+  private[streams] def tryCleanUp[Out](elem: Out, cleanUp: Out => Unit) = {
+    Try(cleanUp(elem)).recover {
+      case NonFatal(_) => ()
+      case ex => throw ex
+    }.foreach(_ => ())
```
talst: I'm used to run with `"-Ywarn-value-discard"` compiler flag XD

docs/flow-timeout.md
akara: What I meant is HttpClient cleanup function needs to be provided as part of the `httpClient` package. The cleanup call type (and implicit) need to be provided as part of this package. These are just separate concerns. And yes, we may want to consider it for `failureDecider`, too, but just for the Scala API (obviously).

anilgursel: Yes, separate concerns.  The question is whether we want to make it `implicit` or not.  If we decide to, we need to make it for `failureDecider` as well, potentially for others.  

I believe my earlier comment confused.  I am not suggesting to provide http clean up function in `CircuitBreaker`, I meant we need to make the parameter `implicit` at `CircuitBreaker` level.

talst:
- `cleanUp` parameter on the Java API changed to be a `Consumer[Out]` instead of `Function[Out, Void]`.
- Fixed the readme.
- Add new tests for the Java API and `CircuitBreakerBidiFlow`

On the point of `implicit` I didn't make any changes, TBH I'm not sure `implicit` is the right way to go for the `cleanUp` as it can lead to confusing behavior.
If you do deiced to go with `implicit` I think it's better to wrap the `cleanUp` function in a case class because `implict` `Out => Unit` is too general type

anilgursel:
Regarding, `cleanUp` being `implicit` we can consider it as a follow up issue.  In any case, for 0.9.2, we will not be able change the API.  If we decide to do so, we can make the change only in 0.10.0.

talst:
Squashed and formatted readme

anilgursel:
Are the failures related to your changes or regarding `PersistentBuffer` or `BroadcastBuffer`?  If the latter, then we just need to re-try (Travis gives very little resource which makes timing a problem).  We will try to address those though.   If related to your changes, need to look in what's going on.

talst:
Ok looks like it passed now, maybe he was pissed I didn't squash 😉

PR #491, opened 2017-08-10
#489 Upgrade to Akka 2.4.18 and Akka HTTP 10.0.9

anilgursel:

- [x] Commits are squashed.
- [ ] Tests added.

### DIFF
--- a/squbs-ext/src/main/scala/akka/http/org/squbs/util/JavaConverters.scala
+++ b/squbs-ext/src/main/scala/akka/http/org/squbs/util/JavaConverters.scala
@@ -57,14 +57,20 @@ object JavaConverters {
     javaToScalaAdapter.atop(bidiFlow).atop(scalaToJavaAdapter).asJava
   }
 
-  private def adaptTupleFlow[T, Mat](scalaFlow: Flow[(HttpRequest, T), (Try[HttpResponse], T), Mat]):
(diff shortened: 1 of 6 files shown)


squbs-ext/src/main/scala/akka/http/org/squbs/util/JavaConverters.scala
```diff
@@ -57,14 +57,20 @@ object JavaConverters {
     javaToScalaAdapter.atop(bidiFlow).atop(scalaToJavaAdapter).asJava
   }
 
-  private def adaptTupleFlow[T, Mat](scalaFlow: Flow[(HttpRequest, T), (Try[HttpResponse], T), Mat]):
-  js.Flow[Pair[jm.HttpRequest, T], Pair[Try[jm.HttpResponse], T], Mat] = {
+  private def adaptTupleFlow[T](scalaFlow: Flow[(HttpRequest, T), (Try[HttpResponse], T), HostConnectionPool]):
```
akara: Hmm, we removed a generic type `Mat` and replaced it with concrete type `HostConnectionPool`. It would work if `adaptTupleFlow` does not have any further use case. But was that the intention behind this?

anilgursel: `Mat` would only be `HostConnectionPool` in this case, right?  There is not other scenario.  Because, the flow is already designed as an http flow.

Issue #112, opened 2015-10-22
Upgrade to Scala 2.12


anilgursel:
Scala - dual build 2.11.7 and 2.12 (currently M3) We'll decide before release how ready Scala 2.12 is.
Note: Java API/adaptation may change between the two

PR #501, opened 2017-08-21
#112 Remove usage of FunctionConverters from Java 8 compat package

anilgursel:

- [x] Commits are squashed.
- [ ] Tests added.

### DIFF
--- a/project/Versions.scala
+++ b/project/Versions.scala
@@ -19,7 +19,6 @@ object Versions {
   val akkaHttpV = "10.0.7"
   val scalatestV = "3.0.3"
   val scalaLoggingV = "3.7.2"
-  val scalaJava8CompatV = "0.8.0"
   val jacksonV = "2.8.5"
   val json4sV = "3.5.3"
   val metricsV = "3.2.2"
--- a/squbs-actorregistry/build.sbt
+++ b/squbs-actorregistry/build.sbt
@@ -10,7 +10,6 @@ libraryDependencies ++= Seq(
   "com.typesafe.scala-logging" %% "scala-logging" % scalaLoggingV,
   "com.typesafe.akka" %% "akka-actor" % akkaV,
   "com.typesafe.akka" %% "akka-agent" % akkaV,
-  "org.scala-lang.modules" %% "scala-java8-compat" % scalaJava8CompatV,
   "com.typesafe.akka" %% "akka-testkit" % akkaV % "test",
   "ch.qos.logback" % "logback-classic" % "1.1.3" % "test",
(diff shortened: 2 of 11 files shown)


anilgursel:
I have basically done a search for `scala.compat.java8` in the code base and checked all the matches.  We have quite many `OptionConversters` and `FutureConverters` usage; however, no more `FunctionConverters`.

There is one more place we can do some improvement, please share your suggestions:

• https://github.com/paypal/squbs/blob/RELEASE-0.9.1/squbs-ext/src/main/scala/org/squbs/streams/circuitbreaker/japi/CircuitBreakerSettings.scala#L35

```scala
uniqueIdMapper: Function[Context, Optional[Any]] = new Function[Context, Optional[Any]] {
    override def apply(t: Context): Optional[Any] = Optional.empty()
  }
```

• https://github.com/paypal/squbs/blob/master/squbs-ext/src/main/scala/org/squbs/streams/circuitbreaker/japi/CircuitBreakerSettings.scala#L34

```scala
cleanUp: Consumer[Out] = new Consumer[Out] {
    override def accept(t: Out): Unit = ()
  },
```

akara:
I don't really see issues with tho code above. Yes, both of them make use of the Java 8 functional interfaces directly instead of closures. But these are in JAPI package so I see it as a legit use. Please let me know what issues you see here. Thanks!

anilgursel:
I think it is ok, but IntelliJ is not happy.  I tried to use lambda style, but I could not make it happy.

akara:
Here's how I made IntelliJ happy:

```scala
case class CircuitBreakerSettings[In, Out, Context] private[japi] (
  circuitBreakerState: CircuitBreakerState,
  fallback: Optional[Function[In, Try[Out]]] = Optional.empty[Function[In, Try[Out]]],
  cleanUp: Consumer[Out] = ((_: Out) => ()).asInstanceOf[Consumer[Out]],
  failureDecider: Optional[Function[Try[Out], Boolean]] = Optional.empty[Function[Try[Out], Boolean]],
  uniqueIdMapper: Function[Context, Optional[Any]] =
  ((_: Context) => Optional.empty()).asInstanceOf[Function[Context, Optional[Any]]]) {
```

akara:
Compilation went through. Tests are running. I'm just waiting to check whether I'll fail any tests with this.

akara:
Nope I take it back. I got a runtime exception. Looking further...

akara:
OK. Looks like we're dealing with a compiler bug. While I agree this issues is not immediately an issue, we should get it resolved.

Issue #502, opened 2017-08-22
Allow customizing ActorMaterializerSettings for HTTP endpoints


anilgursel:
A squbs app should be able customize `ActorMaterializer` for HTTP flows, or even provide a  `ActorMaterializer`.

PR #513, opened 2017-09-09
#502 Allow customizing ActorMaterializerSettings for HTTP endpoints

anilgursel:

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/squbs-unicomplex/src/main/resources/reference.conf
+++ b/squbs-unicomplex/src/main/resources/reference.conf
@@ -45,6 +45,9 @@ default-listener {
 
   # Any custom SSLContext provider? Setting to "default" means platform default.
   ssl-context = default
+
+  # Which materializer to use for HTTP streams.  default-materializer is used if not specified
+  # materializer = default-materializer
 }
 
 blocking-dispatcher {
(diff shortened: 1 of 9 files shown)


anilgursel:
I am yet to add JMX beans to list active materializer and the corresponding settings.

squbs-unicomplex/src/main/resources/reference.conf
```diff
@@ -78,8 +81,10 @@ blocking-dispatcher {
   throughput = 2
 }
 
-akka.http {
-  server {
-    pipelining-limit = 1
-  }
-}
```
sebady: With the removal here of the pipelining-limit.  Are we enabling http pipelining by default?

anilgursel: The configuration here does not guarantee disabling pipelining.  Akka HTTP enables it by default in `reference.conf` http://doc.akka.io/docs/akka-http/current/scala/http/configuration.html.  It becomes indeterministic which `reference.conf` to win.  Removing this config just makes it deterministic that it will always be enabled.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/Unicomplex.scala
```diff
@@ -754,3 +785,7 @@ class CubeSupervisor extends Actor with ActorLogging with GracefulStopHelper {
 private[unicomplex] class NoopActor extends Actor {
   def receive = PartialFunction.empty
 }
+
+class DefaultMaterializer {
+  def createMaterializer(implicit system: ActorSystem): ActorMaterializer = ActorMaterializer()
```
anilgursel: I should probably add a decider here which is same as the default, but just logs the exception.

Issue #519, opened 2017-09-26
High CPU usage with Circuit Breaker combined with ClientFlow


vertexclique:
There is a huge CPU usage with `CircuitBreaker` inside `ClientFlow`. Flow is trying to get elements but `NoSuchElementException` arises and wraps it to `None` and empty queue dequeuing allocates high CPU usage.

With configuration of:

```hocon
circuit-breaker {
  call-timeout = 5s
  reset-timeout = 10s
  max-failures = 5
  max-reset-timeout = 20s
}
```

and single instance of circuit breaker which is created with:
```scala
  val circuitBreakerName = s"circuit-breaker"

  val circuitBreakerState: CircuitBreakerState =
    AtomicCircuitBreakerState(
      name = circuitBreakerName,
      maxFailures = circuitBreakerConfig.maxFailures,
      callTimeout = circuitBreakerConfig.callTimeout,
      resetTimeout = circuitBreakerConfig.resetTimeout,
      maxResetTimeout = circuitBreakerConfig.maxResetTimeout
    )

  val cbSettings: CircuitBreakerSettings[HttpRequest, HttpResponse, Rewarding] =
    CircuitBreakerSettings[HttpRequest, HttpResponse, Rewarding](circuitBreakerState)
      .withFailureDecider(failureDecider)
      .withFallback(failureFallback)

  private def failureDecider(tryHttpResponse: Try[HttpResponse]): Boolean =
    tryHttpResponse.isFailure || tryHttpResponse.get.status.isFailure

  private def failureFallback(request: HttpRequest): Try[HttpResponse] = {
    val uri = request.uri.toString()
    val cbFailureBody = s"Failing fast after ${circuitBreakerConfig.maxFailures} attempts to $uri"

    logger.info(s"Circuit Breaker Activated for following url: $uri")

    Success(HttpResponse(status = StatusCodes.ServiceUnavailable, entity = HttpEntity(cbFailureBody)))
  }
```

and `ClientFlow` was initialized like:
```scala
    ClientFlow[Context](
      baseUrl,
      connectionContext = getSSLConnectionContext(scheme),
      circuitBreakerSettings = Some(circuitBreakerStorage.cbSettings)
    )
```

Here is the CPU sample showcase:
![screen shot 2017-09-26 at 14 26 41](https://user-images.githubusercontent.com/578559/30860626-120086e0-a2c8-11e7-990f-2b8463c912ea.png)

Also graphs:
[...]

PR #523, opened 2017-10-10
#519: fix high CPU usage with Circuit Breaker combined with ClientFlow

sebady:

- [ ] Description of the change added.
- [ ] Commits are squashed.
- [ ] Tests added.

### DIFF
--- a/squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
+++ b/squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
@@ -89,7 +89,13 @@ abstract class TimeoutGraphStageLogic[In, FromWrapped, Out](shape: BidiShape[In,
   protected def isBuffersEmpty: Boolean
 
   protected def timeLeftForNextElemToTimeout: Long = {
-    timeoutAsMillis - NANOSECONDS.toMillis(System.nanoTime() - firstElemStartTime)
+    val firstElemTime = firstElemStartTime
+    if (firstElemTime == 0) timeoutAsMillis
+    else {
+      val timeLeftInmillis = timeoutAsMillis - NANOSECONDS.toMillis(System.nanoTime() - firstElemTime)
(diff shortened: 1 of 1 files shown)


squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -89,7 +89,9 @@ abstract class TimeoutGraphStageLogic[In, FromWrapped, Out](shape: BidiShape[In,
   protected def isBuffersEmpty: Boolean
 
   protected def timeLeftForNextElemToTimeout: Long = {
-    timeoutAsMillis - NANOSECONDS.toMillis(System.nanoTime() - firstElemStartTime)
+    val firstElemTime  = firstElemStartTime
+    if (firstElemTime == 0) timeoutAsMillis
+    else timeoutAsMillis - NANOSECONDS.toMillis(System.nanoTime() - firstElemTime)
```
akara: There seems to be a chance for timeoutAsMillis to be 0 or negative. Should we set it to 0 in such cases?
Also, if the scheduled time needs to be below wakeup granularity (10ms on Linux), it may be better not to schedule but to execute the task right away.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -334,9 +336,13 @@ final class TimeoutBidiUnordered[In, Out, Context](timeout: FiniteDuration,
       } map { case(id, (context, _)) =>
         timeouts.remove(id)
         (Failure(FlowTimeoutException()), context)
-      } orElse Try(readyToPush.dequeue()).toOption.map { case(elem, _) => elem }
+      } orElse dequeueOption().map { case(elem, _) => elem }
     }
 
+    private def dequeueOption(): Option[((Try[Out], Context), Long)] =
+      if (readyToPush.nonEmpty) Some(readyToPush.dequeue())
+      else None
```
akara: I like this. It will make sure we do not throw/catch an exception.

PR #552, opened 2017-12-11
#520 Expose a codahale Metric for new connections rate

anilgursel:

- [x] Commits are squashed.
- [x] Tests added.


anilgursel:
We can definitely do two `meter`s one for connection creation one for connection close.  I agree what I have today would just give me what was the open connections in the last one minute, not how many opened in the last one minute.  

But, I think we should also keep the overall connection count rate, though.  If we just have two `meter`s, we won't be able to see if the number of active connections increased in the last 15 minutes or not.  We can get to that info; however, would be indirect.

akara:
Absolutely. I don't mean to do away with the connection count. That's extremely important. The new connection rate and termination rate tells us more of the connection behavior, whether we're dealing with a lot of reconnects or not.

docs/materialization-metrics-collector.md
```diff
@@ -0,0 +1,52 @@
...
+
+##### Scala
+
+```scala
+val stream = Source(1 to 10)
+  .via(MaterializationMetricsCollector[Int]("my-stream"))
+  .to(Sink.ignore)
+
+stream.run()
+stream.run()
```
akara: Eh? Duplicate?

docs/materialization-metrics-collector.md
```diff
@@ -0,0 +1,52 @@
...
+##### Java
+
+```java
+RunnableGraph<NotUsed> stream =
+        Source.range(1, 10)
+                .via(MaterializationMetricsCollector.create("my-stream", system))
+                .to(Sink.ignore());
+
+stream.run(mat);
+stream.run(mat);
```
akara: Same?

squbs-ext/src/main/scala/org/squbs/metrics/MetricsFlow.scala
```diff
@@ -64,3 +66,60 @@ object MetricsFlow {
...
+    setHandler(out, new OutHandler {
+      override def onPull(): Unit = pull(in)
+
+      override def onDownstreamFinish(): Unit = {
+        activeMaterializationCount.dec()
+        materializationTerminationCount.mark()
+        super.onDownstreamFinish()
+      }
+    })
+  }
```
akara: Looks good!

squbs-ext/src/test/scala/org/squbs/metrics/MaterializationMetricsCollectorSpec.scala
```diff
@@ -0,0 +1,160 @@
...
+      .via(MaterializationMetricsCollector[Int]("upstream-finishes"))
+      .toMat(Sink.ignore)(Keep.both)
+    val (probe1, future1) = stream.run()
+    val (probe2, future2) = stream.run()
+
+    jmxValue("upstream-finishes-active-count", "Count").value shouldBe 2
+    jmxValue("upstream-finishes-creation-count", "Count").value shouldBe 2
+
+    probe1.sendComplete()
+    Await.ready(future1, awaitMax)
```
akara: Just a new trick up the sleeve. You can mix in the `ScalaFutures` trait and do something like `future1.futureValue`. No need to change, though.

squbs-ext/src/test/scala/org/squbs/metrics/MaterializationMetricsCollectorSpec.scala
```diff
@@ -0,0 +1,160 @@
...
+      jmxValue("upstream-fails-termination-count", "Count").value shouldBe 3
+    }
+  }
+
+  it should "update metrics when downstream finishes" in {
+
+    val stream = TestSource.probe[Int]
+      .via(MaterializationMetricsCollector[Int]("downstream-finishes"))
+      .map { elem =>
+        if(elem == 3) throw new Exception("boom")
```
akara: Hmm, I'm not sure the test description coincides. In this case, downstream terminated but it really did not finish. Should it rather be "update metrics when downstream terminates"?

docs/materialization-metrics-collector.md
anilgursel: This is on purpose.  The explanation under the "Usage" section reads:

> `my-stream-creation-count` with `Count` value 2.

If you think this is confusing, I can try to reduce it to one or add a comment like 

```scala
stream.run() // Materializing the stream a second time
```

docs/materialization-metrics-collector.md
anilgursel: Same comment I have.  I am not sure if it explains well with single.  But, it is also obvious that it creates a confusion with double. I may change the example to and description to have only one.

squbs-ext/src/test/scala/org/squbs/metrics/MaterializationMetricsCollectorSpec.scala
anilgursel: I named it as such because I was trying to test `onDownstreamFinish`.  I can name it as "update metrics when downstream terminates".

docs/materialization-metrics-collector.md
```diff
@@ -0,0 +1,52 @@
...
+### Overview
+
+`MaterializationMetricsCollector` is an Akka Streams `GraphStage` to collect materialization metrics for a stream:
+
+   * active materialization counts
+   * new materialization creation rates
+   * materialization termination rates (aggregation of successful termination and failures)
+
+
+These metrics can provide details about the health and status of your application.  For instance, `MaterializationMetricsCollector` is out-of-the-box integrated to [squbs HTTP(S) service](http-services.md) implementations, so users can see the active connection count per [listener](http-services.md#listener-binding) as well as connection creation and termination rates (Akka HTTP creates a new materialization of the HTTP stream for each new connection).
```
akara: Let me try...

A prominent use case for the `MaterializationMetricsCollector` is with server-side Akka HTTP, where each new connection results in a stream materialization and therefore each connection termination results in the termination of the associated stream materialization. For this reason, the `MaterializationMetricsCollector` is out-of-the-box integrated with [squbs HTTP(S) service](http-services.md) implementations publishing active connection, connection creation, and connection termination metrics.

docs/materialization-metrics-collector.md
akara:
```scala
stream.run() // Materializing the stream a second time, thus increasing the Count value to 2
```

docs/materialization-metrics-collector.md
akara: Ditto.

squbs-ext/src/test/scala/org/squbs/metrics/MaterializationMetricsCollectorSpec.scala
akara: That would be clearer. Thanks!

Issue #312, opened 2016-11-10
Exception Mappers for Akka Http app flow


anilgursel:
Please see Akara's comment in this PR https://github.com/paypal/squbs/pull/303:

> Just roughly: <br>
Request ~> Pipeline ~> AppFlow ~> Pipeline ~> Response <br>
If AppFlow causes the stream to totally collapse due to a thrown exception, I'd like to catch it in the pipeline and perhaps send a more meaningful response than just a blank 500 that is the case today.

PR #553, opened 2017-12-12
#556: Add a response metric to MetricsFlow

sebady:

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/squbs-ext/src/main/scala/org/squbs/metrics/MetricsFlow.scala
+++ b/squbs-ext/src/main/scala/org/squbs/metrics/MetricsFlow.scala
@@ -34,6 +34,7 @@ object MetricsFlow {
 
     val requestCount = MetricRegistry.name(domain, s"$name-request-count")
     val requestTime = MetricRegistry.name(domain, s"$name-request-time")
+    val responseCount = MetricRegistry.name(domain, s"$name-response-count")
     val count2XX = MetricRegistry.name(domain, s"$name-2XX-count")
     val count3XX = MetricRegistry.name(domain, s"$name-3XX-count")
     val count4XX = MetricRegistry.name(domain, s"$name-4XX-count")
@@ -49,13 +50,15 @@ object MetricsFlow {
 
       rc.response map {
         case Success(response) =>
+          metrics.meter(responseCount).mark()
           val statusCode = response.status.intValue()
 
-          if(statusCode >= 500) metrics.meter(count5XX).mark()
-          else if(statusCode >= 400) metrics.meter(count4XX).mark()
-          else if(statusCode >= 300) metrics.meter(count3XX).mark()
-          else if(statusCode >= 200) metrics.meter(count2XX).mark()
+          if (statusCode >= 500) metrics.meter(count5XX).mark()
+          else if (statusCode >= 400) metrics.meter(count4XX).mark()
+          else if (statusCode >= 300) metrics.meter(count3XX).mark()
+          else if (statusCode >= 200) metrics.meter(count2XX).mark()
         case Failure(ex) =>
+          metrics.meter(responseCount).mark()
           val causeExceptionClass = Try(ex.getCause.getClass.getSimpleName) getOrElse ex.getClass.getSimpleName
           metrics.meter(MetricRegistry.name(domain, s"$name-$causeExceptionClass-count")).mark()
       }
--- a/squbs-unicomplex/src/test/scala/org/squbs/unicomplex/pipeline/MetricsFlowSpec.scala
+++ b/squbs-unicomplex/src/test/scala/org/squbs/unicomplex/pipeline/MetricsFlowSpec.scala
@@ -163,6 +163,19 @@ class MetricsFlowSpec extends TestKit(MetricsFlowSpec.boot.actorSystem) with Asy
     }
   }
 
+  it should "aggregate 5XX and 2XX counts in response-count metric" in {
+
+    val f = Source(hello("/sample7") :: internalServerError("/sample7") :: internalServerError("/sample7") :: Nil).
+      via(poolClientFlow).
+      runWith(Sink.ignore)
+
+    f map { _ =>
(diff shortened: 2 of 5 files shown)


squbs-ext/src/main/scala/org/squbs/metrics/MetricsFlow.scala
```diff
@@ -47,13 +48,15 @@ object MetricsFlow {
...
-          if(statusCode >= 500) metrics.meter(count5XX).mark()
-          else if(statusCode >= 400) metrics.meter(count4XX).mark()
-          else if(statusCode >= 300) metrics.meter(count3XX).mark()
-          else if(statusCode >= 200) metrics.meter(count2XX).mark()
+          if (statusCode >= 500) metrics.meter(count5XX).mark()
+          else if (statusCode >= 400) metrics.meter(count4XX).mark()
+          else if (statusCode >= 300) metrics.meter(count3XX).mark()
+          else if (statusCode >= 200) metrics.meter(count2XX).mark()
         case Failure(ex) =>
+          metrics.meter(responseCount).mark()
```
anilgursel: Do we actually send a response to the user in this scenario?  If so, what is the status code?  It should be 500?

Here we map it to 500, but this comes before the response pipeline: https://github.com/paypal/squbs/blob/RELEASE-0.9.2/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/FlowHandler.scala#L113

akara: Yes, and if we don't respond and it times out, a response is still sent.

sebady: This new response counter metric totals all success/failure responses.  This aggregate counter should make for easier comparison to requests counter for example.  It does overlap the existing n-xxx counters we have but hopefully that won't be too confusing.

anilgursel: > Yes, and if we don't respond and it times out, a response is still sent.

@akara we need to discuss and write all scenarios down to make sure the metrics are all correct, e.g.,

• what happens if the client side times out first
• what happens if the server times out.  What's the code path in such scenario?  We should not be incrementing 200 count.

anilgursel: We should also increment 5xx count here (please make sure and test the client actually gets a 500).  We have https://github.com/paypal/squbs/issues/312 to add an exception mapper; however, that should not change the status code and keep it as 5xx, but just allow to add payload, IMO.  @akara any comments?

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/pipeline/MetricsFlowSpec.scala
```diff
@@ -163,6 +163,31 @@ class MetricsFlowSpec extends TestKit(MetricsFlowSpec.boot.actorSystem) with Asy
...
+  }
+
+  it should "count exceptions in response-count metric" in {
+
+    val future = Source(hello("/sample8") :: connectionException("/sample8") :: timeoutException("/sample8") :: Nil).
+      via(poolClientFlow).
+      runWith(Sink.ignore)
+
+    future map { _ =>
+      jmxValue("sample8-response-count", "Count").value shouldBe 3
```
anilgursel: Do we return a response or not in this scenario? If we do, is it 5xx?  If 5xx, need to add an assertin for that.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/pipeline/MetricsFlowSpec.scala
```diff
@@ -163,6 +163,31 @@ class MetricsFlowSpec extends TestKit(MetricsFlowSpec.boot.actorSystem) with Asy
     }
   }
 
+  it should "count 5XX server responses in response-count metric" in {
```
anilgursel: Not clear what this test is doing on top of the existing ones.  Don't we test this already?

squbs-ext/src/main/scala/org/squbs/metrics/MetricsFlow.scala
anilgursel: Have you tested and seen we actually do not return a 5xx in this scenario?

sebady: Only on the client side. I add MetricsFlowSpec  check this scenario on the client side.   See "dropped" test.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/pipeline/MetricsFlowSpec.scala
sebady: Each request here (hello, connectionException and timeoutException) return a response. I can add an additional assertion for the 5xx counter but there were already several existing test in this suite for the 5xx metric

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/pipeline/MetricsFlowSpec.scala
sebady: Let me fix that

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/pipeline/MetricsFlowSpec.scala
anilgursel: Just want to make sure that 5XX count is incremented whenever we return a 5xx.  For the exception scenario in the other tests, I am not sure if we already cover 5xx (unless you added, it does not cover.  Because, that was an issue).

akara:
Can we please write out the scenarios that are already in the tests and ones that are not? If the PR is complete I hate to hold it longer. Just want to make sure it is complete.

Issue #575, opened 2018-01-11
Error in stage [RetryBidi] - "cannot push port twice" 


sebady:
This failure occurs intermittently in RetryBid on 0.9.X

Error in stage [RetryBidi]: requirement failed: Cannot push port (RetryBidi.out2) twice
java.lang.IllegalArgumentException: requirement failed: Cannot push port (RetryBidi.out2) twice
   at scala.Predef$.require(Predef.scala:224)
   at akka.stream.stage.GraphStageLogic.push(GraphStage.scala:463)
   at org.squbs.streams.RetryBidi$$anon$1$$anon$4.onPush(RetryBidi.scala:286)
   at akka.stream.impl.fusing.GraphInterpreter.processPush(GraphInterpreter.scala:747)
   at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:649)

anilgursel:
Please see the code at https://github.com/paypal/squbs/blob/master/squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala#L377

```scala
setHandler(in2, new InHandler {
      override def onPush(): Unit = {
        val (elem, context) = grab(in2)
        if (isFailure(elem)) {
          if (emitOrQueueFailure(context))
            push(out2, (elem, context))
          else if (isAvailable(out1)) pushIfReady()

          if (retryRegistry.nonEmpty && !hasBeenPulled(in2))
            pull(in2)
        } else {
          retryRegistry.get(uniqueId(context)) foreach(entry => {
            retryDelayQ.remove(entry._3)
            retryRegistry.remove(uniqueId(entry._2))
          })
          push(out2, (elem, context))
        }
      }
```

IMO, the problem happens when retries are exhausted.  Because we push to downstream, but without getting a demand from downstream we immediately create demand on `in2`.  For any element we actually send to downstream, we should not be creating a demand to `in2` until we get a demand from `out2`.  We should `pull(in2)` whenever we retry an element, but we should not do it for actual success and exhausted elements.  Please see the success scenario, we do not do `pull(in2)`, why would that scenario be different than retries exhausted scenario?

So, I think the demand creation should be inside `else if`:

```scala
if (emitOrQueueFailure(context)) push(out2, (elem, context))
else if (isAvailable(out1)) {
  pushIfReady()
  if (retryRegistry.nonEmpty && !hasBeenPulled(in2)) pull(in2)
}
```

sebady:
[...]

PR #581, opened 2018-01-16
 #575: Fix "cannot push port twice error"

sebady:
Check if out2 is available befor push.
Log if we cant push to out2 on any succesful/exhausted element

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
+++ b/squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
@@ -256,23 +256,23 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
     private val noDelay = delay == Duration.Zero
     private var upstreamFinished = false
 
-    private def emitOrQueueFailure(context: Context): Boolean =
+    private def queueFailure(context: Context): Boolean =
       retryRegistry.get(uniqueId(context)) match {
         case None =>
           log.debug("Element for context [{}] dropped", context)
-          true
+          false
         case Some((_, ctx, retryTracker)) if retryTracker.count >= maxRetries =>
           log.debug("Retries exhausted for context [{}]", context)
           retryRegistry.get(uniqueId(ctx)) foreach(entry => {
             retryDelayQ.remove(entry._3)
             retryRegistry -= uniqueId(entry._2)
           })
-          true
+          false
         case Some((_, ctx, retryTracker)) =>
           log.debug("Queueing retry {} for context [{}]", retryTracker.count + 1, context)
           updateTracker(ctx).foreach(tracker => retryDelayQ.add(tracker))
           if (!isTimerActive(timerName) && !noDelay) scheduleOnce(timerName, sleepTimeLeft)
-          false
+          true
       }
 
     // Some useful hidden types for pattern match
@@ -378,18 +378,21 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
       override def onPush(): Unit = {
         val (elem, context) = grab(in2)
         if (isFailure(elem)) {
-          if (emitOrQueueFailure(context))
-            push(out2, (elem, context))
-          else if (isAvailable(out1)) pushIfReady()
-
-          if (retryRegistry.nonEmpty && !hasBeenPulled(in2))
+          if (queueFailure(context)) {
+            if (isAvailable(out1)) pushIfReady()
+            // continue propagating demand on in2 if grabbed element is queued for retry
             pull(in2)
+          } else {
+            if (isAvailable(out2)) push(out2, (elem, context))
+            else log.error("out2 is not available for push.  Dropping exhausted element")
+          }
         } else {
           retryRegistry.get(uniqueId(context)) foreach(entry => {
             retryDelayQ.remove(entry._3)
             retryRegistry.remove(uniqueId(entry._2))
           })
-          push(out2, (elem, context))
+          if (isAvailable(out2)) push(out2, (elem, context))
+          else log.error("out2 is not available for push.  Dropping successful element")
         }
       }
 
--- a/squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
+++ b/squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
@@ -21,10 +21,10 @@ import java.util.concurrent.atomic.AtomicLong
 import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.stream.Attributes.inputBuffer
-import akka.stream.{ActorMaterializer, BufferOverflowException, OverflowStrategy}
+import akka.stream.{ActorMaterializer, BufferOverflowException, OverflowStrategy, ThrottleMode}
 import akka.stream.scaladsl.{Flow, Keep, Sink, Source}
 import akka.stream.testkit.scaladsl.{TestSink, TestSource}
-import akka.testkit.{TestKit, TestProbe}
+import akka.testkit.TestKit
 import org.scalatest.{AsyncFlatSpecLike, Matchers}
 
 import scala.concurrent.duration._
@@ -68,7 +68,7 @@ class RetryBidiSpec extends TestKit(ActorSystem("RetryBidiSpec")) with AsyncFlat
 
     val expected = (Success("a"), 1) :: (Success("b"), 2) :: (Success("c"), 3) :: Nil
     result map {
-      _ should contain theSameElementsAs expected
+      _ should contain theSameElementsInOrderAs expected
     }
   }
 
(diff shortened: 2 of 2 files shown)


squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -378,19 +378,20 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
...
-          if (emitOrQueueFailure(context))
-            push(out2, (elem, context))
-          else if (isAvailable(out1)) pushIfReady()
-
-          if (retryRegistry.nonEmpty && !hasBeenPulled(in2))
-            pull(in2)
+          if (emitOrQueueFailure(context)) {
+            if (isAvailable(out2)) push(out2, (elem, context))
+            else log.error("out2 is not available for push.  Dropping exhausted element")
+          } else if (isAvailable(out1)) pushIfReady()
```
anilgursel: Shouldn't we create demand for `in2` when if actually re-try an element?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -378,19 +378,20 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
...
           retryRegistry.get(uniqueId(context)) foreach(entry => {
             retryDelayQ.remove(entry._3)
             retryRegistry.remove(uniqueId(entry._2))
           })
-          push(out2, (elem, context))
+          if (isAvailable(out2)) push(out2, (elem, context))
+          else log.error("out2 is not available for push.  Dropping successful element")
         }
+        // continue propagating demand on in2 if we still have retries queue'd.
+        if (retryRegistry.nonEmpty && !hasBeenPulled(in2)) pull(in2)
```
anilgursel: The demand should be at the point when we decide to retry an element.  There are only two places where we should create a demand on `in2`:

• re-try scenarios.
• `onPull` of `out2`.

So, I suggest moving this to within `else if (isAvailable(out1)) pushIfReady()` block above.

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
```diff
@@ -497,4 +497,39 @@ class RetryBidiSpec extends TestKit(ActorSystem("RetryBidiSpec")) with AsyncFlat
     succeed
   }
 
+  it should "retry when the joined flow buffers" in {
+    val bottom = Flow[(Long, Long)].map {
+      case (elem, ctx) => if (ctx % 2 == 0) (failure, ctx) else (Success(elem), ctx) // fail every even element
+    } buffer(5, OverflowStrategy.backpressure)
```
anilgursel: Probably a `.` before `buffer`?  Also, moving it to a separate line may help readability.

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
```diff
@@ -497,4 +497,39 @@ class RetryBidiSpec extends TestKit(ActorSystem("RetryBidiSpec")) with AsyncFlat
     succeed
   }
 
+  it should "retry when the joined flow buffers" in {
```
anilgursel: I think the name of the test is not very clear.  If the joined flow buffers, we should just make sure we do not halt the system or causing the stream to stuck.  We do not want the demand on `in2` any greater than re-tries + actual demand from out2.

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
```diff
@@ -497,4 +497,39 @@ class RetryBidiSpec extends TestKit(ActorSystem("RetryBidiSpec")) with AsyncFlat
...
+      .via(retry.join(bottom))
+      .runWith(Sink.seq)
+
+    val expected = 1 to 100 map (x => if (x % 2 == 0) (failure, x) else (Success(x), x))
+    result map {
+      _ should contain theSameElementsInOrderAs expected
+    }
+  }
+
+  it should "retry when the downstream flow throttle" in {
```
anilgursel: Again the name of the test could be improved.

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
anilgursel: Also, I would suggest having a larger buffer, e.g., 50.

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
```diff
@@ -497,4 +497,39 @@ class RetryBidiSpec extends TestKit(ActorSystem("RetryBidiSpec")) with AsyncFlat
...
+    val bottom = Flow[(Long, Long)].map {
+      case (elem, ctx) => if (ctx % 2 == 0) (failure, ctx) else (Success(elem), ctx) // fail every even element
+    }
+    val retry = RetryBidi[Long, Long, Long](2)
+    val result = Source(1L to 100)
+      .map(x => (x, x))
+      .throttle(10, 10.millis, 10, ThrottleMode.shaping)
```
anilgursel: I think we should put throttle only in downstream.  The idea is what happens if upstream is fast but downstream is not able to keep up.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
sebady: Yes.  See below addition of pull(in2)

sebady: We now create demand on in2 if a retry is queue'd (emitOrQueueFailure returns false).  Whether out1 is available or not here we still create demand.   If we need to be more selective here let me know.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
sebady: I will adjust this appropriately.

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
```diff
@@ -497,4 +497,40 @@ class RetryBidiSpec extends TestKit(ActorSystem("RetryBidiSpec")) with AsyncFlat
...
+
+    val retry = RetryBidi[Long, Long, Long](2)
+    val result = Source(1L to 100)
+      .map(x => (x, x))
+      .via(retry.join(bottom))
+      .runWith(Sink.seq)
+
+    val expected = 1 to 100 map (x => if (x % 2 == 0) (failure, x) else (Success(x), x))
+    result map {
+      _ should contain theSameElementsInOrderAs expected
```
akara: Retry would cause things to go out of order. Why would we test `theSameElementsInOrderAs`?

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
```diff
@@ -497,4 +497,40 @@ class RetryBidiSpec extends TestKit(ActorSystem("RetryBidiSpec")) with AsyncFlat
...
+    val retry = RetryBidi[Long, Long, Long](2)
+    val result = Source(1L to 100)
+      .map(x => (x, x))
+      .via(retry.join(bottom))
+      .throttle(1, 10.millis, 10, ThrottleMode.shaping)
+      .runWith(Sink.seq)
+
+    val expected = 1 to 100 map (x => if (x % 2 == 0) (failure, x) else (Success(x), x))
+    result map {
+      _ should contain theSameElementsInOrderAs expected
```
akara: Ditto, please explain why in order?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -378,18 +378,21 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
       override def onPush(): Unit = {
         val (elem, context) = grab(in2)
         if (isFailure(elem)) {
-          if (emitOrQueueFailure(context))
-            push(out2, (elem, context))
-          else if (isAvailable(out1)) pushIfReady()
-
-          if (retryRegistry.nonEmpty && !hasBeenPulled(in2))
+          if (emitOrQueueFailure(context)) {
```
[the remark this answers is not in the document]
sebady: `queueFailure(context)` would be a better name here.  I will change it and cleanup readyOption parens here.

squbs-ext/src/test/scala/org/squbs/streams/RetryBidiSpec.scala
sebady: I can relax the order verification here on elements.  As well as above.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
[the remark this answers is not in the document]
sebady: I believe readyOption should keep the parens.  The `poll()` on retryDelayQ in readyOption removes an element here from the queue.

PR #583, opened 2018-01-22
PerpetualStream auto-recognize KillSwitch as materialized value (fixes #574)

### LINKED ISSUE #574: PerpetualStream should auto-recognize streams materializing to KillSwitch
This would avoid users having to override `shutdown()` if materialized value or first element of materialized value is a `KillSwitch`.

akara:
First materialized value being `KillSwitch` is auto-recognized by `shutdown`.

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/docs/perpetualstream.md
+++ b/docs/perpetualstream.md
@@ -1,18 +1,44 @@
-# Streams Lifecycle
-Akka Streams/Reactive stream needs to be integrated with the [Runtime Lifecycle](lifecycle.md) of the server. For this, an automated or semi-automated integration is provided through `PerpetualStream` Scala trait and `AbstractPerpetualStream` abstract class for Java. For fine-grained control over stream sources directly, `LifecycleManaged` provides a wrapping that can control any source component for proper stopping and shutdown so the flow can start/stop gracefully.
+# Perpetual Stream
 
-## PerpetualStream
-The `PerpetualStream` allows declaration of a stream that would start when the server starts and stop gracefully without dropping messages when the server stops. It gets exposed as the `PerpetualStream` trait for Scala and `AbstractPerpertualStream` abstract class for Java. For brevity, we'll refer to both as `PerpetualStream`.
+### Overview
+
+The `PerpetualStream` allows declaration of a stream that would start when the server starts and stop gracefully without dropping messages when the server stops. It is commonly used for message consumers from Kafka or JMS, but also used as a consolidation point for data from multiple streams received through HTTP requests.
+
+`PerpetualStream` can be customized in various ways to fit your streams' needs. Those are discussed in the sections listed below:
+
+* [Basic Use](#basic-use)
+* [Override Lifecycle State to run the stream](#override-lifecycle-state-to-run-the-stream)
+* [Shutdown Overrides](#shutdown-overrides)
+* [Kill Switch Overrides](#kill-switch-overrides)
+* [Receiving and forwarding a message to the stream](#receiving-and-forwarding-a-message-to-the-stream)
+* [Handling Stream Errors](#handling-stream-errors)
+* [Connecting a Perpetual Stream with an HTTP Flow](#connecting-a-perpetual-stream-with-an-http-flow)
+
+### Dependency
+
+`PerpetualStream` is part of core squbs. In general, you do not need to add an extra dependency. The classes are part of the following dependency:
+
+```scala
+"org.squbs" %% "squbs-unicomplex" % squbsVersion
+```
+
+### Usage
+
+The `PerpetualStream` gets exposed as the `PerpetualStream` trait for Scala and `AbstractPerpertualStream` abstract class for Java. For brevity, we'll refer to both as `PerpetualStream`.
+
+#### Basic Use
 
 ##### Scala
 
-Streams making use of `PerpetualStream` will want to conform to the following requirements, which will allow the hooks in `PerpetualStream` to work seamlessly with minimal amount of custom overrides:
+Streams making use of `PerpetualStream` will want to materialize to certain known types, allowing the hooks in `PerpetualStream` to work seamlessly with minimal amount of custom overrides. The options are:
 
-1. Have `killSwitch.flow` as the first stream processing stage after the source(s). `killSwitch` is a standard Akka `SharedKillSwitch` that is provided by the `PerpetualStream` trait.
-2. The stream materializes to a `Future` or a `Product` with a `Future` as its last element. A `Product` is the super class for `Tuple`, `List`, and much more. It is natural for `Sink`s to materialize to a `Future`. If more materialized values are asked for, it usually comes in some form of a `Product`. The `Sink` being the last element in the stream is also commonly materialized to the last element of the `Product`.
-3. This `Future` (materialized value or last element of the materialized value) represents the completion of the stream. In other words, this future is completed when the stream completes.
+* Materialize to a `Future[_]`, meaning a future of any type. In this case the shared `killSwitch` from `PerpetualStream` should be embedded or `shutdown()` would need to be overridden.
+* Materialize to a `(KillSwitch, Future[_])` tuple. The `KillSwitch` will be used for initiating the shutdown of the stream.
+* Materialize to a `List` or any `Product` (`List`s, `Tuple`s are all subtypes of `Product`) where the first element is a `KillSwitch` and the last element is a `Future`.
 
-If all the requirements above are met, no other custom overrides are needed for `PerpetualStream` to function. The following code shows a fully conformant `PerpetualStream`:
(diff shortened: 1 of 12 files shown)


docs/streams-lifecycle.md
```diff
@@ -35,13 +35,36 @@ class WellBehavedStream extends PerpetualStream[Future[Done]] {
...
+```
+
+That's it. These streams are well behaved because they materialize to the sink's materialized value, which is a `Future[Done]` in the first example, or a `(KillSwitch, Future[Done])` in the second one.
 
 ##### Java
 
 Streams making use of `AbstractPerpetualStream` will want to conform to the following requirements, which will allow the hooks in `AbstractPerpetualStream` to work seamlessly with minimal amount of custom overrides:
 
-1. Have `killSwitch().flow()` as the first stream processing stage after the source(s). `killSwitch` is a standard Akka `SharedKillSwitch` that is provided by `AbstractPerpetualStream`.
+1. Stream should materialize to `KillSwitch` as the first element of the materialized value - OR - have `killSwitch().flow()` as the first stream processing stage after the source(s). `killSwitch` is a standard Akka `SharedKillSwitch` that is provided by `AbstractPerpetualStream`.
```
anilgursel: Ditto

docs/streams-lifecycle.md
```diff
@@ -74,11 +97,40 @@ public class WellBehavedStream extends AbstractPerpetualStream<CompletionStage<D
...
+            }));
+    }
+```
+
+That's it. These streams are well behaved because they materialize to the sink's materialized value, which is a `CompletionStage<Done>` in the first example, or a `Pair<KillSwitch, CompletionStage<Done>>` in the second one.
 
 ### Override Lifecycle State to run the stream
 
-There may be scenarios where a stream need to be materialized at a different lifecycle than `active`.  In such scenarios, override `streamRunLifecycleState`, e.g.,:
+There may be scenarios where a stream need to be materialized at a different lifecycle than `active`. In such scenarios, override `streamRunLifecycleState`, e.g.,:
```
anilgursel: Two spaces is actually not a typo.  I remember taking a technical writing class, where they were advising this.  But, not something significant at all, IMO.  Also, for consistency across the page, you can remove.

squbs-unicomplex/src/main/scala/org/squbs/stream/PerpetualStream.scala
```diff
@@ -75,15 +75,24 @@ trait PerpetualStream[T] extends PerpetualStreamBase[T] {
...
   def shutdown(): Future[Done] = {
-    killSwitch.shutdown()
     import context.dispatcher
     matValue match {
-      case f: Future[_] => f.map(_ => Done)
-      case p: Product if p.productArity > 0 => p.productElement(p.productArity - 1) match {
+      case f: Future[_] =>
+        killSwitch.shutdown()
+        f.map(_ => Done)
+      case p: Product if p.productArity > 0 =>
```
anilgursel: Is there any scenario where the stream materializes to only `KillSwitch`?

docs/streams-lifecycle.md
```diff
@@ -35,13 +35,36 @@ class WellBehavedStream extends PerpetualStream[Future[Done]] {
 }
 ```
 
-That's it. This stream is well behaved because it materializes to the sink's materialized value, which is a `Future[Done]`.
+Alternatively, the following code shows another conformant `PerpetualStream` materializing its first element as `KillSwitch`:
```
anilgursel: I think this should be a separate sub header.  Also, this one and custom kill switch section should be next to each other.

squbs-unicomplex/src/main/scala/org/squbs/stream/PerpetualStream.scala
akara: No. If there is only one materialized value, it must be a `Future` or `CompletionStage`.

docs/streams-lifecycle.md
akara: 😄  I thought I placed a typo into my own text. In the times before variable character width, it was a much bigger issue. Right now it is pretty much a non-issue.

akara:
@anilgursel Let me take care of the doc as you suggested. This is a big change so it may not get done by today. Thanks!

docs/perpetualstream.md
```diff
@@ -0,0 +1,522 @@
+# PerpetualStream
+
+### Overview
+
+The `PerpetualStream` allows declaration of a stream that would start when the server starts and stop gracefully without dropping messages when the server stops. It is commonly used for message consumers from Kafka or JMS, but also used as a consolidation point for data from multiple streams received through HTTP requests.
+
+The `PerpetualStream` gets exposed as the `PerpetualStream` trait for Scala and `AbstractPerpertualStream` abstract class for Java. For brevity, we'll refer to both as `PerpetualStream`.
+
+`PerpetualStream` can be customized in various ways to fit your streams' needs. Those are discussed in the sections listed below:
```
anilgursel: I think this should go under "Usage" section.  My first reaction was to click "Basic Use" and it caused me to skip Dependency section.

docs/perpetualstream.md
```diff
@@ -0,0 +1,522 @@
+# PerpetualStream
```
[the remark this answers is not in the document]
anilgursel: Please change the link to Perpetual Stream in [mkdocs.yml](https://github.com/paypal/squbs/blob/master/mkdocs.yml#L33)

docs/streams-lifecycle.md
```diff
@@ -1,574 +1,5 @@
...
-}
-```
-
-Let's see what's happening here:
-`matValue` method finds the `RunnableGraph` component registered under the `/user/mycube/perpetualStreamWithMergeHub`.
-This happens to be the bootstrapped instance of our `PerpetualStreamWithMergeHub`.
-`alsoTo` expects result of `matValue` to be a `Sink` for `MyMessage`.
-I.e. `Sink[MyMessage, NotUsed]`. And as we've seen above this is exactly what `PerpetualStreamWithMergeHub.streamGraph` will produce. 
-(Remembering our two prospectives: here `alsoTo` looks at `PerpetualStreamWithMergeHub` from the outside prospective and sees a `Sink[MyMessage, NotUsed]`.)
+Akka Streams/Reactive stream needs to be integrated with the [Runtime Lifecycle](lifecycle.md) of the server. For this, an automated or semi-automated integration is provided through the [`PerpetualStream`](perpetualstream.md) infrastructure. For fine-grained control over stream sources directly, `LifecycleManaged` provides a wrapping that can control any source component for proper stopping and shutdown so the flow can start/stop gracefully.
```
anilgursel: I wonder how much trigger source is used.  If it is one of the main things, we probably need to re-visit the API and documentation.  I do not find it consistent with others and also not sure if certain things in the doc are outdated, e.g., `Source.actorPublisher`.

akara: `TriggerSource` is actually barely used. I'm OK with re-visiting. But that is definitely low-priority. Should be a separate issue and perhaps for 0.10.1 or even later.

docs/perpetualstream.md
akara: You mean at the top level of "Usage"? How would the ToC look like? Should we add indentation there, too?

docs/perpetualstream.md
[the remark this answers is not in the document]
akara: Thanks, will do.

akara:
As for moving, just moving `streams-lifecycle.md` to a different name should already reflect the move. Git is pretty good at recognizing that.

docs/perpetualstream.md
anilgursel: Yes, I meant top level of "Usage".  Indentation should not be any different than it is under "Overview".  Another option is to add "Dependency" to TOC, but you may need to delete or reword the last sentence before the TOC in that case "Those are discussed in the sections listed below:".

I am ok with both options.

anilgursel:
Regarding git move, I was referring to the git history on Perpetual Stream related lines.  Since those are in a different file which was not moved by `git mv`, the history is lost.  but, it is not a big deal, IMO.

akara:
I'll make sure history is not lost.

docs/streams-lifecycle.md
akara: Actually, this follows standard patterns and should be under "Basic Use". That's why I left it here. Starting with JMS support, this will become a very common pattern.  Please let me know if you think otherwise.

Latest update is ready.

anilgursel: What Paradox does with Scala vs Java tabs is really powerful and prevents the doc looking really long.  When things get long, sub sectioning makes it easier.  

But, it looks ok now.  I will merge once the PR is green.

Issue #559, opened 2017-12-21
Refactor Retry stage to use PriorityQueue


sebady:
Move from the existing DelayQueue to (java) PriorityQueue

PR #588, opened 2018-01-25
 #559: refactor RetryBidi to use PriorityQueue

sebady:
Use (java) PriorityQueue internally RetryBIDI for queue'd elements
currently being retried.
Also minor cleanup of Java API and some simplification of RetryTracker.

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/docs/flow-retry.md
+++ b/docs/flow-retry.md
@@ -36,9 +36,10 @@ Source("a" :: "b" :: "c" :: Nil)
 ##### Java
 
 ```java
-final BidiFlow<Pair[String, Context], Pair[String, Context],
-    Pair[Try[String], Context], Pair[Try[String], Context], NotUsed> retryBidi =
-    RetryBidi.create(2);
+final BidiFlow<Pair<String, Long>,
+               Pair<String, Long>,
+               Pair<Try<String>, Long>,
+               Pair<Try<String>, Long>, NotUsed> retryBidi = RetryBidi.create(2L);
 
 final Flow<String, String, NotUsed> flow =
     Flow.<String>create().map(s -> findAnEnglishWordThatStartWith(s));
@@ -72,17 +73,13 @@ RetryBidi[In, Out, Context](maxRetries: Int, uniqueIdMapper: Context => Option[A
 This `BidiFlow` can be joined with any flow that takes in a `(In, Context)` and outputs a `(Out, Context)`.
 
 ```scala
-case class MyContext(s: String, uuid: UUID)
+import org.squbs.streams.RetrySettings
 
-val retryBidi = RetryBidi[String, String, MyContext](maxRetries = 2, (context: MyContext) => Some(context.uuid))
-val flow = Flow[(String, MyContext)].mapAsyncUnordered(10) { elem =>
-  (ref ? elem).mapTo[(String, MyContext)]
-}
+case class MyContext(s: String, id: Long)
 
-Source("a" :: "b" :: "c" :: Nil)
-  .map( _ -> MyContext("dummy", UUID.randomUUID))
-  .via(retryBidi.join(flow))
-  .runWith(Sink.seq)
+val retrySettings =
+  RetrySettings[String, String, MyContext](maxRetries = 3)
+    .withUniqueIdMapper((context: MyContext) => Some(context.id))
 ```
 
 ###### Java
@@ -90,36 +87,37 @@ Source("a" :: "b" :: "c" :: Nil)
 The following API is used to pass a uniqueId mapper:
 
 ```java
-public class RetryBidi {
-    public static <In, Out, Context> BidiFlow<Pair<In, Context>,
-                                              Pair<In, Context>,
-                                              Pair<Try<Out>, Context>,
-                                              Pair<Try<Out>, Context>,
-                                              NotUsed>
-    create(Long maxRetries, Function<Context, Optional<Object>> uniqueIdMapper);
-}
-```
+import org.squbs.streams.RetrySettings;
 
+class MyContext {
+    private String s;
+    private long id;
+
+    public MyContext(String s, long id) {
+        this.s = s;
+        this.id = id;
+    }
+
+    public long id() {
+        return id;
+    }
+}
+
+RetrySettings<String, String, MyContext> retrySettings =
+        RetrySettings.<String, String, MyContext>create(3)
+                .withUniqueIdMapper(context -> Optional.of(context.id));
+```
 This `BidiFlow` can be joined with any flow that takes in a `akka.japi.Pair<In, Context>` and outputs a `akka.japi.Pair<Try<Out>, Context>`.
 
 ```java
-class MyContext {
-    private String s;
-    private UUID uuid;
-
-    public MyContext(String s, UUID uuid) {
-        this.s = s;
-        this.uuid = uuid;
-    }
-
-    public UUID uuid() {
-        return uuid;
-    }
-};
+final Function<MyContext, Optional<Object>> uniqueIdMapper = context -> Optional.of(context.id);
+RetrySettings<String, String, MyContext> retrySettings =
+        RetrySettings.<String, String, MyContext>create(maxRetries)
+                .withUniqueIdMapper(uniqueIdMapper);
 
 final BidiFlow<Pair<String, MyContext>, Pair<String, MyContext>,
                Pair<Try<String>, MyContext>, Pair<Try<String>, MyContext>, NotUsed>
-                retryBidi = RetryBidi.create(maxRetries, context -> Optional.of(context.uuid));
+                retryBidi = RetryBidi.create(retrySettings);
 
 final Flow<Pair<String, MyContext>, Pair<String, MyContext>, NotUsed> flow =
         Flow.<Pair<String, MyContext>>create()
@@ -183,9 +181,9 @@ val retryBidi = RetryBidi[String, Long](maxRetries = 3, delay = 200 millis)
 ##### Java
 
 ```java
-final BidiFlow<Pair[String, Context], Pair[String, Context],
-    Pair[Try[String], Context], Pair[Try[String], Context], NotUsed> retryBidi =
-    RetryBidi.create(3, Duration.create("200 millis"));
+final BidiFlow<Pair<String, Context>, Pair<String, Context>,
+    Pair<Try<String>, Context>, Pair<Try<String>, Context>, NotUsed> retryBidi =
+    RetryBidi.create(3L, Duration.create("200 millis"));
 
 ```
 ##### Exponential backoff
@@ -201,14 +199,14 @@ val retryBidi = RetryBidi[String, Long](maxRetries = 4, delay = 200 millis, expo
 ###### Java
 
 ```java
-final BidiFlow<Pair[String, Context], Pair[String, Context],
-    Pair[Try[String], Context], Pair[Try[String], Context], NotUsed> retryBidi =
-    RetryBidi.create(4, Duration.create("200 millis"), 2.0);
+final BidiFlow<Pair<String, Context>, Pair<String, Context>,
+    Pair<Try<String>, Context>, Pair<Try<String>, Context>, NotUsed> retryBidi =
+    RetryBidi.create(4L, Duration.create(200, TimeUnit.MILLISECONDS), 2.0);
 
 ```
 
-The first failure of any element will be retried after 200ms, and then any second attempt will be retried after 800ms.
-  In general the retry delay duration will continue to increase using the formula { delay duration } * N ^ exponentialBackOff.
+The first failure of any element will be retried after a delay of 200ms, and then any second attempt will be retried after 800ms.
+  In general the retry delay duration will continue to increase using the formula { delay duration } * N ^ exponentialBackOff (where N is the retry number).
 
 ##### Maximum delay
 An optional maximum delay duration (`maxDelay`) can also be specified to provide an upper bound on the exponential back delay
@@ -223,8 +221,8 @@ val retryBidi = RetryBidi[String, Long](maxRetries = 4, delay = 200 millis, expo
 ###### Java
 
 ```java
-final BidiFlow<Pair[String, Context], Pair[String, Context],
-    Pair[Try[String], Context], Pair[Try[String], Context], NotUsed> retryBidi =
-    RetryBidi.create(4, Duration.create("200 millis"), 2.0, Duration.create("400 millis"));
+final BidiFlow<Pair<String, Context>, Pair<String, Context>,
+    Pair<Try<String>, Context>, Pair<Try<String>, Context>, NotUsed> retryBidi =
+    RetryBidi.create(4L, Duration.create(200, TimeUnit.MILLISECONDS), 2.0, Duration.create("400 millis"));
 
 ```
--- a/squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
+++ b/squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
@@ -236,12 +236,11 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Int, unique
         case Some(InputBuffer(_, max)) => max
       }
 
-    case class RetryTracker(ctx: Context, count: Long, lastFailureTime: Long)
+    case class RetryTracker(ctx: Context, count: Long, nextRetryTime: Long)
     class RetryComparator() extends Comparator[RetryTracker] {
       // Note: this comparator imposes orderings that is inconsistent with equals.
       override def compare(r1: RetryTracker, r2: RetryTracker): Int =
-        if (noDelay) 0
-        else if (sleepTime(r1.count) < sleepTime(r2.count)) -1 else +1
+        (r1.nextRetryTime - r2.nextRetryTime).toInt
     }
     // Ordered from shorter retry sleep time to long retry sleep time
     private val retryQ = new PriorityQueue[RetryTracker](new RetryComparator())
@@ -310,15 +309,10 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Int, unique
       push(out1, (elem, ctx))
     }
 
-    def expired(retryTracker: RetryTracker): Boolean = {
-      val elapsedTime = System.nanoTime() - retryTracker.lastFailureTime
-      (sleepTime(retryTracker.count) - elapsedTime) <= 0
-    }
-
     def readyOption(): Option[(In, Context)] =
-      // peek if it has "expired" and remove if it has
+      // peek if it has "expired" and remove from PQ if it has
       Option(retryQ.peek()) flatMap { retryTracker =>
-        if (noDelay || expired(retryTracker)) {
+        if (noDelay || retryTracker.nextRetryTime <= System.nanoTime()) {
           retryQ.poll()
           retryRegistry.get(uniqueId(retryTracker.ctx))
         } // refactor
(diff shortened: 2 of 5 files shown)


docs/flow-retry.md
```diff
@@ -72,34 +72,40 @@ RetryBidi[In, Out, Context](maxRetries: Int, uniqueIdMapper: Context => Option[A
...
-  (ref ? elem).mapTo[(String, MyContext)]
-}
+case class MyContext(s: String, id: Long)
 
-Source("a" :: "b" :: "c" :: Nil)
-  .map( _ -> MyContext("dummy", UUID.randomUUID))
-  .via(retryBidi.join(flow))
-  .runWith(Sink.seq)
+val retrySettings =
+  RetrySettings[String, String, MyContext](3)
```
akara: Better to spell out the meaning of `(3)`, such as `RetrySettings[String, String, MyContext](maxRetries = 3)`.

docs/flow-retry.md
```diff
@@ -117,9 +123,14 @@ class MyContext {
     }
 };
 
+
+RetrySettings retrySettings =
```
akara: Eh? No type parameters on `RetrySettings`?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -264,13 +259,13 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
         case Some((_, ctx, retryTracker)) if retryTracker.count >= maxRetries =>
           log.debug("Retries exhausted for context [{}]", context)
           retryRegistry.get(uniqueId(ctx)) foreach(entry => {
-            retryDelayQ.remove(entry._3)
+            retryQ.remove(entry._3)
```
akara: Couple of stylistic and readability comments on this block:

1. `... foreach { entry =>` instead of `... foreach(entry => {`
2. If you need to refer to `._3`, `._2`, etc. The code is more readable if you use case matching

```scala
... foreach { case (foo, bar, baz) =>
  retryDelayQ.remove(baz)
  retryQ.remove(baz)
  retryRegistry -= uniqueId(bar)
```
If you don't use the first parameter, just `_` it saying you don't care. Of course, name the parameters appropriately.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -318,18 +313,28 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
...
-      log.debug("1st attempt for context [{}] ", tracker.ctx) // refactor
+      log.debug("1st attempt for context [{}] ", tracker.ctx)
       push(out1, (elem, ctx))
     }
 
     def readyOption(): Option[(In, Context)] =
-      Option(retryDelayQ.poll()) flatMap (retryTracker => {
-        retryRegistry.get(uniqueId(retryTracker.ctx))
+      // peek if it has "expired" and remove if it has
+      Option(retryQ.peek()) flatMap (retryTracker => {
```
akara: Same thing on the style `... flatMap { retryTracker =>`

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -388,15 +393,15 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
           }
         } else {
           retryRegistry.get(uniqueId(context)) foreach(entry => {
-            retryDelayQ.remove(entry._3)
+            retryQ.remove(entry._3)
```
akara: Ditto here.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -413,26 +418,26 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
 
     final override def onTimer(key: Any): Unit = {
       if (isAvailable(out1)) pushIfReady()
-      if (!retryDelayQ.isEmpty) scheduleOnce(timerName, sleepTimeLeft)
+      if (!retryQ.isEmpty) scheduleOnce(timerName, sleepTimeLeft)
     }
 
-    private def updateTracker(context: Context): Option[RetryTracker] =
+    private def updateTracker(context: Context): Unit =
```
akara: No need to do type annotation if it is private.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -511,14 +516,12 @@ case class RetrySettings[In, Out, Context] private[streams](
...
-  import scala.compat.java8.OptionConverters._
-
   // Java API
   def withUniqueIdMapper(uniqueIdMapper: JFunction[Context, Optional[Any]]): RetrySettings[In, Out, Context] =
     copy(uniqueIdMapper = UniqueId.javaUniqueIdMapperAsScala(uniqueIdMapper))
 
-  def withFailureDecider(failureDecider: Optional[JFunction[Try[Out], JBoolean]]): RetrySettings[In, Out, Context] =
-    copy(failureDecider = failureDecider.asScala.map(f => (out: Try[Out]) => f(out).asInstanceOf[Boolean]))
+  def withFailureDecider(failureDecider: JFunction[Try[Out], JBoolean]): RetrySettings[In, Out, Context] =
+    copy(failureDecider = Some((out: Try[Out]) => failureDecider(out).asInstanceOf[Boolean]))
```
akara: Good!

squbs-ext/src/test/java/org/squbs/streams/RetryBidiTest.java
```diff
@@ -283,4 +274,41 @@ public MyContext(String s, UUID uuid) {
...
+                            else
+                                return new Pair<>(Success.apply(elem.first()), elem.second());
+                        });
+
+        final Function<MyContext, Optional<Object>> uniqueIdMapper = context -> Optional.of(context.uuid);
+        final Function<Try<String>, Boolean> failureDecider =
+                out -> out.isFailure() || out.equals(Success.apply("a")); // treat "a" as failure
+
+        final RetrySettings<String, String, MyContext> retrySettings =
+                RetrySettings.<String, String, MyContext>create(3L)
```
akara: Kindly look at the Codacy feedback and validate whether we should follow or not.

squbs-ext/src/test/java/org/squbs/streams/RetryBidiTest.java
```diff
@@ -283,4 +274,41 @@ public MyContext(String s, UUID uuid) {
...
+        final CompletionStage<List<Try<String>>> result =
+                Source.from(Arrays.asList("a", "b", "c"))
+                        .map(s -> new Pair<>(s, new MyContext("dummy", UUID.randomUUID())))
+                        .via(retryFlow.join(bottom))
+                        .map(t -> t.first())
+                        .runWith(Sink.seq(), mat);
+
+        final List<Try<String>> expected = Arrays.asList(Success.apply("a"), Success.apply("c"), failure);
+        assertTrue(result.toCompletableFuture().get().containsAll(expected),
+                "Did not get the expected elements from retry stage");
```
akara: Isn't there a better assertion we can use? `assertTrue` is a very weak one.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -237,19 +236,15 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
...
-        }
-      }
-
-      override def compareTo(o: Delayed): Int =
-        if (noDelay) 0 else if (getDelay(TimeUnit.NANOSECONDS) < o.getDelay(TimeUnit.NANOSECONDS)) -1 else +1
+        else if (computeSleepTime(r1.count) < computeSleepTime(r2.count)) -1 else +1
     }
-    private val retryDelayQ = new DelayQueue[RetryTracker]()
+    // Ordered from shorter retry sleep time to long retry sleep time
+    private val retryQ = new PriorityQueue[RetryTracker](new RetryComparator())
```
akara: Did you get to evaluate the Java vs Scala priority queue?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -286,7 +281,7 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
     private def handleBufferFull(): Unit = overflowStrategy match {
       case `dropHead` =>
         val head = retryRegistry.head
-        retryDelayQ.remove(head._2._3)
+        retryQ.remove(head._2._3)
```
akara: I'd rather use matching here, makes code more readable:

```scala
val (elem, (_, _, tracker)) = retryRegistry.head
retryQ.remove(tracker)
retryRegistry -= elem
```

Ditto for subsequent ones.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
sebady: I did and direct  removal (`retryQ.remove(retryTracker)` was not available.   I still have a goal to build a Buffer with the properties we need in several of squbs BIDIs.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
sebady: Will fix here and a few other places.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -237,19 +236,15 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
...
         if (noDelay) 0
-        else {
-          val elapsedNanos = System.nanoTime() - lastFailureTime
-          unit.convert(computeSleepTime(count) - elapsedNanos, NANOSECONDS)
-        }
-      }
-
-      override def compareTo(o: Delayed): Int =
-        if (noDelay) 0 else if (getDelay(TimeUnit.NANOSECONDS) < o.getDelay(TimeUnit.NANOSECONDS)) -1 else +1
+        else if (computeSleepTime(r1.count) < computeSleepTime(r2.count)) -1 else +1
```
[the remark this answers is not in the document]
sebady: `compare` must return -1 / +1  (or 0) as return value here for ordering.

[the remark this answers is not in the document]
akara: I thought it has to be <0, 0, >0. But I may be wrong. Just don't like the unnecessary branching.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
akara: OK. Thanks!

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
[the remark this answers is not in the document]
sebady: You are correct.  We could return positive or negative value here.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -318,23 +313,33 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
...
-      }) match {
-        case Some(e) => Some(e._1, e._2)
+      // peek if it has "expired" and remove if it has
+      Option(retryQ.peek()) flatMap { retryTracker =>
+        if (noDelay || expired(retryTracker)) {
+          retryQ.remove(retryTracker)
+          retryRegistry.get(uniqueId(retryTracker.ctx))
+        } // refactor
+        else None
+      } match {
```
anilgursel: IMO, `map` would be easier to read here.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -318,23 +313,33 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
...
+        case Some((elem, ctx, _)) => Some((elem, ctx))
         case None => None
       }
 
+    def expired(retryTracker: RetryTracker): Boolean = {
+      val elapsedTime = System.nanoTime() - retryTracker.lastFailureTime
+      (computeSleepTime(retryTracker.count) - elapsedTime) <= 0
```
anilgursel: Do we need the parenthesis?

anilgursel: Should it be `<=0` or `<=10`, I mean precision time which is 10ms I believe.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -318,23 +313,33 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
...
+    }
+
     def pushIfReady(): Unit =
       readyOption() match {
-        case Some(elemWithContext) =>
-          log.debug("Retrying context {} ", elemWithContext._2)
-          push(out1, (elemWithContext._1, elemWithContext._2))
+        case Some((elem, ctx)) =>
+          log.debug("Retrying context {} ", ctx)
```
anilgursel: May it potentially be logging any sensitive info ?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -387,16 +392,16 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
             else log.error("out2 is not available for push.  Dropping exhausted element")
           }
         } else {
-          retryRegistry.get(uniqueId(context)) foreach(entry => {
-            retryDelayQ.remove(entry._3)
-            retryRegistry.remove(uniqueId(entry._2))
-          })
+          retryRegistry.get(uniqueId(context)) foreach { case (_, ctx, retryTracker) =>
+            retryQ.remove(retryTracker)
```
anilgursel: I am little confused here.  We receive a successful element from the joined flow; how can it be in the `retryQ`?  If we are receiving it from the joined flow, that means at some point we removed it from `retryQ`, right?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
anilgursel: Why don't we compute the sleep time at the time of adding an element to the queue and keep `System.nanoTime() + computeSleepTime` in `RetryTracker`?  Otherwise, we are unnecessarily making the same calculation multiple times for an element.

In addition, I am not sure if the above ordering is correct.  Assuming `r1` is the new element being added to queue and `r2` is for an element already in the queue.  It does not take into account the time passed since `r2` has been in the queue.  Let's say r2 is added to queue at relative time 10 seconds while it was its 4th retry and delay is 1 sec.  Then, around time 13 r1 gets added to queue and it is its second retry.  If you just think in terms of sleep time, then r1 would be ahead, which is incorrect.  I might be missing some points though, please clarify if so.

We should have a test case for this scenario.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
anilgursel: Ditto on `Boolean.box`

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
sebady: There are a few place in stream stages we use a Map and Queue and try to keep elements in "sync"  etc.  Buffer would just have the operations we need here and would replace the Map and Queue combinations.

docs/flow-retry.md
```diff
@@ -201,14 +212,14 @@ val retryBidi = RetryBidi[String, Long](maxRetries = 4, delay = 200 millis, expo
 ###### Java
 
 ```java
-final BidiFlow<Pair[String, Context], Pair[String, Context],
-    Pair[Try[String], Context], Pair[Try[String], Context], NotUsed> retryBidi =
-    RetryBidi.create(4, Duration.create("200 millis"), 2.0);
+final BidiFlow<Pair<String, Context>, Pair<String, Context>,
+    Pair<Try<String>, Context>, Pair<Try<String>, Context>, NotUsed> retryBidi =
+    RetryBidi.create(4L, Duration.create("200 millis"), 2.0);
```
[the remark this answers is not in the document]
sebady: We could simplify this and have `maxRetries` be an `Int` to simplify the API.

docs/flow-retry.md
```diff
@@ -201,14 +212,14 @@ val retryBidi = RetryBidi[String, Long](maxRetries = 4, delay = 200 millis, expo
...
 
 ```
 
-The first failure of any element will be retried after 200ms, and then any second attempt will be retried after 800ms.
-  In general the retry delay duration will continue to increase using the formula { delay duration } * N ^ exponentialBackOff.
+The first failure of any element will be retried after a delay of 200ms, and then any second attempt will be retried after 800ms.
```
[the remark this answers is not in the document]
sebady: The backoff computation is `{ delay duration } * N ^ exponentialBackOff`

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
sebady: r1 is < r2 with regards to PQ when the sleep time for r1 < r2.

An insertion (for the next try) is done using an updated tracker (with the current nanotime, of failure and retry count) (see `updateTracker`) . "time passes" equally for all elements already in the retryQ. 

Let me think about this idea of storing the "next retry time" instead of `lastFailureTime`
 in retryTracker

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -318,23 +313,33 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Long, uniqu
...
 
     def readyOption(): Option[(In, Context)] =
-      Option(retryDelayQ.poll()) flatMap (retryTracker => {
-        retryRegistry.get(uniqueId(retryTracker.ctx))
-      }) match {
-        case Some(e) => Some(e._1, e._2)
+      // peek if it has "expired" and remove if it has
+      Option(retryQ.peek()) flatMap { retryTracker =>
+        if (noDelay || expired(retryTracker)) {
+          retryQ.remove(retryTracker)
```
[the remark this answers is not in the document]
sebady: peek is not doing a search here.  `poll` will remove the element but that is premature if it has not expired.

docs/flow-retry.md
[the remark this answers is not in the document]
anilgursel: I see an inconsistency in Akka circuitbreaker and Backoff supervisor in implementing the exponential back off:

   * Akka Circuit Breaker uses exponential backoff here https://github.com/akka/akka/blob/v2.5.9/akka-actor/src/main/scala/akka/pattern/CircuitBreaker.scala#L940
  
   * Backoff supervisor - https://github.com/akka/akka/blob/v2.5.9/akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala#L158
 
I would suggest using one or the other for both Circuit Breaker and Retry (unless you think the scenarios are really different).

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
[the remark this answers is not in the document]
anilgursel: I am suggesting using `retryQ.poll()` instead of `retryQ.remove(retryTracker)`.  Remove calls `indexOf`, which has the following logic:

```java
private int indexOf(Object o) {
        if (o != null) {
            for (int i = 0; i < size; i++)
                if (o.equals(queue[i]))
                    return i;
        }
        return -1;
    }
```

docs/flow-retry.md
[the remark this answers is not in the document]
anilgursel: Don't have to address with this PR though.  We can do with a follow up PR, but needs to happen with 0.10, IMO.

PR #589, opened 2018-01-29
 #549: Adjust CircuitBreaker failurdecider JAPI

sebady:
Change the japi CircuitBreakerSettings for failuredecider to return Java
Boolean.
Also updated JavaTestKit -> TestKit in CircuitBreakerBidiFlowTest

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/squbs-ext/src/main/scala/org/squbs/streams/circuitbreaker/japi/CircuitBreakerSettings.scala
+++ b/squbs-ext/src/main/scala/org/squbs/streams/circuitbreaker/japi/CircuitBreakerSettings.scala
@@ -17,6 +17,7 @@ package org.squbs.streams.circuitbreaker.japi
 
 import java.util.Optional
 import java.util.function.{Consumer, Function}
+import java.lang.{Boolean => JBoolean}
 
 import org.squbs.streams.UniqueId
 import org.squbs.streams.circuitbreaker.CircuitBreakerState
@@ -34,7 +35,7 @@ case class CircuitBreakerSettings[In, Out, Context] private[japi] (
   cleanUp: Consumer[Out] = new Consumer[Out] {
     override def accept(t: Out): Unit = ()
   },
-  failureDecider: Optional[Function[Try[Out], Boolean]] = Optional.empty[Function[Try[Out], Boolean]],
+  failureDecider: Optional[Function[Try[Out], JBoolean]] = Optional.empty[Function[Try[Out], JBoolean]],
   uniqueIdMapper: Function[Context, Optional[Any]] = new Function[Context, Optional[Any]] {
     override def apply(t: Context): Optional[Any] = Optional.empty()
   }) {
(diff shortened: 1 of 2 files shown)


squbs-ext/src/test/java/org/squbs/streams/circuitbreaker/CircuitBreakerBidiFlowTest.java
```diff
@@ -354,4 +355,44 @@ public long next() {
...
+            circuitBreakerState.subscribe(getRef(), TransitionEvents.instance());
+            ref.tell("a", ActorRef.noSender());
+            ref.tell("b", ActorRef.noSender());
+            ref.tell("b", ActorRef.noSender());
+            expectMsgEquals(Open.instance());
+            expectMsgEquals(HalfOpen.instance());
+            ref.tell("a", ActorRef.noSender());
+            expectMsgEquals(Closed.instance());
+        }};
+    }
```
akara: Nice test!

squbs-ext/src/main/scala/org/squbs/streams/circuitbreaker/japi/CircuitBreakerSettings.scala
```diff
@@ -58,7 +59,7 @@ case class CircuitBreakerSettings[In, Out, Context] private[japi] (
       circuitBreakerState,
       fallback.asScala.map(f => (in: In) => f(in)),
       (out: Out) => cleanUp.accept(out),
-      failureDecider.asScala.map(f => (out: Try[Out]) => f(out)),
+      failureDecider.asScala.map(f => (out: Try[Out]) => f(out).asInstanceOf[Boolean]),
```
anilgursel: Instead of casting, should we use `Boolean.box`?  Please see https://stackoverflow.com/questions/9638492/conversion-of-scala-map-containing-boolean-to-java-map-containing-java-lang-bool .

sebady: I think `Boolean.box` could work but I believe it makes more sense when converting containers.

anilgursel: I believe it should be used whenever you need an explicit conversion, not sure why it would make more sense only for collections.

anilgursel: I am just wondering if there may be some compiler optimizations happening under the hood with `Boolean.box`.

crankydillob: `f(out).booleanValue` is another option.

sebady: Perhaps Predef `boolean2Boolean` would be cleaner to use here.

Issue #563, opened 2017-12-27
Add option to `OrderingStage` to make it lenient


anilgursel:
`OrderingStage` uses a `PriorityQueue`.  If the incoming messages are out of order to a great extent (let's say the next element is somehow dropped prior to reaching `OrderingStage`), it will cause a halt in the stream.  We should have options to let elements being passed down after certain numbers of elements reached or some time is passed.

akara:
Instead of having a timeout (involves complexity, scheduler), I'm thinking using buffer fill as the point to drop it and move forward. What do you think?

anilgursel:
I agree.  I think that's good enough for the first cut.  If we see some scenarios that would require timeout, we can add it later on.

akara:
Let me take a stab at this one, and all `OrderingStage` issues together.

PR #591, opened 2018-02-05
Re-worked OrderingStage into BoundedOrdering (Fixes #557, #562, #563)

### LINKED ISSUE #557: Add documentation for `OrderingStage`

akara:
Added sliding window, Java API, and documentation.

- [X] Commits are squashed.
- [X] Tests added.


squbs-ext/src/main/scala/org/squbs/streams/OrderWithin.scala
```diff
@@ -0,0 +1,180 @@
...
+  val out: Outlet[A] = Outlet[A]("Filter.out")
+  val shape: FlowShape[A, A] = FlowShape.of(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) {
+
+      var expectId: B = initialId
+
+      implicit val elementPriority: Ordering[A] = Ordering.by(getId).reverse
+      val pq: mutable.PriorityQueue[A] = mutable.PriorityQueue.empty
```
anilgursel: I wonder if it would be more efficient if it allowed it to be fixed size to prevent re-sizings behind the scenes.  Fixed one might cause unnecessary memory usage though (I mean, always assuming worst case scenario)..

squbs-ext/src/main/scala/org/squbs/streams/OrderWithin.scala
```diff
@@ -0,0 +1,180 @@
...
+      def outOfBand(element: A): Boolean = ordering.lt(getId(element), expectId)
+
+      setHandler(in, new InHandler {
+        override def onPush(): Unit = {
+
+          val elem = grab[A](in)
+
+          // 1. Element id matches expected id.
+          // 2. Element is out-of-band and missed buffer, should have arrived long ago, just send it down.
+          // 3. Buffer full, need to flush elements downstream. Potentially causes out-of-band later
```
anilgursel: "." at the end

squbs-ext/src/main/scala/org/squbs/streams/OrderWithin.scala
```diff
@@ -0,0 +1,180 @@
...
+ *
+ *           '''Emits when''' [[getId(elem)]] matches the current state, if not enqueues it to priority queue
+ *
+ *           '''Backpressures when''' downstream backpressures
+ *
+ *           '''Completes when''' upstream completes
+ *
+ *           '''Cancels when''' downstream cancels
+ */
+class OrderWithin[A, B](within: Int, initialId: B, nextId: B => B, getId: A => B)
```
anilgursel: Just think out loud..  Should we set `within` to a default value?  For instance to default internal buffer size (which can be overridden with `withAttributes`).  May that cause the stream to get to a starvation though as all other components would have the same internal buffer size, or maybe, internal buffer size -1.  

It also makes sense to require it so that users need to make a conscious choice.

squbs-ext/src/main/scala/org/squbs/streams/OrderWithin.scala
```diff
@@ -0,0 +1,180 @@
...
+object OrderWithin {
+
+  /**
+   * Scala API: Creates a [[Flow]] representing an [[OrderWithin]].
+   * Orders the stream within a sliding window of n elements specified by the within parameter. Each element passing
+   * through the [[OrderWithin]] has an id. The elements can be ordered by the ordering function and the next expected
+   * id can be derived from the current id.
+   *
+   * @param within   The number of elements to wait for an out-of-order element.
+   * @param initialId The identity of the first element getting to the [[OrderWithin]].
```
anilgursel: Instead of using "identity", I would suggest using "id" to make it simpler overall.  Applies to other parts of the file.

squbs-ext/src/main/scala/org/squbs/streams/OrderWithin.scala
```diff
@@ -0,0 +1,180 @@
...
+import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
+
+import scala.annotation.tailrec
+import scala.collection.immutable.Queue
+import scala.collection.mutable
+
+object OrderWithin {
+
+  /**
+   * Scala API: Creates a [[Flow]] representing an [[OrderWithin]].
```
anilgursel: I think it gets more clear when `* Scala API` and `* Java API` without the ":" are on their own lines, then leave an empty line.

squbs-ext/src/main/scala/org/squbs/streams/OrderWithin.scala
```diff
@@ -0,0 +1,180 @@
...
+ *
+ * @param within   The number of elements to wait for an out-of-order element.
+ * @param initialId The identity of the first element getting to the [[OrderWithin]].
+ * @param nextId    Function to calculate the subsequent identity from the current identity.
+ * @param getId     Function to get the identity from an element.
+ * @param ordering  Defines how the element ids are being ordered.
+ * @tparam A The input and output type of this stream component.
+ * @tparam B The type of the element's identity.
+ *
+ *           '''Emits when''' [[getId(elem)]] matches the current state, if not enqueues it to priority queue
```
anilgursel: Please fix the indentation.

squbs-ext/src/test/java/org/squbs/streams/OrderWithinTest.java
```diff
@@ -0,0 +1,145 @@
...
+    static final Materializer mat = ActorMaterializer.create(system);
+
+    @AfterClass
+    public static void cleanUp() {
+        system.terminate();
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testInvalidWithin() {
+        OrderWithin.create(-1, 1, i -> i + 1, Function.identity());
```
anilgursel: What would be the behaviour when `within` is set to `0`?

squbs-ext/src/main/scala/org/squbs/streams/OrderWithin.scala
```diff
@@ -0,0 +1,180 @@
...
+ *
+ *           '''Cancels when''' downstream cancels
+ */
+class OrderWithin[A, B](within: Int, initialId: B, nextId: B => B, getId: A => B)
+                         (implicit val ordering: Ordering[B])
+  extends GraphStage[FlowShape[A, A]] {
+
+  require(within > 0)
+
+  val in: Inlet[A] = Inlet[A]("Filter.in")
```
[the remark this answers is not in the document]
akara: I'd rather remove the type on the right side. Since the val is public, ScalaStyle requires it to have a type annotation.

squbs-ext/src/main/scala/org/squbs/streams/OrderWithin.scala
```diff
@@ -0,0 +1,180 @@
...
+  val out: Outlet[A] = Outlet[A]("Filter.out")
+  val shape: FlowShape[A, A] = FlowShape.of(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) {
+
+      var expectId: B = initialId
```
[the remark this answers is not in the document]
akara: Thinking about that, too. Just a tad longer. Will do.

squbs-ext/src/main/scala/org/squbs/streams/OrderWithin.scala
akara: I checked the API. It does not have a way to reserve space. Looked at it already.

squbs-ext/src/test/java/org/squbs/streams/OrderWithinTest.java
akara: Line 122: `require(within > 0)`

akara:
@anilgursel To your review points:
1. `mkdocs.yml` sorry for missing this. Will put it in.
2. The name ending with `Within`. I agree it sounds like a time bound based on Akka naming conventions. Could not come up with a better name yet. Kindly suggest a name that also gives the meaning of the space bound ordering in this case.
3. Custom implicit `Ordering` or even `Comparator` needs some idea. As I've changed the ordering to be `Ordering[B]` (ordering on the id) from `Ordering[A]` (ordering on the elements) which is generally a `Long`, or if you want to go to the extreme a `BigInteger`, it is becoming much harder to come up with an id type that does not already have natural ordering, and you really don't want to use one. Those are becoming very rare circumstances. I can't think of a good one. Suggestions definitely welcome.

squbs-ext/src/main/scala/org/squbs/streams/OrderWithin.scala
akara: Generally, you need a much bigger buffer than the default in the attributes. I'd rather stick with a conscious choice how to configure this buffer.

akara:
Just some idea. How about renaming the component `BoundedOrdering` and what is now the `within` parameter to `waitElems`. I'll sync the PR once we have an agreement on the naming. Thanks for all your comments.

squbs-ext/src/main/scala/org/squbs/streams/OrderWithin.scala
anilgursel: I agree.  But, I think we need to add a note in documentation that this stage may detach downstream and upstream demand to some extend, or temporarily for certain scenarios.  Let's say downstream is requested one element.  But, OrderWithin received an out-of-order element from upstream, then it would keep demanding until the element is received or `within` is reached.

squbs-ext/src/main/scala/org/squbs/streams/OrderWithin.scala
anilgursel: Yeah, I had checked as well.  We may use other implementation, but not worth until we see an issue and this gets hot.

squbs-ext/src/test/java/org/squbs/streams/OrderWithinTest.java
anilgursel: To test the boundary, should we have a test with 0 as well ?

Actually, should we allow 0, which would mean no ordering (equal to not having the stage), right?

anilgursel:
`BoundedOrdering` sounds good to me.  But, `waitElems` is not clear, IMO.  How about `max`, `maxOutOfOrder`, `maxBounded`?

squbs-ext/src/test/java/org/squbs/streams/OrderWithinTest.java
akara: Why would you put in the stage, then?

sebady:
What about `maxSlidingWindow`?

anilgursel:
It is not a sliding window, though.

squbs-ext/src/test/java/org/squbs/streams/OrderWithinTest.java
anilgursel: Normally, you would not.  Let's say you are debugging an issue and would like to see the behavior if the ordering is not there, instead of removing the stage and reconnecting graph, you can just set it to 0.

Is there any harm on allowing 0?

docs/flow-ordering.md
```diff
@@ -0,0 +1,83 @@
...
+3. The `nextId` parameter specifies the function to derive the subsequent element's identifier from the current element's identifier.
+4. The `getId` parameter specifies the function used to extract the identifier from the element.
+5. Optionally, a `Comparator` may be passed if the system does not know how to compare and order the identifier type.
+
+Initialization and usage can be seen in the following examples:
+
+##### Scala
+
+```scala
+// This is the sample message type.
```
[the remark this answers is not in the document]
akara: Have we not said that?

squbs-ext/src/test/java/org/squbs/streams/OrderWithinTest.java
akara: Just need to establish the behavior how exactly a buffer of 0 should work and what needs to be tested.

PR #592, opened 2018-02-07
Replaced `java.net.URI` with `akka.http.scaladsl.model.Uri` (Fixes #449)

### LINKED ISSUE #449: `HttpEndpoint` should use `akka.http.*.Uri` instead of `java.net.Uri`
In the previous `Resolver` chain design, there was only one `Endpoint` being shared for different purposes, and that was the reason of `java.net.URI` to be used.  However, in the new design, different usage can register different endpoint types.  Accordingly, `HttpEndpoint` is http specific, so we should be able to use Akka HTTP `Uri`.

akara:

- [X] Commits are squashed.
- [x] Tests added.


akara:
This was a quick mechanical fix according to the issue. However, I'm not certain the test cases cover Java use cases adequately. The `Uri` type used here is from `scaladsl`. There is not quite any sharing between `javadsl` and `scaladsl` and this may cause further complications. Needs some discussions before merging.

PR #595, opened 2018-02-08
#559 Address Retry ordering issues and refactor the logic

anilgursel:

- [x] Commits are squashed.
- [x] Tests added.

**Description:**

* Fixed ordering issue for retried elements.
* Fixed not completing the stage until there is further demand from downstream.
* A retried element with a lower retry time should be able to go to the head of the retry queue
* Optimizations on method calls and map accesses.
* Stream should complete when upstream completes. Should not require a new demand from downstream to complete (the tests were requiring an extra demand from sink).
* Dropped Overflow strategy support as anything other than backpressure  does not make sense for Retry.  It is not really a buffer, any element in retryregistry is actually inflight, which we cannot drop.

**Things to do:**

* Add test case for putting an element to the head of the queue when retry time is sooner than head's.
* Go over onDownstreamFinish scenarios.

### DIFF
--- a/squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
+++ b/squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
@@ -16,7 +16,7 @@
 
 package org.squbs.streams
 
-import java.util.{Comparator, Optional, PriorityQueue}
+import java.util.Optional
 import java.util.function.{Function => JFunction}
 import java.lang.{Boolean => JBoolean}
 
@@ -27,7 +27,6 @@ import akka.stream.Attributes.InputBuffer
 import akka.stream.scaladsl.{BidiFlow, Flow}
 import akka.stream.stage._
 import akka.stream._
-import akka.stream.OverflowStrategy._
 
 import scala.concurrent.duration._
 import scala.collection.mutable
@@ -64,7 +63,6 @@ object RetryBidi {
     * @param uniqueIdMapper the function that maps [[Context]] to a unique id
     * @param failureDecider function to determine if an element passed by the joined [[Flow]] is
     *                       actually a failure or not
-    * @param overflowStrategy the overflowStrategy to use on Retry buffer filling
     * @param delay            the delay duration between retrying each failed retry
     * @param exponentialBackoffFactor exponential factor the delay duration will be increased upon each retry
     * @param maxDelay the maximum delay duration during retry backoff
@@ -75,12 +73,11 @@ object RetryBidi {
     */
   def apply[In, Out, Context](maxRetries: Int, uniqueIdMapper: Context => Option[Any] = (_: Any) => None,
                               failureDecider: Option[Try[Out] => Boolean] = None,
-                              overflowStrategy: OverflowStrategy = OverflowStrategy.backpressure,
                               delay: FiniteDuration = Duration.Zero,
                               exponentialBackoffFactor: Double = 1.0,
                               maxDelay: FiniteDuration = Duration.Zero):
   BidiFlow[(In, Context), (In, Context), (Try[Out], Context), (Try[Out], Context), NotUsed] =
-    BidiFlow.fromGraph(new RetryBidi[In, Out, Context](maxRetries, uniqueIdMapper, failureDecider, overflowStrategy,
+    BidiFlow.fromGraph(new RetryBidi[In, Out, Context](maxRetries, uniqueIdMapper, failureDecider,
       delay, exponentialBackoffFactor, maxDelay))
 
   /**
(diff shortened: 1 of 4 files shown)


squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -236,124 +224,44 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Int, unique
...
-              else grabAndPush()
-            } else if (pullCondition && !upstreamFinished && !hasBeenPulled(in1)) pull(in1)
+        if(!retryQ.isEmpty && isHeadReady) {
+          val (elem, ctx, _) = retryRegistry(retryQ.dequeue()._2)
+          push(out1, (elem, ctx))
+          // If a timer is active, that would be for the element we just pushed down.  So, not valid anymore.
+          // Also, if the onTimer is called, cannot push down until a demand is created with onPull.
+          cancelTimer(timerName)
+        } else {
+          if (pullIn1Condition) pull(in1)
```
anilgursel: There is a potential of stream stuck here..  if onPush(in1) cannot push and element remains at the port, we probably need to grab the element here.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -236,124 +224,48 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Int, unique
...
+          push(out1, (elem, ctx))
+          // If a timer is active, that would be for the element we just pushed down.  So, not valid anymore.
+          // Also, if the onTimer is called, cannot push down until a demand is created with onPull.
+          cancelTimer(timerName)
+        } else {
+          if(isAvailable(in1)) grabAndPush
+          else {
+            if (pullIn1Condition) pull(in1)
+            // If the head is not ready yet, while there is a demand, we should schedule a timer.
+            if (!retryQ.isEmpty && !isTimerActive(timerName)) scheduleOnce(timerName, remainingDelay)
```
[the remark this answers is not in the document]
anilgursel: I agree.  That would prevent unnecessary scheduling of the timer.

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -236,124 +224,48 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Int, unique
...
-        retryRegistry.clear()
-        log.debug("Buffer full dropping buffer")
-        grabAndPush()
-      case `fail` =>
-        failStage(BufferOverflowException(s"Buffer overflow for Retry stage (max capacity was: $internalBufferSize)!"))
-      case `backPressure` => // NOP.  Don't grab any elements from upstream in backpressure mode when full
-      case _ ⇒
-        throw new IllegalStateException("Retry buffer overflow mode not supported")
-    }
+    private def pullIn1Condition = !hasBeenPulled(in1) && retryRegistry.size < internalBufferSize && !upstreamFinished
```
akara: Function name does not communicate that well. Can we make it `shouldPullIn1`? So to use it this will look like

```scala
else if (shouldPullIn1) pull(in1)
```

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -401,38 +338,22 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Int, unique
     })
 
     final override def onTimer(key: Any): Unit = {
-      if (isAvailable(out1)) pushIfReady()
-      if (!retryQ.isEmpty) scheduleOnce(timerName, sleepTimeLeft)
+      if (isAvailable(out1)) {
+        val (elem, ctx, _) = retryRegistry(retryQ.dequeue()._2)
+        push(out1, (elem, ctx))
+      }
```
akara: Normally, if there was a demand in `in2`, `out1` should automatically be available. Is there any chance a component on the `out1` -> `in2` loop would cause back-pressure making the demand not propagate? A common test case is a `ClientFlow` call which does not properly consume or discard its entity.

At this time we only rely on an `onPull` on `out1` after this point, and no longer a scheduler. Perhaps we should leave a comment.

anilgursel: Since `Retry` does not control the joined flow, we cannot guarantee the propagation.  Also, the propagation is async.  I am not clear what comment we should leave?  You mean something like below?

```scala
final override def onTimer(key: Any): Unit = {
  if (isAvailable(out1)) {
    val (elem, ctx, _) = retryRegistry(retryQ.dequeue()._2)
    push(out1, (elem, ctx))
  } // else push the element when the demand arrives with the next onPull
```

anilgursel:
@akara 

> Cancelling the scheduler and just rely on the next onPull is a good strategy. But it needs proper documentation.

You mean documentation as in comments or in the actual user facing documentation.  Since it is implementation detail, I'd prefer it to be in the code.  

Please see if the latest commit addresses your comments.

Issue #576, opened 2018-01-11
Unique Id Mapper should not return an option - Retry, CircuitBreaker, Timeout


anilgursel:
`uniqueIdMapper: Context => Option[Any]` should actually be `uniqueIdMapper: Context => Any` or an `Option` of the function provided.

PR #598, opened 2018-02-16
#576 Unique Id Mapper should not return an option

anilgursel:

- [x] Commits are squashed.
- [ ] Tests added.

### DIFF
--- a/docs/flow-timeout.md
+++ b/docs/flow-timeout.md
@@ -126,7 +126,7 @@ The following API is used to pass a uniqueId mapper:
 
 ```scala
 TimeoutBidiFlowUnordered[In, Out, Context](timeout: FiniteDuration,
-                                           uniqueIdMapper: Context => Option[Any])
+                                           uniqueIdMapper: Option[Context => Any])
 ```
   
   This `BidiFlow` can be joined with any flow that takes in a `(In, Context)` and outputs a `(Out, Context)`.
@@ -135,7 +135,7 @@ TimeoutBidiFlowUnordered[In, Out, Context](timeout: FiniteDuration,
 case class MyContext(s: String, uuid: UUID)
 
 val timeoutBidiFlow = TimeoutBidiFlowUnordered[String, String, MyContext](timeout,
-                                                                         (context: MyContext) => Some(context.uuid))
+                                                                         Some((context: MyContext) => context.uuid))
 val flow = Flow[(String, MyContext)].mapAsyncUnordered(10) { elem =>
   (ref ? elem).mapTo[(String, MyContext)]
 }
@@ -157,7 +157,7 @@ public class TimeoutBidiFlowUnordered {
 	                                          Pair<Out, Context>,
 	                                          Pair<Try<Out>, Context>,
 	                                          NotUsed>
-	create(FiniteDuration timeout, Function<Context, Optional<Object>> uniqueIdMapper);
+	create(FiniteDuration timeout, Function<Context, Object> uniqueIdMapper);
 }
 ```
 
@@ -183,7 +183,7 @@ final BidiFlow<Pair<String, MyContext>,
                Pair<String, MyContext>,
                Pair<Try<String>, MyContext>,
                NotUsed>
-    timeoutBidiFlow = TimeoutBidiFlowUnordered.create(timeout, context -> Optional.of(context.uuid));
+    timeoutBidiFlow = TimeoutBidiFlowUnordered.create(timeout, context -> context.uuid);
 
 final Flow<Pair<String, MyContext>, Pair<String, MyContext>, NotUsed> flow =
         Flow.<Pair<String, MyContext>>create()
--- a/squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
+++ b/squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
@@ -16,69 +16,27 @@
 
 package org.squbs.streams
 
+import java.lang.{Boolean => JBoolean}
 import java.util.Optional
 import java.util.function.{Function => JFunction}
-import java.lang.{Boolean => JBoolean}
 
 import akka.NotUsed
 import akka.http.org.squbs.util.JavaConverters
 import akka.japi.Pair
 import akka.stream.Attributes.InputBuffer
+import akka.stream._
 import akka.stream.scaladsl.{BidiFlow, Flow}
 import akka.stream.stage._
-import akka.stream._
 
-import scala.concurrent.duration._
 import scala.collection.mutable
-import scala.concurrent.duration.{Duration, FiniteDuration}
-import scala.util.{Failure, Try}
+import scala.concurrent.duration.{Duration, FiniteDuration, _}
+import scala.util.Try
 
 object RetryBidi {
-  /**
-    * Creates a [[BidiFlow]] that can be used to provide Retry functionality.
-    * This API is specifically for flows that are using [[Try]]'s for elements that may occasionally fail.  By default,
-    * any [[Failure]] is considered a failure that should be retried. However, a failureDecider function can be
-    * specified to control what constitutes a failure.
-    *
-    * Retry functionality requires each element passing through this flow to be uniquely identifiable for retrying, so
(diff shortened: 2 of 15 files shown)


squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -71,7 +71,7 @@ object RetryBidi {
     * @tparam Context the type of the context that is carried along with the elements.
     * @return a [[BidiFlow]] with Retry functionality
     */
-  def apply[In, Out, Context](maxRetries: Int, uniqueIdMapper: Context => Option[Any] = (_: Any) => None,
+  def apply[In, Out, Context](maxRetries: Int, uniqueIdMapper: Context => Any = (c: Context) => c,
```
akara: I think it is more clear and compact if you say `uniqueIdMapper: Context => Any = identity`. Please try it out.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -300,11 +300,9 @@ final class TimeoutBidiUnordered[In, Out, Context](timeout: FiniteDuration,
...
 
   private[streams] def uniqueId(context: Context) =
-    uniqueIdMapper(context).getOrElse {
-      context match {
-        case uniqueIdProvider: UniqueId.Provider ⇒ uniqueIdProvider.uniqueId
-        case context ⇒ context
-      }
+    uniqueIdMapper(context) match {
+      case uniqueIdProvider: UniqueId.Provider ⇒ uniqueIdProvider.uniqueId
+      case uniqueId ⇒ uniqueId
```
akara: One reason for getting rid of `Context => Option[Any]` is to optimize the decision to be just once. I don't think we have achieved this just yet. If you would model it as `Option[Context => Any]` then you could make this decision upfront. Does not mean the API needs to show `Option` or `Optional`. That we can do behind the scenes instead of a function that gets called all the time. Just for consideration...

squbs-ext/src/main/scala/org/squbs/streams/UniqueId.scala
```diff
@@ -37,11 +37,8 @@ object UniqueId {
...
   }
 
-  private[squbs] def javaUniqueIdMapperAsScala[Context](uniqueIdMapper: JFunction[Context, Optional[Any]]):
-  Context => Option[Any] = {
-
-    import scala.compat.java8.OptionConverters._
-
-    (context: Context) => toScala(uniqueIdMapper(context))
+  private[squbs] def javaUniqueIdMapperAsScala[Context](uniqueIdMapper: JFunction[Context, Any]):
+  Context => Any = {
```
akara: `... Context => Any = uniqueIdMapper.apply`

May need an `_` after the apply just in case the compiler does not recognize it.

squbs-ext/src/main/scala/org/squbs/streams/circuitbreaker/japi/CircuitBreakerSettings.scala
```diff
@@ -60,7 +60,7 @@ case class CircuitBreakerSettings[In, Out, Context] private[japi] (
       fallback.asScala.map(f => (in: In) => f(in)),
       (out: Out) => cleanUp.accept(out),
       failureDecider.asScala.map(f => (out: Try[Out]) => f(out).asInstanceOf[Boolean]),
-      UniqueId.javaUniqueIdMapperAsScala(uniqueIdMapper))
+      uniqueIdMapper.asScala)
```
akara: Ditto. You don't need a function converter to convert a Java 8 `Function` to Scala. Just the `Function.apply _` is by itself a Scala function.

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
anilgursel: The only thing we could save is the `match`, we still need to call the function.  Because the `match` won't change based on the request.

However, do we want to optimize not calling `uniqueIdMapper` when one not provided?  I would assume if we set it to `identity`, compiler would do the optimization.

akara: Yes, I also assume `identity` will reach that same result. Compiler should optimize it away. So if we can make sure the match does not happen for every message, I'm good.

squbs-ext/src/main/scala/org/squbs/streams/UniqueId.scala
anilgursel: I had actually eliminated all uses of this function, but apparently, forgot to delete it.

anilgursel:
@akara Please check if the last commit addresses the concerns.  Please note, I am not very happy how the API currently looks and also, there is some inconsistency with `TimeoutBidiUnordered` (which takes an `Option`).   However, I will address the concerns with another PR as follows:

* Rename stages as `Retry`, `CircuitBreaker`, `Timeout`.
* Each takes a `Settings`, defined inside the corresponding companion object.
* Retry, CircuitBreaker and Timeout will have APIs that take settings, but also their simple form, e.g., Retry(5).

docs/flow-timeout.md
```diff
@@ -126,7 +126,7 @@ The following API is used to pass a uniqueId mapper:
 
 ```scala
 TimeoutBidiFlowUnordered[In, Out, Context](timeout: FiniteDuration,
-                                           uniqueIdMapper: Context => Option[Any])
+                                           uniqueIdMapper: OPtion[Context => Any])
```
akara: Eh? Typo in `OPtion`?

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -204,13 +163,12 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Int, unique
   require(exponentialBackoffFactor >= 0.0, "backoff factor must be >= 0.0")
   require(maxDelay == Duration.Zero || maxDelay >= delay, "maxDelay must be larger than delay")
 
-  private[streams] def uniqueId(context: Context) =
-    uniqueIdMapper(context).getOrElse {
-      context match {
-        case uniqueIdProvider: UniqueId.Provider ⇒ uniqueIdProvider.uniqueId
-        case `context` ⇒ `context`
-      }
+  val uniqueId: Context => Any = uniqueIdMapper.getOrElse{
```
akara: Yes, I do like this much more. Thanks!

squbs-ext/src/main/scala/org/squbs/streams/TimeoutBidi.scala
```diff
@@ -299,13 +298,12 @@ final class TimeoutBidiUnordered[In, Out, Context](timeout: FiniteDuration,
   private val out = Outlet[(Try[Out], Context)]("TimeoutBidiUnordered.out")
   val shape = BidiShape(in, toWrapped, fromWrapped, out)
 
-  private[streams] def uniqueId(context: Context) =
-    uniqueIdMapper(context).getOrElse {
-      context match {
-        case uniqueIdProvider: UniqueId.Provider ⇒ uniqueIdProvider.uniqueId
-        case context ⇒ context
-      }
+  val uniqueId: Context => Any = uniqueIdMapper.getOrElse{
```
akara: 👍

squbs-ext/src/main/scala/org/squbs/streams/RetryBidi.scala
```diff
@@ -423,14 +381,14 @@ final class RetryBidi[In, Out, Context] private[streams](maxRetries: Int, unique
   */
 case class RetrySettings[In, Out, Context] private[streams](
    maxRetries: Int,
-   uniqueIdMapper: Context => Option[Any] = (_: Any) => None,
+   uniqueIdMapper: Option[Context => Any] = None,
```
akara: We wanted to rename `RetrySettings` to `Retry.Settings`, right? Is that supposed to be in a different PR?

anilgursel: Yes, that will be a separate PR.

Issue #451, opened 2017-04-18
Change dependencies in squbs-pattern and squbs-ext to "provided"


akara:
For most external dependencies, unless they are needed for ALL functionality of said modules.

This reflects comments in https://github.com/akka/alpakka/issues/258 not wanting the dependencies to be heavyweight, or not even pulling them in unless absolutely needed.

akara:
Using "optional" instead of "provided" as it matches the expected semantics better.

PR #608, opened 2018-03-02
 #451: Add java friendly api for PipelineFlowFactory

sebady:
Add Java friendly PipelineFlowFactory API.
Add additional javadsl HttpRequest/Response APIs into RequestContext

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -110,8 +110,8 @@ class DummyBidiFlow extends PipelineFlowFactory {
 
   override def create(context: Context)(implicit system: ActorSystem): PipelineFlow = {
      BidiFlow.fromGraph(GraphDSL.create() { implicit b =>
-      val inbound = b.add(Flow[RequestContext].map { rc => rc.addRequestHeader(RawHeader("DummyRequest", "ReqValue")) })
-      val outbound = b.add(Flow[RequestContext].map{ rc => rc.addResponseHeader(RawHeader("DummyResponse", "ResValue"))})
+      val inbound = b.add(Flow[RequestContext].map { rc => rc.withRequestHeader(RawHeader("DummyRequest", "ReqValue")) })
+      val outbound = b.add(Flow[RequestContext].map{ rc => rc.withResponseHeader(RawHeader("DummyResponse", "ResValue"))})
       BidiShape.fromFlows(inbound, outbound)
     })
   }
@@ -128,10 +128,10 @@ public class DummyBidiFlow extends AbstractPipelineFlowFactory {
         return BidiFlow.fromGraph(GraphDSL.create(b -> {
             final FlowShape<RequestContext, RequestContext> inbound = b.add(
                     Flow.of(RequestContext.class)
-                            .map(rc -> rc.addRequestHeader(RawHeader.create("DummyRequest", "ReqValue"))));
+                            .map(rc -> rc.withRequestHeader(RawHeader.create("DummyRequest", "ReqValue"))));
             final FlowShape<RequestContext, RequestContext> outbound = b.add(
                     Flow.of(RequestContext.class)
-                            .map(rc -> rc.addResponseHeader(RawHeader.create("DummyResponse", "ResValue"))));
+                            .map(rc -> rc.withResponseHeader(RawHeader.create("DummyResponse", "ResValue"))));
 
             return BidiShape.fromFlows(inbound, outbound);
         }));
@@ -156,10 +156,10 @@ class DummyAbortableBidiFlow extends PipelineFlowFactory {
 
     BidiFlow.fromGraph(GraphDSL.create() { implicit b =>
       import GraphDSL.Implicits._
-      val inboundA = b.add(Flow[RequestContext].map { rc => rc.addRequestHeader(RawHeader("keyInA", "valInA")) })
-      val inboundC = b.add(Flow[RequestContext].map { rc => rc.addRequestHeader(RawHeader("keyInC", "valInC")) })
-      val outboundA = b.add(Flow[RequestContext].map { rc => rc.addResponseHeaders(RawHeader("keyOutA", "valOutA"))})
-      val outboundC = b.add(Flow[RequestContext].map { rc => rc.addResponseHeaders(RawHeader("keyOutC", "valOutC"))})
+      val inboundA = b.add(Flow[RequestContext].map { rc => rc.withRequestHeader(RawHeader("keyInA", "valInA")) })
+      val inboundC = b.add(Flow[RequestContext].map { rc => rc.withRequestHeader(RawHeader("keyInC", "valInC")) })
+      val outboundA = b.add(Flow[RequestContext].map { rc => rc.withResponseHeader(RawHeader("keyOutA", "valOutA"))})
+      val outboundC = b.add(Flow[RequestContext].map { rc => rc.withResponseHeader(RawHeader("keyOutC", "valOutC"))})
 
       val inboundOutboundB = b.add(authorization abortable)
 
@@ -198,16 +198,16 @@ public class DummyAbortableBidiFlow extends japi.PipelineFlowFactory {
         return BidiFlow.fromGraph(GraphDSL.create(b -> {
             final FlowShape<RequestContext, RequestContext> inboundA = b.add(
                 Flow.of(RequestContext.class)
-                    .map(rc -> rc.addRequestHeader(RawHeader.create("keyInA", "valInA"))));
+                    .map(rc -> rc.withRequestHeader(RawHeader.create("keyInA", "valInA"))));
             final FlowShape<RequestContext, RequestContext> inboundC = b.add(
                 Flow.of(RequestContext.class)
-                    .map(rc -> rc.addRequestHeader(RawHeader.create("keyInC", "valInC"))));
+                    .map(rc -> rc.withRequestHeader(RawHeader.create("keyInC", "valInC"))));
             final FlowShape<RequestContext, RequestContext> outboundA = b.add(
                 Flow.of(RequestContext.class)
-                    .map(rc -> rc.addRequestHeader(RawHeader.create("keyOutA", "valOutA"))));
+                    .map(rc -> rc.withRequestHeader(RawHeader.create("keyOutA", "valOutA"))));
             final FlowShape<RequestContext, RequestContext> outboundC = b.add(
                 Flow.of(RequestContext.class)
-                    .map(rc -> rc.addResponseHeader(RawHeader.create("keyOutC", "valOutC"))));
+                    .map(rc -> rc.withResponseHeader(RawHeader.create("keyOutC", "valOutC"))));
 
             final BidiShape<RequestContext, RequestContext> inboundOutboundB =
                 b.add(abortable(authorization));
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -189,7 +189,7 @@ class DummyAbortableBidiFlow extends PipelineFlowFactory {
 ##### Java
 
 ```java
-public class DummyAbortableBidiFlow extends AbstractPipelineFlowFactory {
+public class DummyAbortableBidiFlow extends japi.PipelineFlowFactory {
 
     @Override
     public BidiFlow<RequestContext, RequestContext, RequestContext, RequestContext, NotUsed>
@@ -210,7 +210,7 @@ public class DummyAbortableBidiFlow extends AbstractPipelineFlowFactory {
                     .map(rc -> rc.addResponseHeader(RawHeader.create("keyOutC", "valOutC"))));
 
             final BidiShape<RequestContext, RequestContext> inboundOutboundB =
-                b.add(Pipeline.abortable(authorization));
+                b.add(abortable(authorization));
 
             b.from(inboundA).toInlet(inboundOutboundB.in1());
             b.to(inboundC).fromOutlet(inboundOutboundB.out1());
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -103,6 +103,8 @@ dummyflow {
 
 A sample `DummyBidiFlow` looks like below:
 
+##### Scala
+
 ```scala
 class DummyBidiFlow extends PipelineFlowFactory {
 
@@ -116,6 +118,27 @@ class DummyBidiFlow extends PipelineFlowFactory {
 }
 ```
 
+##### Java
+
+```java
+public class DummyBidiFlow extends AbstractPipelineFlowFactory {
+
+    @Override
+    public BidiFlow<RequestContext, RequestContext, RequestContext, RequestContext, NotUsed> create(Context context, ActorSystem system) {
+        return BidiFlow.fromGraph(GraphDSL.create(b -> {
+            final FlowShape<RequestContext, RequestContext> inbound = b.add(
+                    Flow.of(RequestContext.class)
+                            .map(rc -> rc.addRequestHeader(RawHeader.create("DummyRequest", "ReqValue"))));
+            final FlowShape<RequestContext, RequestContext> outbound = b.add(
+                    Flow.of(RequestContext.class)
+                            .map(rc -> rc.addResponseHeader(RawHeader.create("DummyResponse", "ResValue"))));
+
+            return BidiShape.fromFlows(inbound, outbound);
+        }));
+    }
+}
+```
+
 #### Aborting the flow
 In certain scenarios, a stage in pipeline may have a need to abort the flow and return an `HttpResponse`, e.g., in case of authentication/authorization.  In such scenarios, the rest of the pipeline should be skipped and the request should not reach to the squbs service.  To skip the rest of the flow: 
 
@@ -123,6 +146,8 @@ In certain scenarios, a stage in pipeline may have a need to abort the flow and 
 * call `abortWith` on `RequestContext` with an `HttpResponse` when you need to abort.
 
 In the below `DummyAbortableBidiFlow ` example, `authorization ` is a bidi flow with `abortable` and it aborts the flow is user is not authorized: 
+
+##### Scala
 
 ```scala
 class DummyAbortableBidiFlow extends PipelineFlowFactory {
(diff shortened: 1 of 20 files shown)


docs/pipeline.md
```diff
@@ -161,6 +186,54 @@ class DummyAbortableBidiFlow extends PipelineFlowFactory {
...
+                .map(rc -> rc.addRequestHeader(RawHeader.create("keyInA", "valInA"))));
+            final FlowShape<RequestContext, RequestContext> inboundC = b.add(Flow.of(RequestContext.class)
+                .map(rc -> rc.addRequestHeader(RawHeader.create("keyInC", "valInC"))));
+            final FlowShape<RequestContext, RequestContext> outboundA = b.add(Flow.of(RequestContext.class)
+                .map(rc -> rc.addRequestHeader(RawHeader.create("keyOutA", "valOutA"))));
+            final FlowShape<RequestContext, RequestContext> outboundC = b.add(Flow.of(RequestContext.class)
+                .map(rc -> rc.addResponseHeader(RawHeader.create("keyOutC", "valOutC"))));
+
+            final BidiShape<RequestContext, RequestContext> inboundOutboundB =
+                b.add(package$.MODULE$.AbortableBidiFlow(authorization.asScala()).abortable().asJava());
```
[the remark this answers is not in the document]
sebady: I'll fix this by adding another static method.

build.sbt
```diff
@@ -28,7 +28,7 @@ val par = {
 
 concurrentRestrictions in Global := Seq(Tags.limitAll(par))
 
-lazy val `squbs-pipeline` = project
+lazy val `squbs-pipeline` = project.enablePlugins(de.johoop.testngplugin.TestNGPlugin)
```
anilgursel: I am not sure if it is because of this line, but IntelliJ does not recognize `squbs-pipeline` as a project anymore.  We should probably enableplugin in `build.sbt`?

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
```diff
@@ -18,25 +18,47 @@ package org.squbs.pipeline
...
+  def instance = this
+}
+case object ClientPipeline extends PipelineType {
+  def instance = this
+}
 
 case class Context(name: String, pipelineType: PipelineType)
 
-trait PipelineFlowFactory {
+sealed trait PipeLineFlowFactoryType
```
anilgursel: Typo in "PipeLine", capital "L".

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
```diff
@@ -18,25 +18,47 @@ package org.squbs.pipeline
...
 import com.typesafe.config.ConfigObject
 
-import scala.annotation.tailrec
 import scala.util.Try
 
 sealed trait PipelineType
-case object ServerPipeline extends PipelineType
-case object ClientPipeline extends PipelineType
+case object ServerPipeline extends PipelineType {
+  def instance = this
```
anilgursel: Can you please clarify why `instance` is needed?

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
```diff
@@ -46,36 +68,30 @@ class PipelineExtensionImpl(flowFactoryMap: Map[String, PipelineFlowFactory],
...
+      val flowFactory = flowFactoryMap
+        .getOrElse(name, throw new IllegalArgumentException(s"Invalid pipeline name $name"))
 
-    @tailrec
-    def connectFlows(currentFlow: PipelineFlow, flowList: List[PipelineFlow]): PipelineFlow = {
-
-      flowList match {
-        case Nil => currentFlow
-        case head :: tail => connectFlows(currentFlow atop head, tail)
+      flowFactory match {
```
anilgursel: IMO, the logic is getting rather more complex here.  Do we really need `PipelineFlowFactoryType`?  I would rather prefer the Java one extend the scala one or vice versa..  For instance, the following works:

```scala
/**
  * Java API
  */
trait AbstractPipelineFlowFactory  {

  def create(context: Context, system: ActorSystem):
    javadsl.BidiFlow[RequestContext, RequestContext, RequestContext, RequestContext, NotUsed]
}

trait PipelineFlowFactory extends AbstractPipelineFlowFactory {

  def create(context: Context)(implicit system: ActorSystem):
  scaladsl.BidiFlow[RequestContext, RequestContext, RequestContext, RequestContext, NotUsed]

  override def create(context: Context, system: ActorSystem):
  javadsl.BidiFlow[RequestContext, RequestContext, RequestContext, RequestContext, NotUsed] = create(context)(system).asJava
}
```

Since it is a `trait` you may need to drop `Abstract` from the name and put it into package `jap {}i` within this file.

docs/pipeline.md
```diff
@@ -161,6 +186,54 @@ class DummyAbortableBidiFlow extends PipelineFlowFactory {
...
+                .map(rc -> rc.addRequestHeader(RawHeader.create("keyInA", "valInA"))));
+            final FlowShape<RequestContext, RequestContext> inboundC = b.add(Flow.of(RequestContext.class)
+                .map(rc -> rc.addRequestHeader(RawHeader.create("keyInC", "valInC"))));
+            final FlowShape<RequestContext, RequestContext> outboundA = b.add(Flow.of(RequestContext.class)
+                .map(rc -> rc.addRequestHeader(RawHeader.create("keyOutA", "valOutA"))));
+            final FlowShape<RequestContext, RequestContext> outboundC = b.add(Flow.of(RequestContext.class)
+                .map(rc -> rc.addResponseHeader(RawHeader.create("keyOutC", "valOutC"))));
+
+            final BidiShape<RequestContext, RequestContext> inboundOutboundB =
+                b.add(Pipeline.abortable(authorization));
```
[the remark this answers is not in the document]
anilgursel: I understand your point.  But, I find it more readable without the static call.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
anilgursel: When I tried on my IDE, it was all good.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
sebady: Java needs this to gain a reference to ServerPipeline/ClientPipeline singleton.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
sebady: Good eye.  let me fix that.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
```diff
@@ -18,25 +18,47 @@ package org.squbs.pipeline
...
   def create(context: Context)(implicit system: ActorSystem):
-  BidiFlow[RequestContext, RequestContext, RequestContext, RequestContext, NotUsed]
+  scaladsl.BidiFlow[RequestContext, RequestContext, RequestContext, RequestContext, NotUsed]
+}
+
+/**
+  * Java API
+  */
+object Pipeline {
+  def abortable(flow: javadsl.BidiFlow[RequestContext, RequestContext, RequestContext, RequestContext, NotUsed]):
```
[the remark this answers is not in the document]
sebady: The current implementation, I believe more closely matches the scala API. Which I think is preferable here.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
sebady: Since the factory here is just an interface that a client implements, I don't see a lot of value in having the Scala one extend the Java specific one.   If there was some implementation logic, I can see the benefit of having one delegate or extend the other, but that's not the case with the PipelineFlowFactory.  I think we also considered creating a japi package here and prefer the simpler approach here with AbstractPipelineFlowFactory (which follows the pattern in other places in squbs)

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
```diff
@@ -46,36 +68,30 @@ class PipelineExtensionImpl(flowFactoryMap: Map[String, PipelineFlowFactory],
...
-    @tailrec
-    def connectFlows(currentFlow: PipelineFlow, flowList: List[PipelineFlow]): PipelineFlow = {
-
-      flowList match {
-        case Nil => currentFlow
-        case head :: tail => connectFlows(currentFlow atop head, tail)
+      flowFactory match {
+        case factory: PipelineFlowFactory => factory.create(context)
+        case factory: AbstractPipelineFlowFactory => factory.create(context, system).asScala
+        case factory => throw new IllegalArgumentException(s"Unsupported flow factory type ${factory.getClass}")
```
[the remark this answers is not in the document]
sebady: That's a good improvement.

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
```diff
@@ -16,11 +16,42 @@
...
+import scala.collection.JavaConverters._
+import scala.compat.java8.OptionConverters._
 import scala.util.{Success, Try}
 
+object RequestContext {
+
+  /**
+    * Java API. For creating instance of RequestContext
+    */
+  def create(request: jmHttpRequest, httpPipeliningOrder: JLong): RequestContext =
```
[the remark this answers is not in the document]
sebady: I was hoping this shorter convention for common java types would start becoming familiar in the codebase :-) .   We could also simply use Long here if we thing the primitive long that Java clients would get it acceptable.

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
```diff
@@ -34,7 +65,7 @@ case class RequestContext(request: HttpRequest,
   Java API
    */
   def withAttributes(attributes: java.util.List[(String, Any)]): RequestContext = {
-    ++(attributes: _*)
+    ++(attributes.asScala: _*)
```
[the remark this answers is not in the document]
sebady: I can add `withAttributes` but should we keep `++` for backwards compatibility?

sebady:
The java tests that are added here for PipelineFlow mimic the scala specs we already had.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
anilgursel: I do not think Java code ever needs that.  That's used in `getFlow` which is only called by squbs, not apps.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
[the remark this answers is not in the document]
anilgursel: I am not strong on this one; however, `Pipeline.abortable(...)` referencing back to `Pipeline` does not look quite simple to me.  I am not sure how it better matches with Scala API though.  I would think without `Pipeline.` it looks slightly closer.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
anilgursel: We have both `japi.` and `Abstract...`.  So, consistency wise there is no difference.  I prefer `japi` for a few reasons:

- Our code gets simpler (no need for extra base type, and case match).
- We never know if there will be logic in there at some point in the future.
- We handle it only single way, so there is less potential of "works for Scala but not works for Java".
- Everywhere else we share.

docs/pipeline.md
```diff
@@ -116,6 +118,27 @@ class DummyBidiFlow extends PipelineFlowFactory {
...
+            final FlowShape<RequestContext, RequestContext> inbound = b.add(Flow.of(RequestContext.class)
+              .map(rc -> rc.addRequestHeader(RawHeader.create("DummyRequest", "ReqValue"))));
+            final FlowShape<RequestContext, RequestContext> outbound = b.add(Flow.of(RequestContext.class)
+              .map(rc -> rc.addResponseHeader(RawHeader.create("DummyResponse", "ResValue"))));
+
+            return BidiShape.fromFlows(inbound, outbound);
+        }));
+    }
+}
+```
```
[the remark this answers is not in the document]
sebady: I will adjust this code block in pipeline.md

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
sebady: getFlow is still used from java tests though and requires a Context that has reference to either ServerPipeline or ClientPipeline.  For example in testAbortableFlowWithAbort:
```
        final BidiFlow<RequestContext, RequestContext, RequestContext, RequestContext, NotUsed> pipelineFlow =
            pipeLineExtension.getFlow(new Tuple2<>(Option.apply("dummyFlow1"), Option.apply(true)),
                new Context("dummy", ServerPipeline.instance())).get().asJava();
```

ClientPipeline.instance() is not used currently and I can remove the `def instance=this` addition there.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
sebady: Let me look at this.

sebady: @akara, @anilgursel I have added an additional commit in this pr that adds a japi package and refactors the existing java APIs appropriately.   Please take a look and let me know if you prefer this code organization with the added japi and I will squash the commits here.

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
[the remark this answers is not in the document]
anilgursel: Makes sense.

build.sbt
anilgursel: Please make sure IntelliJ happy after a refresh.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
anilgursel: I suggest to remote `.instance()` as there is no actual external API.  To use if from Java tests, you can simply use `ServerPipeline$.MODULE$`.

squbs-pipeline/src/main/scala/org/squbs/pipeline/PipelineExtension.scala
anilgursel: If you are keeping the two base factories independent, there is no need for a separate package.  It does not add any value, IMO.

My suggestion is to make one of the factories extend the other one and eliminate `case/match` and extra base factory type (because of the reasons I mentioned in an earlier comment).  With that `AbstractPipelineFlowFactory` becomes a `trait`that's why I was suggesting `japi.PipelineFlowFactory`.

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
```diff
@@ -16,36 +16,61 @@
...
+
+}
+
 case class RequestContext(request: HttpRequest,
                           httpPipeliningOrder: Long, // TODO Come up with a better val name
                           response: Option[Try[HttpResponse]] = None,
                           attributes: Map[String, Any] = Map.empty) {
 
-  def ++(attributes: (String, Any)*): RequestContext = {
+  def withResponse(response: Option[Try[HttpResponse]]): RequestContext = {
```
anilgursel: this should be taking `response: Try[HttpResponse]` instead of an `Option[Try[HttpResponse]]`, because we are adding a response here..  either an exception or a successful response.  `Option` gets redundant.

anilgursel: I wonder if it should be `withHttpResponse` or `withResponse`.

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
```diff
@@ -16,36 +16,61 @@
...
+    copy(response = response)
+  }
+
+  /*
+  Java API
+   */
+  def withResponse(response: Optional[Try[jmHttpResponse]]): RequestContext = {
```
anilgursel: Ditto

sebady:
@agursel I have updated the PR with the delta you suggested.  Still two commits on this PR here.  If it looks good please squash and merge.

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
sebady: Oops.  I forget to address these comments.

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
```diff
@@ -16,36 +16,61 @@
...
 
-  def ++(attributes: (String, Any)*): RequestContext = {
+  def withResponse(response: Try[HttpResponse]): RequestContext = {
+    copy(response = Option(response))
+  }
+
+  /*
+  Java API
+   */
+  def withJavaResponse(response: Try[jmHttpResponse]): RequestContext = {
```
anilgursel: Oops, I was about to merge, I did not realized this one before.  Let me check what we can do to make sure the name of the function does not have "Java".

akara: Very good point. Somehow I missed this one, too. I don't think method overload works well with generic types, in which case we could have two methods of `withResponse`.

I also don't see `withResponse` removed from the old code. How did we do that before? Was it just a `copy` call? So, what can be the solution to provide a uniform Scala or Java API. Here could be the list of withResponse signatures we can support:

* `withResponse(response: Try[HttpResponse]) // Scala API`
* `withResponse(response: HttpResponse) // Scala API`
* `withResponse(response: jmHttpResponse) // Java API`
* `withResponse(ex: Throwable) // Both Scala and Java`

The last one could also be called `withResponseError`.

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
```diff
@@ -16,60 +16,70 @@
...
+    apply(request.asInstanceOf[sm.HttpRequest], httpPipeliningOrder)
+
+}
+
+case class RequestContext private (request: sm.HttpRequest,
+                                   httpPipeliningOrder: Long,
+                                   response: Option[Try[sm.HttpResponse]] = None,
+                                   attributes: Map[String, Any] = Map.empty) {
+
+  def withResponse(response: Try[jm.HttpResponse]): RequestContext =
```
akara: Where is the Scala version of the same? I don't seem to see a Scala equivalent of this call.

anilgursel: Probably because we use `rc.copy(response = ...)` in the unit-tests, we have not caught this.  We should probably change the tests to use `withResponse` instead, e.g., https://github.com/paypal/squbs/blob/master/squbs-pipeline/src/test/scala/org/squbs/pipeline/AbortableBidiFlowSpec.scala#L39.

Issue #551, opened 2017-12-06
expose metrics from the Retry stage


sebady:
This is continuation of #511 to expose metrics in the Retry stage for monitoring purposed.

PR #615, opened 2018-03-14
 #551: Expose metrics from Retry stage

sebady:
Expose some metrics from the Retry stage for visibility.

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/docs/flow-retry.md
+++ b/docs/flow-retry.md
@@ -260,3 +260,37 @@ final BidiFlow<Pair<String, Context>,
                Pair<Try<String>, Context>,
                Pair<Try<String>, Context>, NotUsed> retry = Retry.create(settings);
 ```
+
+#### Metrics
+
+`Retry` supports Codahale counters for state and failure/success rate of elements passing through it.  Metrics can be enabled by calling `withMetrics` with a name and an actor system.
+
+The following counters are provided:
+
+  * Count of all retry elements ("<name>.retry-count")
+  * Count of all failed elements ("<name>.failed-count")
+  * Count of all successful elements ("<name>.success-count")
+  * Current size of the retry registry
+  * Current size of the retry queue
+
+##### Scala
+
+```scala
+val settings = RetrySettings[String, String, Context](max = 3)
+  .withMetrics("myRetry") // Takes implicit ActorSystem
+
+val retry = Retry(settings)
+```
+
+##### Java
+
+```java
+final RetrySettings settings =
+        RetrySettings.<String, String, Context>create(3)
+                .withMetrics("myRetry", system)
+
+final BidiFlow<Pair<String, Context>,
+               Pair<String, Context>,
+               Pair<Try<String>, Context>,
+               Pair<Try<String>, Context>, NotUsed> retry = Retry.create(settings);
+```
--- a/squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
+++ b/squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
@@ -16,16 +16,22 @@
 
 package org.squbs.streams
 
+import java.lang.management.ManagementFactory
 import java.lang.{Boolean => JBoolean}
+import java.util.concurrent.atomic.AtomicInteger
 import java.util.function.{Function => JFunction}
 
+import javax.management.{MXBean, ObjectName}
 import akka.NotUsed
+import akka.actor.ActorSystem
 import akka.http.org.squbs.util.JavaConverters
 import akka.japi.Pair
 import akka.stream.Attributes.InputBuffer
 import akka.stream._
 import akka.stream.scaladsl.{BidiFlow, Flow}
 import akka.stream.stage._
+import com.codahale.metrics.MetricRegistry
+import org.squbs.metrics.MetricsExtension
 
 import scala.collection.mutable
 import scala.concurrent.duration.{Duration, FiniteDuration, _}
@@ -53,14 +59,7 @@ object Retry {
     */
   def apply[In, Out, Context](retrySettings: RetrySettings[In, Out, Context]):
   BidiFlow[(In, Context), (In, Context), (Try[Out], Context), (Try[Out], Context), NotUsed] =
-    BidiFlow.fromGraph(new Retry(
-      max = retrySettings.max,
-      uniqueIdMapper = retrySettings.uniqueIdMapper,
-      failureDecider = retrySettings.failureDecider,
-      delay = retrySettings.delay,
-      exponentialBackoffFactor = retrySettings.exponentialBackoffFactor,
-      maxDelay = retrySettings.maxDelay,
-      maxWaitingRetries = retrySettings.maxWaitingRetries))
+    BidiFlow.fromGraph(new Retry(retrySettings))
 
   /**
     * Java API
@@ -110,27 +109,16 @@ object Retry {
   *           Out2         +------+        In2
   * }}}
   *
-  * @param max maximum number of retry attempts on any failing [[Try]]'s
-  * @param uniqueIdMapper function that maps a [[Context]] to a unique value per element
-  * @param failureDecider function that gets called to determine if an element passed by the joined [[Flow]] is a
-  *                       failure
-  * @param delay the delay duration to wait between each retry.  Defaults to 0 nanos (no delay)
-  * @param exponentialBackoffFactor The exponential backoff factor that the delay duration will be increased on each
-  *                                 retry
-  * @param maxDelay The maximum retry delay duration during retry backoff
+  * @param settings the settings to configure Retry with
   * @tparam In the type of elements pulled from the upstream along with the [[Context]]
   * @tparam Out the type of the elements that are pushed by the joined [[Flow]] along with the [[Context]].
   *             This then gets wrapped with a [[Try]] and pushed downstream with a [[Context]]
   * @tparam Context the type of the context that is carried around along with the elements.
   */
-final class Retry[In, Out, Context] private[streams](max: Int,
-                                                     uniqueIdMapper: Option[Context => Any] = None,
-                                                     failureDecider: Option[Try[Out] => Boolean] = None,
-                                                     delay: FiniteDuration = Duration.Zero,
-                                                     exponentialBackoffFactor: Double = 1.0,
-                                                     maxDelay: FiniteDuration = Duration.Zero,
-                                                     maxWaitingRetries: Option[Int] = None)
+final class Retry[In, Out, Context] private[streams](settings: RetrySettings[In, Out, Context])
   extends GraphStage[BidiShape[(In, Context), (In, Context), (Try[Out], Context), (Try[Out], Context)]] {
+
+  import settings._
 
   private val in1 = Inlet[(In, Context)]("RetryBidi.in1")
   private val out1 = Outlet[(In, Context)]("RetryBidi.out1")
@@ -139,6 +127,7 @@ final class Retry[In, Out, Context] private[streams](max: Int,
   private val delayAsNanos = delay.toNanos
   private val precisionAsNanos = 10.milliseconds.toNanos // the linux timer precision
   private val timerName = "RetryStageTimer"
+  private val nameSuffix = new AtomicInteger(-1)
   override val shape = BidiShape(in1, out1, in2, out2)
 
   require(max > 0, "maximum retry count must be positive")
@@ -182,6 +171,20 @@ final class Retry[In, Out, Context] private[streams](max: Int,
         case Some(InputBuffer(_, max)) => max
       }
     }
+    private val matNumber = nameSuffix.incrementAndGet
+    private val uniqueRetryName = if (matNumber == 0) name else s"$name-$matNumber"
+    val retryMetrics = metricRegistry.map(registry =>
+      RetryMetricsImpl(
+        max,
+        delay,
+        exponentialBackoffFactor,
+        maxDelay,
+        retryQMaxSize,
+        uniqueRetryName,
+        registry,
+        () => retryQ.size,
+        () => retryRegistry.size)
+    ).getOrElse(NoOpRetryMetrics)
 
     implicit val elementPriority: Ordering[(Long, Any)] = Ordering.by[(Long, Any), Long](e => e._1).reverse
     private val retryQ: mutable.PriorityQueue[(Long, Any)] = mutable.PriorityQueue.empty
@@ -252,6 +255,7 @@ final class Retry[In, Out, Context] private[streams](max: Int,
             if(hasBackpressuredIn1) pull(in1)
 
             if (isAvailable(out2)) {
+              retryMetrics.markFailure()
               push(out2, (elem, context))
               completeStageIfFinished()
             } else {
@@ -259,13 +263,14 @@ final class Retry[In, Out, Context] private[streams](max: Int,
               log.error("out2 is not available for push.  Dropping exhausted element")
             }
           } else {
+            retryMetrics.markRetry()
             val retryTime = System.nanoTime + delayTime(incrementAndGetRetryCount(key))
             val shouldRetryBeforeTheHeadOfQueue = retryQ.headOption.exists(_._1 - retryTime >= precisionAsNanos)
             if (shouldRetryBeforeTheHeadOfQueue) cancelTimer(timerName) // Because, the next retry time just changed
 
             retryQ.enqueue((retryTime, key))
 
-            if(isAvailable(out1)) {
+            if (isAvailable(out1)) {
               if (isHeadReady) {
                 val (elem, ctx, _) = retryRegistry(retryQ.dequeue()._2)
                 push(out1, (elem, ctx))
@@ -281,6 +286,7 @@ final class Retry[In, Out, Context] private[streams](max: Int,
             pull(in2)
           }
         } else {
+          retryMetrics.markSuccess()
           retryRegistry -= key
 
           // If a demand from out1 was not propagated to in1 because of retryQ size earlier
@@ -345,7 +351,7 @@ final class Retry[In, Out, Context] private[streams](max: Int,
 }
 
 /**
-  * A Retry Settings class for configuring a RetryBidi
+  * A Retry Settings class for configuring a Retry
   *
   * Retry functionality requires each element passing through is uniquely identifiable for retrying, so
   * it requires a [[Context]], of any type carried along with the flow's input and output element as a
@@ -372,13 +378,16 @@ final class Retry[In, Out, Context] private[streams](max: Int,
   * @tparam Context the type of the context that is carried along with the elements.
   * @return a [[RetrySettings]] with specified values
   */
-case class RetrySettings[In, Out, Context] private[streams](max: Int,
-                                                            uniqueIdMapper: Option[Context => Any] = None,
-                                                            failureDecider: Option[Try[Out] => Boolean] = None,
-                                                            delay: FiniteDuration = Duration.Zero,
-                                                            exponentialBackoffFactor: Double = 0.0,
-                                                            maxDelay: FiniteDuration = Duration.Zero,
-                                                            maxWaitingRetries: Option[Int] = None) {
+case class RetrySettings[In, Out, Context] private[streams](
+  max: Int,
+  uniqueIdMapper: Option[Context => Any] = None,
+  failureDecider: Option[Try[Out] => Boolean] = None,
+  delay: FiniteDuration = Duration.Zero,
+  exponentialBackoffFactor: Double = 0.0,
+  maxDelay: FiniteDuration = Duration.Zero,
+  maxWaitingRetries: Option[Int] = None,
+  name: String = "Retry",
+  metricRegistry: Option[MetricRegistry] = None) {
 
   def withUniqueIdMapper(uniqueIdMapper: Context => Any): RetrySettings[In, Out, Context] =
     copy(uniqueIdMapper = Some(uniqueIdMapper))
@@ -396,6 +405,9 @@ case class RetrySettings[In, Out, Context] private[streams](max: Int,
     copy(maxDelay = maxDelay)
 
   def withMaxWaitingRetries(max: Int): RetrySettings[In, Out, Context] = copy(maxWaitingRetries = Some(max))
+
+  def withMetrics(name: String)(implicit system: ActorSystem): RetrySettings[In, Out, Context] =
+    copy(name = name, metricRegistry = Some(MetricsExtension.get(system).metrics))
 
   /**
     * Java API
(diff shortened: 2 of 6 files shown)


squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -274,6 +292,7 @@ final class Retry[In, Out, Context] private[streams](max: Int,
             pull(in2)
           }
         } else {
+          if (retryRegistry.contains(key)) retryState.markSuccessfulRetry()
```
akara: The `retryRegistry` contains all elements coming into `in1`. Even a first time success would get into this condition. So it would be incorrect to do a `markSuccessfulRetry`, right? Or am I missing a condition that really makes sure this element has at least one retry?

squbs-ext/src/main/scala/org/squbs/streams/RetryState.scala
```diff
@@ -0,0 +1,174 @@
...
+  val delay: FiniteDuration
+  val maxDelay: FiniteDuration
+  val exponentialBackoffFactor: Double
+  val maxBufferSize: Int
+  val metricRegistry: MetricRegistry
+
+  //publish these constants
+  val RetryCount = s"$name.retry-count"
+  val FailedCount = s"$name.failed-count" // count of exhausted retries
+  val SuccessfulRetry = s"$name.successful-count"
```
akara: Based on my comments above and the string name, the semantic is correct. The variable and method name is not. The variable should be `SuccessCount` and not `SuccessfulRetry`.

squbs-ext/src/main/scala/org/squbs/streams/RetryState.scala
```diff
@@ -0,0 +1,174 @@
...
+
+  /**
+    * Mark a failing element that all retries were exhausted through Retry stage.
+    */
+  final def markAllFailed(): Unit = metricRegistry.meter(FailedCount).mark()
+
+  /**
+    * Mark a Success element passing trough Retry stage.
+    */
+  final def markSuccessfulRetry(): Unit = metricRegistry.meter(SuccessfulRetry).mark()
```
akara: Ditto here. This should be named `markSuccess`.

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -430,3 +473,88 @@ object RetrySettings {
...
+  def getName: String
+  def getMaxRetries: Int
+  def getDelay: String
+  def getMaxDelay: String
+  def getExponentialBackoffFactor: Double
+  def getMaxBufferSize: Int
+}
+
+private case class RetrySettingsMXBeanImpl(
+  name: String,
```
anilgursel: Instead of passing each one individually, I would just pass `RetrySettings` here.

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -430,3 +473,88 @@ object RetrySettings {
...
+}
+
+case class RetryMetrics private[streams](
+  name: String = "Retry",
+  maxRetries: Int,
+  delay: FiniteDuration = Duration.Zero,
+  maxDelay: FiniteDuration = Duration.Zero,
+  exponentialBackoffFactor: Double = 0.0,
+  maxBufferSize: Int,
+  metricRegistry: MetricRegistry = new MetricRegistry()) {
```
anilgursel: Other than the registry, all other area already part of `RetrySettingsMXBean`, why do we need to repeat here?

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -430,3 +473,88 @@ object RetrySettings {
...
+  val SuccessCount = s"$name.success-count"
+
+  /**
+    * The provided [[MetricRegistry]] will be used to register metrics
+    *
+    * @param metricRegistry the registry to use for codahale metrics
+    */
+  def withMetricRegistry(metricRegistry: MetricRegistry): RetryMetrics =
+    RetryMetrics(
+      name,
```
anilgursel: It is a case class so, can do `... = copy(metricRegistry = metricRegistry)`

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -430,3 +473,88 @@ object RetrySettings {
...
+      delay,
+      maxDelay,
+      exponentialBackoffFactor,
+      maxBufferSize,
+      metricRegistry)
+
+  /**
+    * Mark a failure element in Retry stage.
+    */
+  final def markFailure(): Unit = metricRegistry.meter(RetryCount).mark()
```
anilgursel: Would `markRetry` be a better name here?  and `markAllFailed ` gets renamed to `markFailure`?

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -430,3 +473,88 @@ object RetrySettings {
...
+  final def markSuccess(): Unit = metricRegistry.meter(SuccessCount).mark()
+
+  private val mBeanServer = ManagementFactory.getPlatformMBeanServer
+  private val beanName = new ObjectName(
+    s"org.squbs.configuration:type=squbs.retry.settings,name=${ObjectName.quote(name)}")
+
+  if (!mBeanServer.isRegistered(beanName))
+    mBeanServer.registerMBean(
+      RetrySettingsMXBeanImpl(
+        name,
```
anilgursel: I think just need to pass the `RetrySettings` instance.

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -206,6 +216,35 @@ final class Retry[In, Out, Context] private[streams](max: Int,
       // need to be scheduled anyway.  So, by not canceling here, we prevent a possible "cancel & schedule".
     }
 
+    private val matName = name + "-" + nameSuffix.getAndIncrement().toString
+    private val retryMetrics = RetryMetrics(matName, max, delay, maxDelay, exponentialBackoffFactor, retryQMaxSize)
+      .withMetricRegistry(MetricsExtension(system).metrics)
+
+    @MXBean
```
anilgursel: If possible, I would move most of the metrics code to the end (where other metrics beans are located) so that they do not get in the middle of the main logic.   Makes the reading of the code easier that way, IMO.

squbs-ext/src/test/scala/org/squbs/streams/RetryJMXSpec.scala
```diff
@@ -0,0 +1,121 @@
...
+import scala.util.{Failure, Success}
+
+class RetryJMXSpec extends TestKit(ActorSystem("RetryJMXSpec")) with FlatSpecLike with Matchers {
+
+  implicit val materializer = ActorMaterializer()
+  val failure = Failure(new Exception("failed"))
+
+  it should "use default name of Retry for jmx settings" in {
+    new RetryMetrics(maxRetries = 1, maxBufferSize = 16)
+    assertJmxSettings("Retry", "Name", "Retry")
```
anilgursel: The test may read better if the test is written like `jmxValue("Retry", "Name") shouldEqual "Retry"`.

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -71,9 +77,9 @@ object Retry {
     * @tparam Context the type of the context that is carried along with the elements.
     * @return a [[BidiFlow]] with Retry functionality
     */
-  def create[In, Out, Context](max: Integer):
+  def create[In, Out, Context](max: Integer, system: ActorSystem):
```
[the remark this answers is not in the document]
sebady: `RetrySettings` does not currently have an actorSystem.  We could add it there and remove it from here and the apply Method above.

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -206,6 +216,35 @@ final class Retry[In, Out, Context] private[streams](max: Int,
       // need to be scheduled anyway.  So, by not canceling here, we prevent a possible "cancel & schedule".
     }
 
+    private val matName = name + "-" + nameSuffix.getAndIncrement().toString
```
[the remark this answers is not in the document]
sebady: I can change this to "uniqueuRetryName"

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -206,6 +216,35 @@ final class Retry[In, Out, Context] private[streams](max: Int,
...
+    private val retryMetrics = RetryMetrics(matName, max, delay, maxDelay, exponentialBackoffFactor, retryQMaxSize)
```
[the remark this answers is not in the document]
sebady: RetryMetrics is responsible for establishing the RetrySettingsMXBean and registering it.  We could have the Retry directly do this instead and split out the metrics responsibilities here from RetrySettingsMXBean.

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
sebady: I will move the trait out here.

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -206,6 +216,35 @@ final class Retry[In, Out, Context] private[streams](max: Int,
...
+
+      override def getRegistrySize: Int = registry.size
+    }
+
+    private val mBeanServer = ManagementFactory.getPlatformMBeanServer
+    private val beanName = new ObjectName(
+      s"org.squbs.configuration:type=squbs.retry.state,name=${ObjectName.quote(matName)}")
+
+    if (!mBeanServer.isRegistered(beanName))
+      mBeanServer.registerMBean(RetryStateMXBeanImpl(matName, retryQ, retryRegistry), beanName)
```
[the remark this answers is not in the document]
sebady: We could make name default to a more unique value like `"Retry-" + System.identityHashCode"`  If we really want name uniqueness here even if the client of Retry has not specified a name.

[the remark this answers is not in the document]
anilgursel: I think the important thing is to make sure the user can relate the JMX name to the actual code.  With counters and hashcode, etc, it is very hard.  I think, for now, let's go with what's out there and come back to it as an improvement.

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
anilgursel: `new MetricRegistry()` shouldn't this be defaulting to the MetricExtension's registry?

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -153,6 +163,12 @@ final class Retry[In, Out, Context] private[streams](max: Int,
     }
   }
 
+  val retryMetrics = metricRegistry.map(registry => RetryMetricsImpl(name, registry)).getOrElse(NoOpRetryMetrics)
```
akara: Nice!

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -183,6 +199,20 @@ final class Retry[In, Out, Context] private[streams](max: Int,
       }
     }
 
+    if (metricRegistry.isDefined) {
```
akara: Hmm, we don't usually test these in such a way...

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -183,6 +199,20 @@ final class Retry[In, Out, Context] private[streams](max: Int,
...
+          RetrySettingsMXBeanImpl(
+            name,
+            max,
+            delay.toString,
+            maxDelay.toString,
+            exponentialBackoffFactor,
+            retryQMaxSize),
+          settingsBeanName)
+      else log.warning(s"JMX Bean $settingsBeanName is already registered.")
+    }
```
akara: Can I suggest to do this in `RetryMetricsImpl` instead? You don't need another `if` clause for this. I mean the 

```scala
if (metricRegistry.isDefined) {
```

Or even any permutation of it like `metricRegistry.foreach {`

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -206,6 +236,30 @@ final class Retry[In, Out, Context] private[streams](max: Int,
...
+
+      override def getRegistrySize: Int = registry.size
+    }
+
+    private val beanName = new ObjectName(
+      s"org.squbs.configuration:type=squbs.retry.state,name=${ObjectName.quote(name)}")
+
+    if (!mBeanServer.isRegistered(beanName))
+      mBeanServer.registerMBean(RetryStateMXBeanImpl(name, retryQ, retryRegistry), beanName)
+
```
akara: Wondering whether this should also go inside `RetryMetricsImpl` or some class that defines we indeed have a metric.

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -397,6 +457,9 @@ case class RetrySettings[In, Out, Context] private[streams](max: Int,
 
   def withMaxWaitingRetries(max: Int): RetrySettings[In, Out, Context] = copy(maxWaitingRetries = Some(max))
 
+  def withMetrics(name: String, system: ActorSystem): RetrySettings[In, Out, Context] =
+    copy(name = name, metricRegistry = Some(MetricsExtension.get(system).metrics))
```
akara: It is preferred for the Scala API to use the signature...

```scala
withMetrics(name: String)(implicit system: ActorSystem): RetrySettings
```

Please see whether this still works with the same Java API. In such cases, the Scala usage is more optimized.

squbs-ext/src/test/scala/org/squbs/streams/RetryJMXSpec.scala
```diff
@@ -0,0 +1,100 @@
...
+        (failure, ctx)
+      case (elem, ctx) =>
+        first = true
+        (Success(elem), ctx)
+    }
+
+    var context = 0L
+    Source("a" :: "b" :: "c" :: Nil)
+      .map { s => context += 1; (s, context) }
+      .via(Retry(RetrySettings[String, String, Long](1).withMetrics("Retry", system)).join(bottom))
```
akara: With the comment above, it will just be `.withMetrics("Retry")`.

docs/flow-retry.md
```diff
@@ -259,4 +259,36 @@ final BidiFlow<Pair<String, Context>,
...
+#### Metrics
+
+`Retry` supports Codahale counters for elements passing through each Retry.  To enable metrics collection you will have to
+ provide a name (used to uniquely identify each metrics for each Retry instance) as well as actor system in `RetrySettings`.
+
+ The following counters are provided:
+
+  * Count of all retry elements ("<name>.retry-count")
+  * Count of all failed elements ("<name>.failed-count")
+  * Count of all successful elements ("<name>.success-count")
```
anilgursel: How about retryQ size and registry size?

docs/flow-retry.md
```diff
@@ -259,4 +259,36 @@ final BidiFlow<Pair<String, Context>,
                Pair<String, Context>,
                Pair<Try<String>, Context>,
                Pair<Try<String>, Context>, NotUsed> retry = Retry.create(settings);
-```
\ No newline at end of file
+```
+
+#### Metrics
+
+`Retry` supports Codahale counters for elements passing through each Retry.  To enable metrics collection you will have to
```
anilgursel: Something like below ?

> `Retry` provides Codahale metrics for the state and failure/success rate of elements passing through.

docs/flow-retry.md
```diff
@@ -259,4 +259,36 @@ final BidiFlow<Pair<String, Context>,
...
+ provide a name (used to uniquely identify each metrics for each Retry instance) as well as actor system in `RetrySettings`.
+
```
anilgursel: > you will have to ...

Metrics can be enabled by calling `withMetrics` with a name and actor system.

docs/flow-retry.md
```diff
@@ -259,4 +259,36 @@ final BidiFlow<Pair<String, Context>,
...
+ The following counters are provided:
+
+  * Count of all retry elements ("<name>.retry-count")
+  * Count of all failed elements ("<name>.failed-count")
+  * Count of all successful elements ("<name>.success-count")
+
+##### Scala
+
+```scala
+val settings = RetrySettings[String, String, Context](max = 3).withMetrics("myRetry", system)
```
anilgursel: Please drop the system

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -183,6 +172,21 @@ final class Retry[In, Out, Context] private[streams](max: Int,
       }
     }
 
+    private val uniqueRetryName = if (nameSuffix.getAndIncrement().equals(0)) name else s"$name-${nameSuffix}"
```
akara: We get too much Java in here 😄 Perhaps this is a bit more readable and a tad more efficient:

```scala
private val thisNameSuffix = nameSuffix.getAndIncrement()
private val uniqueRetryName = if (thisNameSuffix == 0) name else s"$name-$thisNameSuffix"`
```

Basically, we do not dereference `nameSuffix` twice.

docs/flow-retry.md
akara: Oops, that last part is true only for Java. In Scala the actor system is supplied implicitly.

squbs-ext/src/test/scala/org/squbs/streams/RetryJMXSpec.scala
```diff
@@ -0,0 +1,121 @@
...
+      .map { case (s, ctx) => s }
+      .runWith(Sink.seq)
+
+    awaitAssert {
+      metricsJmxValue("Retry.retry-count", "Count").value shouldBe 2 // a, b
+      metricsJmxValue("Retry.failed-count", "Count").value shouldBe 1 // a
+      metricsJmxValue("Retry.success-count", "Count").value shouldBe 2 // b + c
+    }
+  }
+*/
```
akara: Is this test commented out because it does not pass?

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -182,6 +171,19 @@ final class Retry[In, Out, Context] private[streams](max: Int,
         case Some(InputBuffer(_, max)) => max
       }
     }
+    private val uniqueRetryName = if (nameSuffix.incrementAndGet() == 0) name else s"$name-${nameSuffix.get.toString}"
```
akara: This is much better. Suggestion is also not to dereference the value behind `nameSuffix` twice. But since this is not in the critical path for performance, I don't want to be fuzzy about it.

docs/flow-retry.md
```diff
@@ -259,4 +259,38 @@ final BidiFlow<Pair<String, Context>,
...
+#### Metrics
+
+`Retry` supports Codahale counters for elements passing through each Retry.  Metrics can be enabled by calling `withMetrics` with a name and an actor system (on Java only).
+
+ The following counters are provided:
+
+  * Count of all retry elements ("<name>.retry-count")
+  * Count of all failed elements ("<name>.failed-count")
+  * Count of all successful elements ("<name>.success-count")
+
```
akara: Space line in the list?

docs/flow-retry.md
```diff
@@ -259,4 +259,38 @@ final BidiFlow<Pair<String, Context>,
...
+  * current size of the retry registry
+  * current size of the retry queue
```
akara: Capitalize "current" please. => "Current" on both lines. This is just to keep it consistent with the rest of the list above.

docs/flow-retry.md
```diff
@@ -259,4 +259,38 @@ final BidiFlow<Pair<String, Context>,
...
+
+##### Scala
+
+```scala
+val settings = RetrySettings[String, String, Context](max = 3).withMetrics("myRetry")
```
akara: You may want to state above this line this expects an `implicit val system: ActorSystem`. Can be done as a comment, too.

docs/flow-retry.md
```diff
@@ -259,4 +259,39 @@ final BidiFlow<Pair<String, Context>,
                Pair<String, Context>,
                Pair<Try<String>, Context>,
                Pair<Try<String>, Context>, NotUsed> retry = Retry.create(settings);
-```
\ No newline at end of file
+```
+
+#### Metrics
+
+`Retry` supports Codahale counters for elements passing through each Retry.  Metrics can be enabled by calling `withMetrics` with a name and an actor system (on Java only).
```
anilgursel: I believe my earlier comment was lost.  I do not think the first sentence is clear enough.  

How about something like:

> `Retry` provides Codahale metrics for the state and failure/success rate of elements passing through.

anilgursel: @akara I see you suggested "(on Java only)", but I see it confusing and unnecessary.  Whether explicitly or implicitly, we are passing actor system.  So, not sure if the paranthesis is necessarily needed.

docs/flow-retry.md
```diff
@@ -259,4 +259,39 @@ final BidiFlow<Pair<String, Context>,
...
+  * Count of all retry elements ("<name>.retry-count")
+  * Count of all failed elements ("<name>.failed-count")
+  * Count of all successful elements ("<name>.success-count")
+  * Current size of the retry registry
+  * Current size of the retry queue
+
+##### Scala
+
+```scala
+implicit val system: ActorSystem = ... // assume we have an implicit ActorSystem from somewhere
```
anilgursel: I find this one unnecessary for the docs.  More importantly, this line is quite confusing.

docs/flow-retry.md
akara: I'm fine either way. We just need to make it clear. You can change the parentheses to "`ActorSystem` is passed implicitly in Scala" or whatever. I'm not extremely particular as long as it communicates clearly.

docs/flow-retry.md
akara: Please suggest how to communicate that `.withMetrics` needs an `implicit` `ActorSystem`. Anything is better than nothing.

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -182,6 +171,20 @@ final class Retry[In, Out, Context] private[streams](max: Int,
         case Some(InputBuffer(_, max)) => max
       }
     }
+    private val matNumber = nameSuffix.incrementAndGet
+    private val uniqueRetryName = if (matNumber == 0) name else s"$name-${nameSuffix.get.toString}"
```
akara: Close, but still suffers the potential discrepancy in value as we access `nameSuffix` twice. Please do this:

```scala
private val matNumber = nameSuffix.incrementAndGet
provate val uniqueRetryName -= if (matNumber == 0) name else s"$name-$matNumber"
```

You don't need to do the `toString` in this case as any substitution is already converted to a String.

docs/flow-retry.md
anilgursel: If we would like to communicate that, I think 

```scala
val settings =
  RetrySettings[String, String, Context](max = 3)
    .withMetrics("myRetry") // Takes implicit ActorSystem
```

might be a safer way to communicate rather than creating an actor system or mentioning with an implicit above.  I see we make a similar mistake in http client docs where we create an actor system https://squbs.readthedocs.io/en/latest/httpclient/

docs/flow-retry.md
anilgursel: I was suggesting to say "Metrics can be enabled by calling `withMetrics` with a name and an actor system".  This should suffice, IMO.  In both Scala and Java, a parameter is passed.  Explicitly or implicitly is a detail, IMO (except in the code examples).

docs/flow-retry.md
sebady: ok. I have tweaked this section of the docs in PR. Please take a look.

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
sebady: I'll fix this.

docs/flow-retry.md
```diff
@@ -259,4 +259,39 @@ final BidiFlow<Pair<String, Context>,
...
+  * Count of all retry elements ("<name>.retry-count")
+  * Count of all failed elements ("<name>.failed-count")
+  * Count of all successful elements ("<name>.success-count")
+  * Current size of the retry registry
+  * Current size of the retry queue
+
+##### Scala
+
+```scala
+
```
anilgursel: If you are pushing another commit, please remove this empty line.

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -245,27 +248,30 @@ final class Retry[In, Out, Context] private[streams](max: Int,
         val key = uniqueId(context)
 
         if (isFailure(elem)) {
+
```
anilgursel: And empty line here as well, not sure on purpose.  Only if you are pushing another commit.

Issue #623, opened 2018-03-19
Retry pauses the stream forever once back pressure applied


anilgursel:
Once back pressure applied to upstream, it never gets out of that state.  Once downstream demands more elements, it should be able to resume.  For instance, the below test is failing.  In addition, `Retry` should not affect the throughput of the system when all is fine.  It should back pressure only when failures are happening.

```scala
it should "not back pressure if downstream demands more and retryQ is not growing" in {

    val delayActor = system.actorOf(Props[DelayActor])
    import akka.pattern.ask
    implicit val askTimeout = akka.util.Timeout(10 seconds)

    val delayFlow =
      Flow[(Long, Long)]
        .mapAsyncUnordered(100)(elem => (delayActor ? elem).mapTo[(Long, Long)])
        .map { case (elem, ctx) => (Success(elem), ctx) }

    val retry = Retry[Long, Long, Long](2)
    val result = Source(1L to 500)
      .map(x => (x, x))
      .via(retry.join(delayFlow))
      .runWith(Sink.seq)

    result map {
      _ should contain theSameElementsAs (1L to 500 map(elem => (Success(elem), elem)))
    }
  }
```

PR #624, opened 2018-03-19
#623 Retry should be able to resume after a backpressure

anilgursel:

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/docs/flow-retry.md
+++ b/docs/flow-retry.md
@@ -171,6 +171,7 @@ final BidiFlow<Pair<String, Context>,
                Pair<Try<String>, Context>, NotUsed> retry = Retry.create(settings);
 
 ```
+
 ##### Exponential backoff
 An optional exponential backoff factor can also be specified to increase the delay duration on each subsequent retry attempt (up to a maximum delay duration).
 In the following examples, the first failure of any element will be retried after a delay of 200ms, and then any second attempt will be retried after 800ms.
@@ -228,6 +229,32 @@ final RetrySettings settings =
                 .withExponentialBackoff(2.0)
                 withMaxDelay(Duration.create("400 millis"));
 
+final BidiFlow<Pair<String, Context>,
+               Pair<String, Context>,
+               Pair<Try<String>, Context>,
+               Pair<Try<String>, Context>, NotUsed> retry = Retry.create(settings);
+```
+
+##### Configuring the threshold for backpressure
+
+If downstream is having trouble, `Retry` starts back pressuring when the elements waiting to be retired reaches to a certain threshold.  By default, the threashold is equal to the internal buffer size of `Retry` Akka Stream Grapstage (please see [Akka Stream Attributes](https://doc.akka.io/docs/akka/current/stream/stream-composition.html#attributes)).  The threshold can be made independent of internal buffer size by calling `withMaxWaitingRetries`:
+
+
+##### Scala
+
+```scala
+val settings = RetrySettings[String, String, Context](max = 3).withMaxWaitingRetries(50)
+
+val retry = Retry(settings)
+```
+
+##### Java
+
+```java
+final RetrySettings settings =
+        RetrySettings.<String, String, Context>create(3)
+                .withMaxWaitingRetries(50)
+        
 final BidiFlow<Pair<String, Context>,
                Pair<String, Context>,
                Pair<Try<String>, Context>,
--- a/docs/flow-retry.md
+++ b/docs/flow-retry.md
@@ -237,7 +237,7 @@ final BidiFlow<Pair<String, Context>,
 
 ##### Configuring the threshold for backpressure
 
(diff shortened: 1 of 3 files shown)


squbs-ext/src/test/scala/org/squbs/streams/RetrySpec.scala
```diff
@@ -481,4 +506,36 @@ class RetrySpec extends TestKit(ActorSystem("RetryBidiSpec")) with AsyncFlatSpec
...
+    val delayActor = system.actorOf(Props[DelayActor])
+    import akka.pattern.ask
+    implicit val askTimeout = akka.util.Timeout(10 seconds)
+
+    val delayFlow =
+      Flow[(Long, Long)]
+        .mapAsyncUnordered(100)(elem => (delayActor ? elem).mapTo[(Long, Long)])
+        .map { case (elem, ctx) => (Success(elem), ctx) }
+
+    val retry = Retry[Long, Long, Long](2).withAttributes(inputBuffer(initial = 1, max = 1))
```
akara: I presume you have other tests that cover larger buffer sizes.

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
```diff
@@ -382,6 +395,8 @@ case class RetrySettings[In, Out, Context] private[streams](max: Int,
   def withMaxDelay(maxDelay: FiniteDuration): RetrySettings[In, Out, Context] =
     copy(maxDelay = maxDelay)
 
+  def withMaxWaitingRetries(max: Int): RetrySettings[In, Out, Context] = copy(maxWaitingRetries = Some(max))
```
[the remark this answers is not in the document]
anilgursel: I was debating about that..  But, I think it is better to have the doc.  I will make change.

squbs-ext/src/test/scala/org/squbs/streams/RetrySpec.scala
anilgursel: It backpressures when the queue size grows to internal buffer size.  So, if it is not back pressuring with buffer size one, it won't with larger buffer size.  So, for this specific test, I am not sure what the benefit would be.

Do you just want to make sure things work fine with large buffers?

squbs-ext/src/test/scala/org/squbs/streams/RetrySpec.scala
```diff
@@ -481,4 +506,36 @@ class RetrySpec extends TestKit(ActorSystem("RetryBidiSpec")) with AsyncFlatSpec
...
+    val stream = Source(1L to 500)
+      .map(x => (x, x))
+      .via(retry.join(delayFlow))
+      .runWith(Sink.seq)
+
+    // It should not impact the throughput when the stream is happy.  If it were to back pressure, it would take
+    // 500 seconds, because internal buffer size is 1 and each takes 1 second.  The current setup with concurrency
+    // level 100, it should take 500 / 100 = 5 seconds.  Setting it to 10 seconds to prevent Travis CI problems.
+    val result = Await.result(stream, 10 seconds)
+    result should contain theSameElementsAs (1L to 500 map(elem => (Success(elem), elem)))
```
sebady: This assumption for travis I believe could make this test brittle in the future.

squbs-ext/src/main/scala/org/squbs/streams/Retry.scala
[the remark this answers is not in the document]
sebady: Isn't the value for "internalBufferSize" (I.e. InputBuffer attribute) sufficient ?  Why the additional settings here for maxRetriesWaiting? The overlap here I think might be confusing or we need to document this well. I.e when one or the other should be used.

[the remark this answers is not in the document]
anilgursel: There may be scenarios to decouple it from internal buffer size.

squbs-ext/src/test/scala/org/squbs/streams/RetrySpec.scala
anilgursel: I gave an extra 5 seconds.  Hopefully, Travis would be within that range.  If we see problems, we can revise.

docs/flow-retry.md
```diff
@@ -233,3 +234,29 @@ final BidiFlow<Pair<String, Context>,
                Pair<Try<String>, Context>,
                Pair<Try<String>, Context>, NotUsed> retry = Retry.create(settings);
 ```
+
+##### Configuring the threshold for backpressure
+
+If downstream is having trouble, `Retry` starts back pressuring when the elements waiting to be retired reaches to a certain threshold.  By default, the threashold is equal to the internal buffer size of `Retry` Akka Stream Grapstage (please see [Akka Stream Attributes](https://doc.akka.io/docs/akka/current/stream/stream-composition.html#attributes)).  The threshold can be made independent of internal buffer size by calling `withMaxWaitingRetries`:
```
sebady: Some typos in this paragraph,  `retried`, `threshold` and `Graphstage`.  I would also consider using the term "the joined flow" instead of "downstream".  Also the term "trouble" is also a little vague here.  I assume we mean is "slow to respond or is buffering elements?"

anilgursel: Just pushed a new commit to address your comment.  Please check.

Issue #622, opened 2018-03-19
Make deprecations and feature flag more explicit in compile output


fmasion:
When compiling there's a lot of deprecation and feature warnings 
suggesting to re-run with -feature and -deprecation

There's event more in test cases but tests are very verbose so it's not easy to clearly see them

running sbt commands 
- clean
- test:compile
will show 179 WARNINGS without any detail about which file/line are concerned

Moreover there are many more compiler flag's that could help be aligned with the high quality requirements of this project.
These flags however could be activated later on

I suggest to create a [compileOptions.sbt](https://gist.github.com/fmasion/8e50c98e7e272fd3e48abf6db700b2eb)  (see gist) file with all these flags and activate them on your choice progressively.

PR #625, opened 2018-03-20
 #622 Make deprecations and feature flag more explicit in compile output and addressing some boring warnings

fmasion:

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/squbs-ext/src/main/scala/org/squbs/resolver/Resolver.scala
+++ b/squbs-ext/src/main/scala/org/squbs/resolver/Resolver.scala
@@ -64,11 +64,13 @@ class ResolverRegistryExtension(system: ExtendedActorSystem) extends Extension w
     *
     * @param resolver The resolver implementation
     */
-  def register[T: ClassTag](resolver: Resolver[T]): Unit =
+  def register[T: ClassTag](resolver: Resolver[T]): Unit = {
+    import scala.language.existentials
     resolvers.find { case (_, oldResolver) => oldResolver.name == resolver.name } match {
       case None => resolvers = (classTag[T].runtimeClass, resolver) :: resolvers
       case Some(_) => logger.warn(s"Endpoint Resolver: ${resolver.name} already registered, skipped!")
     }
+  }
 
   /**
     * Scala API to register a resolver on the fly.
--- a/squbs-ext/src/main/scala/org/squbs/util/ConfigUtil.scala
+++ b/squbs-ext/src/main/scala/org/squbs/util/ConfigUtil.scala
@@ -123,16 +123,12 @@ object ConfigUtil extends LazyLogging {
     }
 
 
-    def getOptionalConfigList(path: String): Option[Seq[Config]] = {
-      import scala.language.existentials
-      val list =
-        try {
-          Some(underlying.getConfigList(path))
+    def getOptionalConfigList(path: String): Option[Seq[Config]] = try {
+          Some(underlying.getConfigList(path).asScala)
         } catch {
           case e: ConfigException.Missing => None
         }
-      list map (_.asScala.toSeq)
-    }
+
 
     def getOptionalDuration(path: String): Option[FiniteDuration] = {
       import scala.concurrent.duration._
--- a/squbs-ext/src/main/scala/org/squbs/util/ConfigUtil.scala
+++ b/squbs-ext/src/main/scala/org/squbs/util/ConfigUtil.scala
@@ -124,6 +124,7 @@ object ConfigUtil extends LazyLogging {
 
 
     def getOptionalConfigList(path: String): Option[Seq[Config]] = {
+      import scala.language.existentials
       val list =
         try {
           Some(underlying.getConfigList(path))
--- a/compileOptions.sbt
+++ b/compileOptions.sbt
@@ -1,0 +1,59 @@
+
+val javaVersion = "1.8"
+
+scalacOptions in ThisBuild ++= Seq(    // Copied from https://tpolecat.github.io/2017/04/25/scalac-flags.html
+  "-target:jvm-" + javaVersion,        //
+  "-deprecation",                      // Emit warning and location for usages of deprecated APIs.
+  "-encoding", "utf-8",                // Specify character encoding used by source files.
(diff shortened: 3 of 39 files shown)


anilgursel:
I am not sure if `30.seconds`or `30 seconds` (with `import scala.language.postfixOps`) preferable.

fmasion:
I agree 30 seconds is elegant but 30.seconds is less difficult for the compiler to figure out
Postfix operator notation is unsafe and should not be used.
[Scala style recommendation](https://docs.scala-lang.org/style/method-invocation.html#postfix-notation)

sebady:
Since the scala style guide recommendation is clear here, I would vote we move away from Postfix. I think for test code though postfix improves readability in some cases and I would prefer to optionally continue to use it there.

fmasion:
@sebady tests, by nature, are DLSs :) 
the magic of DSL is built upon implicits, postfix and magnet pattern

Don't conclude that postfix is evil
It's only a matter of balance between readability and safety

IMHO 30.seconds is still very readable ;)

akara:
Sorry for chiming in late. Agree with `30.seconds`.

squbs-ext/src/main/scala/org/squbs/resolver/Resolver.scala
```diff
@@ -64,11 +64,13 @@ class ResolverRegistryExtension(system: ExtendedActorSystem) extends Extension w
     *
     * @param resolver The resolver implementation
     */
-  def register[T: ClassTag](resolver: Resolver[T]): Unit =
+  def register[T: ClassTag](resolver: Resolver[T]): Unit = {
+    import scala.language.existentials
```
akara: I'm wondering about this liberal `import scala.language.existentials`. I'm not clear we're even using existential types in the code block below.

squbs-ext/src/main/scala/org/squbs/util/ConfigUtil.scala
```diff
@@ -124,6 +124,7 @@ object ConfigUtil extends LazyLogging {
 
 
     def getOptionalConfigList(path: String): Option[Seq[Config]] = {
+      import scala.language.existentials
```
akara: Ditto on these.

squbs-ext/src/main/scala/org/squbs/resolver/Resolver.scala
fmasion: Honestly I'm not quite sure how to resolve this.
Here's the compiler output with -feature flag activated
```
[info] Compiling 1 Scala source to /Users/Hit/Dev/squbs/squbs-ext/target/scala-2.12/classes ...
[warn] /Users/Hit/Dev/squbs/squbs-ext/src/main/scala/org/squbs/resolver/Resolver.scala:70:32: inferred existential type (Class[_$1], org.squbs.resolver.Resolver[T]) forSome { type _$1 }, which cannot be expressed by wildcards,  should be enabled
[warn] by making the implicit value scala.language.existentials visible.
[warn] This can be achieved by adding the import clause 'import scala.language.existentials'
[warn] or by setting the compiler option -language:existentials.
[warn] See the Scaladoc for value scala.language.existentials for a discussion
[warn] why the feature should be explicitly enabled.
[warn]       case None => resolvers = (classTag[T].runtimeClass, resolver) :: resolvers
[warn]                                ^
```

I presume it's a warning because the code uses scala.reflect.classTag that makes the type only known at runtime ?
If you know how to fix this I'll do the change
I tried to reduce the "scope" of the import inside the function 
this import is only a feature flag to tell the compiler "I know what I do don't warn me"

squbs-ext/src/main/scala/org/squbs/util/ConfigUtil.scala
fmasion: here the output for this one :
```
[warn] /Users/Hit/Dev/squbs/squbs-ext/src/main/scala/org/squbs/util/ConfigUtil.scala:129:9: inferred existential type Option[java.util.List[?0]] forSome { type ?0 <: com.typesafe.config.Config }, which cannot be expressed by wildcards,  should be enabled
[warn] by making the implicit value scala.language.existentials visible.
[warn] This can be achieved by adding the import clause 'import scala.language.existentials'
[warn] or by setting the compiler option -language:existentials.
[warn] See the Scaladoc for value scala.language.existentials for a discussion
[warn] why the feature should be explicitly enabled.
[warn]         try {
[warn]         ^
[warn] one warning found
```
the signature of ```underlying.getConfigList(path)``` is 
```List<? extends Config> getConfigList(String path); ``` in java 
that scala compiler translates to :
java.util.List[?0]] forSome { type ?0 <: com.typesafe.config.Config} 
subtypes of Config

We can do remove the warning here by explicitly casting the java result like this :
```
    def getOptionalConfigList(path: String): Option[Seq[Config]] = {
      val list =
        try {
          Some(underlying.getConfigList(path).asScala.map(x=> x:Config)) //CAST HERE
        } catch {
          case e: ConfigException.Missing => None
        }
      list map (_.toSeq)
    }
```
Or by importing the 
```
import scala.language.existentials
```

What do you prefer ?

Comment (akara) [reply squbs-ext/src/main/scala/org/squbs/util/ConfigUtil.scala]: I'd be fine if it is a single cast with `.asInstanceOf` but not if we need to cast record by record. Please see whether that works. If not I'll accept it as is.

File: squbs-ext/src/main/scala/org/squbs/resolver/Resolver.scala (hunk 1 above)
Comment (akara) [reply squbs-ext/src/main/scala/org/squbs/resolver/Resolver.scala]: Seems like a pretty common issue with `Class[_]`. Lets take it as in your current PR then.

File: squbs-ext/src/main/scala/org/squbs/util/ConfigUtil.scala (hunk 2 above)
Comment (fmasion) [reply squbs-ext/src/main/scala/org/squbs/util/ConfigUtil.scala]: even better :
``` 
    def getOptionalConfigList(path: String): Option[Seq[Config]] = try {
          Some(underlying.getConfigList(path).asScala)
        } catch {
          case e: ConfigException.Missing => None
        }
```
works without warning 
javaConverters may do the job ?

fmasion:
@akara updated and rebased

PR #646, opened 2018-04-18
Added Java API for accessing the system materializer. (fixes #637)

### LINKED ISSUE #637: Java API access to registered materializers in Unicomplex.
We need to add a couple of interfaces to make the access the Unicomplex extension, and especially the materializer easy for the Java users. Currently it is pretty much impossible to get to from Java.

akara:
Added Java API, test, and documentation updates to reflect this API.

- [X] Commits are squashed.
- [X] Tests added.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -113,9 +113,17 @@ A squbs materializer is nothing but an Akka Streams `Materializer` that is speci
 
    * a `Materializer` can be accessed from different locations through an Akka Extension as follows:
 
-   ```scala
-   implicit val mat = Unicomplex(system).materializer("default-materializer")
-   ```
+     **Scala**
+   
+     ```scala
+     implicit val mat = Unicomplex(system).materializer("default-materializer")
+     ```
+   
+     **Java**
+   
+     ```java
+     final Materializer mat = Unicomplex.get(system).materializer("default-materializer")
+     ```
 
    * a materializer can be referenced from a [squbs listener](#listeners).
    * The materializers in use by an application can be reported on JMX with the corresponding settings.
--- a/squbs-unicomplex/src/test/java/org/squbs/unicomplex/MaterializerTest.java
+++ b/squbs-unicomplex/src/test/java/org/squbs/unicomplex/MaterializerTest.java
@@ -1,0 +1,11 @@
+package org.squbs.unicomplex;
+
+import akka.actor.ActorSystem;
+import akka.stream.Materializer;
+
+public class MaterializerTest {
+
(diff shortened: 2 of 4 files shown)


docs/configuration.md
```diff
@@ -113,9 +113,17 @@ A squbs materializer is nothing but an Akka Streams `Materializer` that is speci
...
+     **Scala**
+   
+     ```scala
+     implicit val mat = Unicomplex(system).materializer("default-materializer")
+     ```
+   
+     **Java**
+   
+     ```java
+     Materializer mat = Unicomplex.get(system).materializer("default-materializer")
```
sebady: It would be ideal to have this be "final"

akara: Sure!

squbs-unicomplex/src/test/java/org/squbs/unicomplex/MaterializerTest.java
```diff
@@ -0,0 +1,11 @@
+package org.squbs.unicomplex;
+
+import akka.actor.ActorSystem;
+import akka.stream.Materializer;
+
+public class MaterializerTest {
+
+    public static Materializer getMaterializer(ActorSystem system, String name) {
```
anilgursel: What's this test asserting?

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/MaterializerSpec.scala
```diff
@@ -210,6 +210,18 @@ class MaterializerSpec extends TestKit(
     }
   }
 
+  it should "obtain the default materializer through the Java API" in {
```
anilgursel: I am not sure why we do not assert these directly on the Java test.

squbs-unicomplex/src/test/java/org/squbs/unicomplex/MaterializerTest.java
akara: This is not a test by itself. It is just being called from the test in Scala to prove the Java API returns the same object.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/MaterializerSpec.scala
akara: Because of the cost and overhead of creating a separate test in Java. Starting UnicomplexBoot from Java is non-trivial. We never tried to make that a Java API. That's why I avoid it altogether.
Otherwise we'll have another task this task will need to depend on.

Issue #515, opened 2017-09-21
Retry Bidi should have 95%+ code coverage


anilgursel:
It looks like some Java APIs and overflow strategies are not covered well.

anilgursel:
Follow up issue https://github.com/paypal/squbs/issues/649

PR #647, opened 2018-04-20
 #515 Improve Retry tests

sebady:
Change the java RetryTest to be a junit test
Also minor test additions to RetrySpec

- [x] Commits are squashed.
- [x] Tests added.


squbs-ext/src/test/java/org/squbs/streams/RetryTest.java
```diff
@@ -26,8 +26,8 @@
 import akka.stream.javadsl.Sink;
 import akka.stream.javadsl.Source;
 import akka.stream.testkit.javadsl.TestSink;
+import org.junit.Test;
```
akara: Please make sure the assertions are also JUnit assertions. The signature of these two between expect and actual is inverted. So your assert calls need to be changed.

akara:
Can you please let me know the test coverage you got after this change? Also, is this PR ready to be merged?

sebady:
The scoverage report shows 85% line coverage

anilgursel:
Comments for [this coverage report](https://github.com/paypal/squbs/files/1953019/Retry.scala.html.log) (remove the .log extension and open in a browser, apparently Github does not support html):

Line 140: There should be a corresponding test case in CircuitBreaker, you can take a look at there.

Line 170: Is there any scenario where it can really be `None`?  Given that we are anyway throwing an exception, we can instead just do a `get` on the `Option`.

Line 217: We can reproduce this scenario by causing retries with really long delays, while elements are still in registry, a stage on the upstream can fail.  However, I am not really sure about the logic here.  In any scenario we fail. Shouldn't we let let the messages be drained?

Line 229: This one might be tricky to reproduce consistently.  Thinking.. 🤔   I wonder if adding a buffer stage on the upstream, or making the internal buffer size high(more than the internal registry size) may help reproduce.

Line 240: After making sure there is no element in the registry, completing a stage in the joined flow should trigger this.  However, it needs to be tested that whether `onDowstreamFinish` of `out1` or `onUpstreamFinish` of `in2` is called, or indeterministic.

Line 255: Any scenario where we keep retrying and reaching to the registry size limit and ultimately causing back pressure, and resuming after a while should cover this line.

Line 263: This will never get called.  I suggest removing this in the next or next-to-next release.  interesting thing though, how come the comment line shows green 😄 

Line 269: When exponential back off is in place, and let's say 2nd element is failing second time with a really long delay, and 3rd (or later) element comes in and will be re-tried very soon..  such scenarios should cover this.

Line 275: Let's say, only a single element is going through the stream and it is no-delay and a failure happens.  It should cover this branch.  We should also be able to reproduce this with multiple elements and a large buffer in the joined flow (the buffer would make sure out1 is always available).

Line 293: Similar to 255.  

Line 300: Same as 263.

Line 310: While all elements are in registry, if an upstream stage completes, that should ultimately trigger this scenario.

Line 464: Don't we already have JMX tests?  I would assume those would cover it.

sebady:
Line 170: The only scenario where InputBuffer would not exist in the `inheritedAttributes` is probably user error.  So I can change this to a simpler `inheritedAttributes.mandatoryAttribute[InputBuffer].max)` 

Line 217: I think  the current implementation of `onUpstreamFailure` that either "propagates" the exception on `out1` or fails the stage seems like reasonable behavior for this stage.
Line 255 and 269:  Still trying to add test for this scenario.

Line 464: Should now be covered in latest commit on PR.

PR #650, opened 2018-05-08
Replace deprecated methods and classes

jiminhsieh:
Replace deprecated methods and classes. I will squash after reviewing.

- [ ] Commits are squashed.
- [ ] Tests added.

### DIFF
--- a/squbs-unicomplex/src/test/scala/org/squbs/unicomplex/dummysvcactor/DummySvcActor.scala
+++ b/squbs-unicomplex/src/test/scala/org/squbs/unicomplex/dummysvcactor/DummySvcActor.scala
@@ -38,7 +38,7 @@ class DummySvcActor extends Actor with WebContext with ActorLogging {
   import context.dispatcher
 
   def receive: Receive = {
-    case req@HttpRequest(_, Uri(_, _, Path("/dummysvcactor/ping"), _, _), _, _, _) =>
+    case req @ HttpRequest(_, Uri(_, _, Path("/dummysvcactor/ping"), _, _), _, _, _) =>
       log.debug("Received request " + req.uri)
       sender() ! HttpResponse(StatusCodes.OK, entity = "pong")
 
@@ -47,9 +47,9 @@ class DummySvcActor extends Actor with WebContext with ActorLogging {
       var chunkCount = 0L
       var byteCount = 0L
 
-      val future = req.entity.dataBytes.filter(_.length > 0).runForeach{ b =>
-          chunkCount += 1
-          byteCount += b.length
+      val future = req.entity.dataBytes.filter(_.length > 0).runForeach { b =>
+        chunkCount += 1
+        byteCount += b.length
       }
 
       val origSender = sender()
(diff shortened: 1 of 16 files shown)


squbs-unicomplex/src/test/scala/org/squbs/unicomplex/dummysvcactor/DummySvcActor.scala
```diff
@@ -52,7 +52,7 @@ class DummySvcActor extends Actor with WebContext with ActorLogging {
 
       val origSender = sender()
 
-      future onSuccess {
+      future onComplete  {
```
akara: Eh? Does `onComplete` not expect a `Try[T] => U`? Below here it just matches the `byteCount` to be use in the String. So `$byteCount` will be a `Try` and print weird messages to the response, right?

Did I miss anything?

akara:
Besides 1 comment, LGTM.

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/dummysvcactor/DummySvcActor.scala
jiminhsieh: Here is the underneath of `onSuccess`:
https://github.com/scala/scala/blob/v2.12.5/src/library/scala/concurrent/Future.scala#L116-L120
As you can see, it still calls `onComplete`.

And here is the warning from compiler:
```
[warn] /XXX/XXX/repository/github/squbs/squbs-unicomplex/src/test/scala/org/squbs/unicomplex/dummysvcactor/DummySvcActor.scala:55:14: method onSuccess in trait Future is deprecated (since 2.12.0): use `foreach` or `onComplete` instead (keep in mind that they take total rather than partial functions)
```

akara: I don't object the use of `onComplete`. I just meant the content of the closure needs to be adjusted accordingly:

```scala
future onComplete {
  case Success(byteCount) => origSender ! HttpResponse(StatusCodes.OK, entity = s"Received $chunkCount chunks and $byteCount bytes.")
  case Failure(e) => // Do what? Ignore? Send a HttpResponse with the error?
}
```

jiminhsieh: Oh, I got it. Thanks for your reminding!   
> Ignore? Send a HttpResponse with the error?

How about sending a HttpResponse? Like this: 
```scala
future onComplete {
  case Success(byteCount) => origSender ! HttpResponse(StatusCodes.OK, entity = s"Received $chunkCount chunks and $byteCount bytes.")
  case Failure(e) => HttpResponse(StatusCodes.InternalServerError, entity = s"Fail to process chunks: $e")
}
```
IMHO, I don't prefer `Ignore` although that's how `onSuccess` handles failure. If we want to use similar behavior as the original, I think we can log a warning or error message.

akara: Sounds great!

Issue #613, opened 2018-03-12
RequestContext should have withAttributes and removeAttributes


anilgursel:
`RequestContext` should support plural versions of attribute methods to prevent multiple object creations.

PR #655, opened 2018-05-11
 #613 Add withAttributes/removeAttributes methods

crankydillo:
See #613 for info.

- [x] Commits are squashed.
- [x] Tests added.

### DIFF
--- a/squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
+++ b/squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
@@ -20,11 +20,10 @@ import java.util.Optional
 
 import akka.http.javadsl.{model => jm}
 import akka.http.scaladsl.{model => sm}
-
 import scala.annotation.varargs
 import scala.collection.JavaConverters._
 import scala.compat.java8.OptionConverters._
-import scala.util.{Success, Try}
+import scala.util.Try
 
 object RequestContext {
 
@@ -50,7 +49,18 @@ case class RequestContext private (request: sm.HttpRequest,
   def withAttribute(name: String, value: Any): RequestContext =
     this.copy(attributes = this.attributes + (name -> value))
 
-  def removeAttribute(name: String): RequestContext = this.copy(attributes = this.attributes - name)
+  @varargs
+  def withAttributes(attributes: (String, Any)*): RequestContext = withAttributes(attributes.toMap)
+
+  /**
+    * Java API
+    */
+  def withAttributes(attributes: java.util.Map[String, Any]): RequestContext = withAttributes(attributes.asScala.toMap)
+
+  def removeAttribute(name: String): RequestContext = removeAttributes(name)
+
+  @varargs
+  def removeAttributes(names: String*): RequestContext = this.copy(attributes = this.attributes -- names)
 
   def withRequestHeader(header: jm.HttpHeader): RequestContext = withRequestHeaders(header)
 
@@ -82,4 +92,7 @@ case class RequestContext private (request: sm.HttpRequest,
     * Java API
     */
   def getResponse: Optional[Try[jm.HttpResponse]] = response.map(_.map(_.asInstanceOf[jm.HttpResponse])).asJava
+
+  private def withAttributes(attributes: Map[String, Any]): RequestContext =
+    this.copy(attributes = this.attributes ++ attributes)
 }
--- a/squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
+++ b/squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
@@ -49,7 +49,6 @@ case class RequestContext private (request: sm.HttpRequest,
   def withAttribute(name: String, value: Any): RequestContext =
     this.copy(attributes = this.attributes + (name -> value))
 
-  @varargs
   def withAttributes(attributes: (String, Any)*): RequestContext = withAttributes(attributes.toMap)
 
   /**
--- a/squbs-pipeline/src/test/java/org/squbs/pipeline/RequestContextTest.java
+++ b/squbs-pipeline/src/test/java/org/squbs/pipeline/RequestContextTest.java
@@ -19,7 +19,6 @@ package org.squbs.pipeline;
 import akka.http.javadsl.model.HttpRequest;
 import akka.http.javadsl.model.HttpResponse;
 import org.testng.annotations.Test;
-import scala.Tuple2;
 import scala.util.Success;
 import scala.util.Try;
 
@@ -28,8 +27,9 @@ import java.util.*;
 import static org.testng.Assert.assertEquals;
 
 public class RequestContextTest {
+
     @Test
-    public void testCreateWithAttributes() {
+    public void testCreateWithAttribute() {
         List<String> testL = Arrays.asList("1", "2", "3");
 
         RequestContext rc = RequestContext.create(HttpRequest.create(), 0)
@@ -40,6 +40,22 @@ public class RequestContextTest {
         assertEquals(rc.getAttribute("keyB"), Optional.of(testL));
         assertEquals(rc.getAttribute("keyC"), Optional.of("strC"));
         assertEquals(rc.getAttribute("keyD"), Optional.empty());
+    }
+
+    @Test
+    public void testCreateWithAttributes() {
+        Map<String, Object> attrs = new HashMap<>();
+        attrs.put("key1", "val1");
+        attrs.put("key2", 1);
+        attrs.put("key3", new Exception("Bad Val"));
+
+        RequestContext rc1 = RequestContext.create(HttpRequest.create(), 0);
+        RequestContext rc2 = rc1.withAttributes(attrs);
+
+        assertEquals(rc1.attributes().size(), 0);
+
+        assertEquals(rc2.attributes().size(), attrs.size());
+        attrs.forEach((key, value) -> assertEquals(rc2.getAttribute(key), Optional.of(value)));
     }
 
     @Test
--- a/squbs-pipeline/src/test/scala/org/squbs/pipeline/RequestContextSpec.scala
+++ b/squbs-pipeline/src/test/scala/org/squbs/pipeline/RequestContextSpec.scala
@@ -16,24 +16,56 @@
 
 package org.squbs.pipeline
 
-import java.util
-
 import akka.actor.ActorSystem
+import akka.http.javadsl.{model => jm}
 import akka.http.scaladsl.model.headers.RawHeader
 import akka.http.scaladsl.model.{HttpRequest, HttpResponse}
 import akka.testkit.TestKit
 import org.scalatest.OptionValues._
 import org.scalatest.{BeforeAndAfterAll, FlatSpecLike, Matchers}
-
 import scala.util.Try
 
-class RequestContextSpec extends TestKit(ActorSystem("RequestContextSpecSys")) with FlatSpecLike with Matchers with BeforeAndAfterAll {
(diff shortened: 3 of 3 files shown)


crankydillo:
I'm unsure why this last push, which was a squashed commit that removed an unused trait from the changed spec, failed.  The previous CI [run](https://travis-ci.org/paypal/squbs/builds/377526399) did not fail.

I tracked the failure to the following and am not sure if it's intermittent.

```
[info] - should share the circuit breaker state across materializations *** FAILED ***
[info]   14 was not less than or equal to 12 (ClientFlowCircuitBreakerSpec.scala:259)
```

Running the tests locally pass.

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
```diff
@@ -48,9 +48,22 @@ case class RequestContext private (request: sm.HttpRequest,
...
   def withAttribute(name: String, value: Any): RequestContext =
-    this.copy(attributes = this.attributes + (name -> value))
+    withAttributes(Map(name -> value))
+
+  // I know this isn't exactly like the withRequestHeaders varargs.  I could achieve that
+  // easily with tuples here, but the Java counterpart would have (Pair, Pair*), which IMO
+  // is way less familiar than a java.Map.  Also, I thought another vararg override for Scala
+  // was needless.
+  def withAttributes(attributes: Map[String, Any]): RequestContext =
+    this.copy(attributes = this.attributes ++ attributes)
```
sebady: Could the java counterpart to this use `JMap` and we could use vararg tuples here, just to remain consistent with withRequestHeaders etc.  I think consistency with `withRequestHeaders/withResponseHeaders` etc  is preferable to have "consistency" between java and scala clients (both using Map).

akara: Like the idea of using vararg tuples for the Scala API. Java API will probably need a Java Map. The alternative to Tuples is `Pair` which is awkward to use.

akara:
```scala
ctx.withAttributes(
  "foo" -> "bar",
  "foobar" -> "baz"
)
```

akara:
Shall we need doc updates for this? The doc updated marker is off.

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
crankydillo: k, tuples it is!

crankydillo:
I switched Scala to tuples.  RE docs, I thought about that.  I should have checked, but I just ASSumed you didn't document every method:)

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
```diff
@@ -48,9 +48,20 @@ case class RequestContext private (request: sm.HttpRequest,
...
 
   def withAttribute(name: String, value: Any): RequestContext =
-    this.copy(attributes = this.attributes + (name -> value))
+    withAttributes(name -> value)
+
+  @varargs
+  def withAttributes(attribute: (String, Any), attributes: (String, Any)*): RequestContext =
+    withAttributes(attributes.toMap + attribute)
+
+  def withAttributes(attributes: JMap[String, Any]): RequestContext =
```
anilgursel: I would prefer having the long `java.util.Map` here instead of `JMap`.  A Java user clicking on the method and seeing `JMap` may get confused.

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
```diff
@@ -48,9 +48,20 @@ case class RequestContext private (request: sm.HttpRequest,
     copy(response = Option(response.map(_.asInstanceOf[sm.HttpResponse])))
 
   def withAttribute(name: String, value: Any): RequestContext =
-    this.copy(attributes = this.attributes + (name -> value))
+    withAttributes(name -> value)
+
+  @varargs
+  def withAttributes(attribute: (String, Any), attributes: (String, Any)*): RequestContext =
```
anilgursel: This is inconsistent with the Java API.  Should we make this a `Map` as well?  Or make the Java one as a varargs?  For efficiency reasons, I would prefer both being a `Map`.

squbs-pipeline/src/test/java/org/squbs/pipeline/RequestContextTest.java
```diff
@@ -42,6 +43,24 @@ public void testCreateWithAttributes() {
...
+        Map<String, Object> attrs = new HashMap<>();
+        attrs.put("key1", "val1");
+        attrs.put("key2", 1);
+        attrs.put("key3", new Exception("Bad Val"));
+
+        RequestContext rc1 = RequestContext.create(HttpRequest.create(), 0);
+        RequestContext rc2 = rc1.withAttributes(attrs);
+        assertEquals(rc1.attributes().size(), 0);
+        assertEquals(rc2.attributes().size(), attrs.size());
+        for (Map.Entry<String, Object> attr : attrs.entrySet()) {
```
anilgursel: Would the following read better than the for loop?:

```java
assertTrue(rc2.attributes().forall(e -> e._2().equals(attrs.get(e._1()))));
```

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
sebady: I believe the previous comment was to be consistent with the other withXXX methods we should support tuples in the withAttributes scala API.  Since `Pairs` are not as java friendly to Java clients we decided to accept a JMap instead for the java API.  (I.e consistency  in API for Scala or Java users is more preferable than consistency between the Java and Scala APIs)

anilgursel: > ...  consistency in API for Scala or Java users is more preferable than consistency between the Java and Scala APIs

I agree.  Regarding consistency among Scala APIs, what exactly are you referring to?  If `withResponseHeaders` and `withRequestHeaders`, that would apply to Java as well, right?

sebady: Yes. `WithRequestsHeaders` and `withResponseHeaders` both accept vararg (javamodel) headers and withRequestHeader/withResponseHeader accept a single header.   Perhaps we can just stick with `'Pair` on java.

squbs-pipeline/src/test/java/org/squbs/pipeline/RequestContextTest.java
crankydillo: This works, but the `for` loop is a tad safer.  Replace your version's `rc2` with `rc1` and the test will still pass.  However, that would be caught by my `for` loop.  I could swith to Java's `forEach`.

Also, but do you typically use Scala methods like `forall` in the Java tests?

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
crankydillo: Will do.

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
```diff
@@ -47,10 +46,20 @@ case class RequestContext private (request: sm.HttpRequest,
   def withResponse(response: Try[jm.HttpResponse]): RequestContext =
     copy(response = Option(response.map(_.asInstanceOf[sm.HttpResponse])))
 
-  def withAttribute(name: String, value: Any): RequestContext =
-    this.copy(attributes = this.attributes + (name -> value))
+  def withAttribute(name: String, value: Any): RequestContext = withAttributes(name -> value)
+
+  @varargs
```
akara: The `@varargs` is only needed for Java APIs, so the compiler generates Java compatible varargs. But since this is a varargs of tuples which obviously are not available in Java, I doubt it makes sense to keep this API Java compatible.

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
```diff
@@ -47,10 +46,20 @@ case class RequestContext private (request: sm.HttpRequest,
...
+  def withAttributes(attribute: (String, Any), attributes: (String, Any)*): RequestContext =
+    withAttributes(attributes.toMap + attribute)
+
+  def withAttributes(attributes: java.util.Map[String, Any]): RequestContext =
+    withAttributes(attributes.asScala.toMap)
 
-  def removeAttribute(name: String): RequestContext = this.copy(attributes = this.attributes - name)
+  def removeAttribute(name: String): RequestContext = removeAttributes(name)
+
+  @varargs
```
akara: In contrast, the `@varargs` here makes total sense.

squbs-pipeline/src/test/scala/org/squbs/pipeline/RequestContextSpec.scala
```diff
@@ -16,24 +16,45 @@
...
 import org.scalatest.{BeforeAndAfterAll, FlatSpecLike, Matchers}
-
 import scala.util.Try
 
-class RequestContextSpec extends TestKit(ActorSystem("RequestContextSpecSys")) with FlatSpecLike with Matchers with BeforeAndAfterAll {
+class RequestContextSpec
+  extends TestKit(ActorSystem("RequestContextSpecSys"))
+    with FlatSpecLike
+    with Matchers
+    with BeforeAndAfterAll {
```
akara: LOL

squbs-pipeline/src/test/java/org/squbs/pipeline/RequestContextTest.java
anilgursel: Being a test, I think it is totally ok to use Scala `forall`.  I am ok with keeping the for, it just looks ugly 😄

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
```diff
@@ -48,9 +48,20 @@ case class RequestContext private (request: sm.HttpRequest,
     copy(response = Option(response.map(_.asInstanceOf[sm.HttpResponse])))
 
   def withAttribute(name: String, value: Any): RequestContext =
-    this.copy(attributes = this.attributes + (name -> value))
+    withAttributes(name -> value)
```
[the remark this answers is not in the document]
anilgursel: Yes 120.  I thought it would still be within 120.  If not, definitely keep it in the next line.

squbs-pipeline/src/test/java/org/squbs/pipeline/RequestContextTest.java
crankydillo: Switched to Java's Map#forEach:)

squbs-pipeline/src/main/scala/org/squbs/pipeline/RequestContext.scala
```diff
@@ -50,7 +49,18 @@ case class RequestContext private (request: sm.HttpRequest,
   def withAttribute(name: String, value: Any): RequestContext =
     this.copy(attributes = this.attributes + (name -> value))
 
-  def removeAttribute(name: String): RequestContext = this.copy(attributes = this.attributes - name)
+  @varargs
```
anilgursel: @akara had a comment earlier about this.  This is needed for Java.  Tuples would not look nice in java anyway.

crankydillo:
@anilgursel I addressed the vararg issue.

PR #658, opened 2018-05-15
 #638: LifecycleManaged should not rely on system

sebady:
Re-add the actorsystem as implicit in unicomplex LifecycleManaged

- [x] Commits are squashed.
- [ ] Tests added.


akara:
The feature is not widely used, but there are customers using it. Would be better to have a KillSwitch-like stage that can pause/resume stream.

PR #667, opened 2018-06-09
[WIP] Bump chronicle queue

jiminhsieh:
- [ ] Commits are squashed.
- [ ] Tests added.


.travis.yml
```diff
@@ -36,7 +36,7 @@ before_install:
   fi
 
 script:
-  - sbt -jvm-opts jvmopts clean coverage test
```
jiminhsieh: There is the supported arguments from sbt: 
https://github.com/sbt/sbt-launcher-package/blob/92067eb3086f45ed8397f6f89257d0533b4f263d/src/universal/bin/sbt-launch-lib.bash#L177-L199

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -201,15 +201,15 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = _
...
-          file.getName.endsWith(SUFFIX) && (fileIdParser.toLong(file) < releasedFileLong)
-        } foreach { file =>
-          logger.info("File released {} - {}", cycle.toString, file.getPath)
-          file.delete()
-        }
+      val releasedFileLong = fileIdParser.toLong(releasedFile)
+      for {
+        allFiles <- Option(persistDir.listFiles).toSeq
+        file <- allFiles
+        if (!(file.getName == Tailer) || !(file.getName == "directory-listing.cq4t"))
```
jiminhsieh: Chronicle Queue use `directory-listing.cq4t` to track the lowest and highest cycle number in the directory. So we don't touch it. 
https://github.com/OpenHFT/Chronicle-Queue/blob/chronicle-queue-4.16.3/docs/How_it_works.adoc#file-rolling

.travis.yml
jiminhsieh: If we want to test with a specific memory, I think it would be better to use `SBT_OPTS` before this line. Or we can use 
```scala
fork in test := true
``` 
at each sub-module.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
anilgursel: Shouldn't it be `&&` ?

```scala
if (file.getName != Tailer && file.getName != "directory-listing.cq4t")
```

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -201,15 +201,15 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = _
...
+        if (file.getName.endsWith(SUFFIX) && (fileIdParser.toLong(file) < releasedFileLong))
+      } yield {
+        logger.info("File released {} - {}", cycle.toString, file.getPath)
+        file.delete()
```
anilgursel: Just a personal opinion, I think it would be easier to read this function if written as below:

```scala
private def deleteOlderFiles(cycle: Int, releasedFile: File): Unit = {

  val releasedFileLong = fileIdParser.toLong(releasedFile)

  Option(persistDir.listFiles).toArray.flatten
    .filterNot(file => file.getName == Tailer || file.getName == directoryListing)
    .filter(file => file.getName.endsWith(SUFFIX) && fileIdParser.toLong(file) < releasedFileLong)
    .foreach { file =>
      logger.info("File released {} - {}", cycle.toString, file.getPath)
      file.delete()
    }
}
```

squbs-unicomplex/src/test/scala/org/squbs/unicomplex/RouteDefinitionSpec.scala
```diff
@@ -77,7 +77,7 @@ class TestRouteDefinition extends RouteDefinition {
       }
     } ~ path("illegalrequest") {
       get {
-        throw new IllegalRequestException(ErrorInfo("Test illegal request", "This is a test"), StatusCodes.Locked)
+        throw new IllegalRequestException(new ErrorInfo("Test illegal request", "This is a test"), StatusCodes.Locked)
```
anilgursel: It is a case class, do we need `new`?

jiminhsieh: At `10.1.2`, it changed to a normal class instead of a case class.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
jiminhsieh: If I didn't misunderstand, this snippet tried to ignore meta-data which will be `Tailer` or `directory-listing.cq4t`. If we use `&&` at this place, it means to ignore the file is `Tailer` and `directory-listing.cq4t` at the same time.

After seeing your suggestion, I will change to this:
```scala
if (file.getName != Tailer || file.getName != DirectoryListing.DIRECTORY_LISTING_FILE)
```

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
jiminhsieh: The reason I use for-comprehension are:
* My own opinion is it is clear. 
* From the technical perspective, it will use `withFilter`. AFAIK, this will produce fewer short-term objects.

@akara @sebady Do you have any thought about this?

crankydillo: As coded, this makes an attempt at deleting.  I'd either be clearer in the signature/doc that this is failure to delete files is ignored, use a different method, or deal with the boolean returned from `delete`.

jvmopts
```diff
@@ -1,4 +1,4 @@
--Xms256m
+-Xms1g
```
akara: A little worried this will use up more memory than needed and potentially kill the Travis build job. It is pretty unforgiving if we use even a bit more memory than the Travis build image provides.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
akara: Should we just not touch it? Or should we actually read the `directory-listing.cq4t` to determine the files outside the range.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
akara: I'd actually agree with @crankydillo on the removal of the `yield`, unless you make use of the result of the `yield`. Which brings me to another point. Should we check the return value of `file.delete` and log errors if the file is not deleted?

jvmopts
jiminhsieh: At container-based build, we can use memory up to 4G. It's not a huge problem. This is also good for low GC pause at production. [Here](https://developer.jboss.org/message/532524#532524) had the explanation.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
anilgursel: `if (file.getName != Tailer || file.getName != DirectoryListing.DIRECTORY_LISTING_FILE)` will always be `true`.  

| `file.getName` | `file.getName != Tailer` | `file.getName != DirectoryListing.DIRECTORY_LISTING_FILE` | Result |
| --------------- | ------------------------- | -------------- | --------------- |
| `Tailer` | false | true | true |
| `DirectoryListing.DIRECTORY_LISTING_FILE` | true | false | true |
| `something else` | true | true | true |

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
anilgursel: Given this is not on the main execution path, I would be less worried on object creation and prefer readability.  But, if the agreement is for comprehension being more readable in this scenario, no objections.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
jiminhsieh: Ah, I see. I was confused with the `filterNot` at test. Thanks for the correction and the table, @anilgursel!

anilgursel: Should we have a test for this?

jiminhsieh: I don't think we have the test for this situation, this logic error didn't fail at any test. I will try to add a test for this.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
jiminhsieh: Thanks for the reminder, @crankydillo!

How about check the boolean returned from `delete?`
```scala
for {
  allFiles <- Option(persistDir.listFiles).toSeq
  file <- allFiles
  if (file.getName != Tailer && file.getName != DirectoryListing.DIRECTORY_LISTING_FILE)
  if (file.getName.endsWith(SUFFIX) && (fileIdParser.toLong(file) < releasedFileLong))
  } {
      logger.info("File released {} - {}", cycle.toString, file.getPath)
      if (!file.delete()) {
          logger.error("Failed to DELETE {}", file.getPath)
    }
}
```

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
jiminhsieh: > should we actually read the `directory-listing.cq4t` to determine the files outside the range.

Could you elaborate more about this? 

After doing some research about `directory-listing.cq4t`, we may use it to for this line: 
https://github.com/paypal/squbs/blob/2662767dbc7103bf0e1451b4c6367f05969b3b9a/squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala#L200

docs/persistent-buffer.md
```diff
@@ -8,7 +8,7 @@ The following dependencies are required for Persistent Buffer to work:
 
 ```scala
 "org.squbs" %% "squbs-pattern" % squbsVersion,
-"net.openhft" % "chronicle-queue" % "4.6.70"
+"net.openhft" % "chronicle-queue" % "4.16.3"
```
anilgursel: It looks like 4.16.5 is the latest.

jvmopts
akara: I understand the production part :-) The problem is the 4GB is shared between our parallel builds. If one takes too much it can impact the test stability as a whole.

project/Versions.scala
```diff
@@ -15,8 +15,8 @@
  */
 
 object Versions {
-  val akkaV = "2.5.12"
-  val akkaHttpV = "10.1.1"
+  val akkaV = "2.5.13"
```
akara: Thanks for doing this!

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
akara: That's what I meant. Thanks for researching.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
akara: I like this check. Thanks!

docs/persistent-buffer.md
jiminhsieh: I noticed that too. But they didn't push to maven yet. I think we could wait for that.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -201,14 +201,14 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = _
...
-          logger.info("File released {} - {}", cycle.toString, file.getPath)
-          file.delete()
+      for {
+        allFiles <- Option(persistDir.listFiles).toSeq
+        file <- allFiles
+        if (file.getName.endsWith(SUFFIX) &&  fileIdParser.toLong(releasedFile) > (fileIdParser.toLong(file)))
+      } {
+        logger.info("File released {} - {}", cycle.toString, file.getPath)
+        if (!file.delete()) {
+          logger.error("Failed to DELETE {}", file.getPath)
```
akara: 👍

anilgursel: I do not see filtering of Tailer and `directory-listing.cq4t` in this commit.  Earlier commits had a check like below:

```scala
if (file.getName != Tailer && file.getName != "directory-listing.cq4t")
```

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -201,14 +201,14 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = _
...
-        files.filter { file =>
-          file.getName.endsWith(SUFFIX) && (fileIdParser.toLong(file) < releasedFileLong)
-        } foreach { file =>
-          logger.info("File released {} - {}", cycle.toString, file.getPath)
-          file.delete()
+      for {
+        allFiles <- Option(persistDir.listFiles).toSeq
+        file <- allFiles
+        if (file.getName.endsWith(SUFFIX) &&  fileIdParser.toLong(releasedFile) > (fileIdParser.toLong(file)))
+      } {
```
anilgursel: I do not see filtering of Tailer and `directory-listing.cq4t` in this commit.  Earlier commits had a check like below:

```scala
if (file.getName != Tailer && file.getName != "directory-listing.cq4t")
```

jiminhsieh: I found we don't need this line
```
if (file.getName != Tailer && file.getName != "directory-listing.cq4t")
```
since we only check files with `cq4` extension. That's why my logic error didn't cause any affect.

squbs-pattern/src/main/scala/org/squbs/pattern/stream/PersistentQueue.scala
```diff
@@ -201,14 +201,14 @@ class PersistentQueue[T](config: QueueConfig, onCommitCallback: Int => Unit = _
...
     private def deleteOlderFiles(cycle: Int, releasedFile: File): Unit = {
-      val files = Option(persistDir.listFiles).toArray.flatten.filterNot(_.getName == Tailer).toSeq
-      if (files.nonEmpty) {
-        val releasedFileLong = fileIdParser.toLong(releasedFile)
-        files.filter { file =>
-          file.getName.endsWith(SUFFIX) && (fileIdParser.toLong(file) < releasedFileLong)
-        } foreach { file =>
-          logger.info("File released {} - {}", cycle.toString, file.getPath)
-          file.delete()
+      for {
```
anilgursel: Just a note..  Originally this function was touched on this commit to add a check for directory listing file along with Trailer.  That change and some comments triggered some more changes in these lines.   But, then both the trailer and directory listing file check is removed as found unnecessary.  So, the only functional difference in this method became checking if delete operation was successful or not.  If not, log an error.

PR #678, opened 2018-07-27
 Fix some compilation issues with our retry doc.

crankydillo:
I also tried to do just a bit more to allow easier copying and tinkering.

- [x ] Description of the change added.
- [ x] Commits are squashed.
- [ ] Tests added.
- [ x] Documentation added/updated.


docs/flow-retry.md
```diff
@@ -94,7 +114,7 @@ class MyContext {
 
 final RetrySettings settings =
         RetrySettings.<String, String, MyContext>create(3)
-                .withUniqueIdMapper(context -> context.id());
+                .withUniqueIdMapper(ctx -> ctx.id());
```
anilgursel: Given `context` is not a long name and there is no line length problem, any reason preferring `ctx`?

docs/flow-retry.md
```diff
@@ -25,28 +25,48 @@ If all the retried attempts fail then the last failure for that request is emitt
 ##### Scala
 
 ```scala
-val retry = Retry[String, Long](max = 10)
-val flow = Flow[Try[String], Long].map(s => findAnEnglishWordThatStartWith(s))
+implicit val system = ActorSystem()
+implicit val mat = ActorMaterializer()
```
anilgursel: On one side, I understand completeness is important; but, the length of the doc and clutter needs to be minimized as well.  Otherwise, doc gets very lengthy and intimidating.  These are basic standard stuff, that's why I was avoiding them here.

docs/flow-retry.md
```diff
@@ -25,28 +25,48 @@ If all the retried attempts fail then the last failure for that request is emitt
...
 
-Source("a" :: "b" :: "c" :: Nil)
+val dict = List("albatross", "cow")
+def findAnEnglishWordThatStartWith(prefix: String): Try[String] =
```
anilgursel: Ditto..  But, mine are just opinions.. If the team agrees, I have no strong opinions.

docs/flow-retry.md
```diff
@@ -25,28 +25,48 @@ If all the retried attempts fail then the last failure for that request is emitt
...
+  .zipWithIndex
   .via(retry.join(flow))
-  .runWith(Sink.seq)
+  .runWith(Sink.foreach(println))
 ```
 
 ##### Java
 
 ```java
+ActorSystem system = ActorSystem.create();
```
anilgursel: `final` ?

docs/flow-retry.md
```diff
@@ -25,28 +25,48 @@ If all the retried attempts fail then the last failure for that request is emitt
...
+Materializer mat = ActorMaterializer.create(system);
```
anilgursel: Ditto.

docs/flow-retry.md
```diff
@@ -25,28 +25,48 @@ If all the retried attempts fail then the last failure for that request is emitt
...
 ```scala
-val retry = Retry[String, Long](max = 10)
-val flow = Flow[Try[String], Long].map(s => findAnEnglishWordThatStartWith(s))
+implicit val system = ActorSystem()
+implicit val mat = ActorMaterializer()
 
-Source("a" :: "b" :: "c" :: Nil)
+val dict = List("albatross", "cow")
+def findAnEnglishWordThatStartWith(prefix: String): Try[String] =
+  Try { dict.find(_.startsWith(prefix)).get }
```
anilgursel: Is this beyond 120? Also, I feel `Try(...)` would read better, but again must personal opinion.

crankydillo: I have always tended to use `{}` for methods that accept functions and `()` for 'normal' method calls.  I've been doing it for so long, I have forgotten all the arguments.  That said, I have started using '()' a bit more.  I attribute it to the Java8 programming I've been doing.

While they can be called interchangeably `def f1(i: Int)` and `def f2(i: => Int)` can behave differently.  For the former, it is guaranteed `f1(1/0)` will throw an arithmetic exception.  This is not true for `f2`.  Other than that, I like how the use of '{}' starts to drive home that everything is a method (e.g. scalatest's DSL are just normal method calls).

Regardless, I will remove all the extra stuff I added to fix the compilation errors and mess around with the `Retry` concept, so we all get more time to think on this.

docs/flow-retry.md
crankydillo: It is consistent with the newly added code above, where space was more limited.

docs/flow-retry.md
```diff
@@ -113,7 +116,7 @@ A function of type `Try[Out] => Boolean` can be provided to `RetrySettings` via
 a response with failing http status code is also considered a failure:
 
 ```scala
-val failureDecider = (tryResponse: Try[HttpResponse]) => tryResponse.isFailure || tryResponse.get().status().isFailure()
+val failureDecider = (tryResponse: Try[HttpResponse]) => tryResponse.isFailure || tryResponse.get.status.isFailure()
```
akara: `isFailure` should not have side-effects. Therefore no `()` at the end. Thanks for catching all this.

crankydillo: Done

Issue #683, opened 2018-09-19
Unicomplex should not eat exception when starting listeners


crankydillo:
I recently had to debug this stacktrace:
```
Exception in thread "main" java.util.concurrent.TimeoutException: Futures timed out after [10 seconds]
	at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:255)
	at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:183)
	at scala.concurrent.Await$.$anonfun$ready$1(package.scala:191)
	at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:53)
	at scala.concurrent.Await$.ready(package.scala:142)
	at org.squbs.unicomplex.UnicomplexBoot$.startServiceInfra(UnicomplexBoot.scala:471)
	at org.squbs.unicomplex.UnicomplexBoot.start(UnicomplexBoot.scala:598)
	at org.squbs.unicomplex.UnicomplexBoot.start(UnicomplexBoot.scala:547)
	at org.squbs.unicomplex.Bootstrap$.delayedEndpoint$org$squbs$unicomplex$Bootstrap$1(Bootstrap.scala:36)
	at org.squbs.unicomplex.Bootstrap$delayedInit$body.apply(Bootstrap.scala:30)
	at scala.Function0.apply$mcV$sp(Function0.scala:34)
	at scala.Function0.apply$mcV$sp$(Function0.scala:34)
	at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:12)
	at scala.App.$anonfun$main$1$adapted(App.scala:76)
	at scala.collection.immutable.List.foreach(List.scala:389)
	at scala.App.main(App.scala:76)
	at scala.App.main$(App.scala:74)
	at org.squbs.unicomplex.Bootstrap$.main(Bootstrap.scala:30)
	at org.squbs.unicomplex.Bootstrap.main(Bootstrap.scala)
```

Debugging led to a one-line `match` statement, which makes it slightly difficult to debug with my IDE.  Regardless, I shouldn't have to use a debugger to see that error message, so we need to do something better with [this exception](https://github.com/paypal/squbs/blob/8ad8cc9f10b7721754082c1bd16251808c1d3b0d/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/Unicomplex.scala#L475).

PR #684, opened 2018-09-21
 #683 Log an error when service startup fails.

crankydillo:

- [X] Commits are squashed.
- [ ] Tests added.

### DIFF
--- a/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/Unicomplex.scala
+++ b/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/Unicomplex.scala
@@ -472,7 +472,10 @@ class Unicomplex extends Actor with Stash with ActorLogging {
           },
             discardOld = false)
 
-        case Failure(t) => updateSystemState(checkInitState())
+        case Failure(t) => {
+          sender ! StartFailure(t)
+          updateSystemState(checkInitState())
+        }
       }
 
     case Started => // Bootstrap startup and extension init done
--- a/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
+++ b/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
@@ -465,8 +465,16 @@ object UnicomplexBoot extends LazyLogging {
     implicit val timeout = Timeout((boot.listeners.size * 10).seconds)
     val ackFutures =
       for ((listenerName, config) <- boot.listeners) yield {
-        Unicomplex(actorSystem).uniActor ? StartListener(listenerName, config)
+        val responseFuture = Unicomplex(actorSystem).uniActor ? StartListener(listenerName, config)
+        def log(t: Throwable): Unit = logger.error(s"Failed to start the listener, $listenerName.", t)
+        responseFuture.onComplete {
+          case Failure(t)               => log(t)
+          case Success(StartFailure(t)) => log(t)
+          case _                        =>
+        }
+        responseFuture
       }
+
     // Block for the web service to be started.
     Await.ready(Future.sequence(ackFutures), timeout.duration)
 
--- a/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
+++ b/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
@@ -466,11 +466,10 @@ object UnicomplexBoot extends LazyLogging {
     val ackFutures =
       for ((listenerName, config) <- boot.listeners) yield {
         val responseFuture = Unicomplex(actorSystem).uniActor ? StartListener(listenerName, config)
-        def log(t: Throwable): Unit = logger.error(s"Failed to start the listener, $listenerName.", t)
         responseFuture.onComplete {
-          case Failure(t)               => log(t)
-          case Success(StartFailure(t)) => log(t)
-          case _                        =>
+          case Failure(t) => logger.error(s"Failed to start the listener, $listenerName.", t)
+          case Success(StartFailure(t)) => logger.error(s"Failed to start the listener, $listenerName.", t)
+          case _ =>
         }
         responseFuture
       }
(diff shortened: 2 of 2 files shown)


crankydillo:
I'm starting to second guess...  I want to look into sending a failure message to the sender.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/Unicomplex.scala
```diff
@@ -472,7 +472,10 @@ class Unicomplex extends Actor with Stash with ActorLogging {
           },
             discardOld = false)
 
-        case Failure(t) => updateSystemState(checkInitState())
+        case Failure(t) => {
+          log.error(t, "Failed to start web service infrastructure.")
```
akara: Looks good. I've checked and we do not log similar messages in `UnicomplexBoot`. However, the `StartListener` message starts each listener. So it is probably better to log something like `log.error(t, "Failed to start service listener {}", name)`

crankydillo: That is much better.  Any thoughts on sending the message back and avoiding the timeout exception that occurs (see #683)?

akara: I see. That sounds reasonable. The next question is how should we handle the failure case? You may not run into a timeout exception. But should the failure to start a listener stop everything else? Or just warn, set system state to failed, and keep going?

crankydillo: Minimally, I want to replace the timeout exception with the `Listener xyz failed to start`.  As to whether that should crash the whole app, that is probably a decision best left to folks more familiar with squbs use cases:)  Personally, I like to crash my apps if the main functionality isn't working, but I can see how that wouldn't be universal.

crankydillo: I refactored it.  Let me know what you think.

crankydillo:
triggering ci

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
```diff
@@ -465,8 +465,16 @@ object UnicomplexBoot extends LazyLogging {
     implicit val timeout = Timeout((boot.listeners.size * 10).seconds)
     val ackFutures =
       for ((listenerName, config) <- boot.listeners) yield {
-        Unicomplex(actorSystem).uniActor ? StartListener(listenerName, config)
+        val fut = Unicomplex(actorSystem).uniActor ? StartListener(listenerName, config)
```
akara: Just a minor comment on `fut`. Can you make the val name a little more meaningful? `responseFuture` or something like that.

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
```diff
@@ -465,8 +465,16 @@ object UnicomplexBoot extends LazyLogging {
...
+        def log(t: Throwable): Unit = logger.error(s"Failed to start the listener, $listenerName.", t)
```
akara: That's too DRY! 😄 That much you can repeat. But don't need to change.

akara: Don't worry. This is not a serious one. Also the compiler should be able to inline these. To me it is just a trade off between readability and avoidance of duplication.

crankydillo:
triggering ci

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
anilgursel: I find this less readable as well.  But, not strong at all.  Especially, having a `def log` in between other lines requires me double check the curly braces if `responseFuture.onComplete` is inside `log` or not.

Just a minor comment.. shouldn't there be a ":" instead of "," at the end of "Failed to start the listener,"?

crankydillo:
trigger ci

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
anilgursel: I don't think there is a project convention.  Just grammar wise, having a comma makes the listener name optional.  Doesn't it?  If it is optional, then "the listener" would mean there is a single listener.  Maybe I get it wrong.  It is a minor comment anyway.

Regarding the previous comment, yes it is a standard Scala one-line definition.  My point is that, having it between two lines of a the parent function just degrades the readability.  While reading, it makes me double check if the next line belongs to the in-line function or not.  Maybe it is just me, but I find it less readable.  It might have been easier to read if `def log` is defined somewhere else.

Also, you are one of the main contributors of the project.

crankydillo: A few PRs != main contributor:P  As it turns out, I think you are correct about the ','  (I thought that was a '.').  I'll make the change.

crankydillo: I do find it ironic that I had the grammar bug in 2 places, which, of course, is due to copy-n-paste:P

crankydillo:
triggering ci

crankydillo:
triggering ci

squbs-unicomplex/src/main/scala/org/squbs/unicomplex/Unicomplex.scala
```diff
@@ -472,7 +472,10 @@ class Unicomplex extends Actor with Stash with ActorLogging {
           },
             discardOld = false)
 
-        case Failure(t) => updateSystemState(checkInitState())
+        case Failure(t) => {
+          sender ! StartFailure(t)
```
anilgursel: I see mostly `sender()`is used throughout the file, bu there are instances of `sender` as well.  Is there a difference between `sender` and `sender()`?  I would prefer to be consistent across.

Issue #569, opened 2018-01-03
Fix intermittent failure in travis CI


sebady:
Travis Intermittent test failure in BroadcastBufferSpec

akara:
We found another build issue to look at:

```
[info]      java.util.NoSuchElementException:
[info]       at scala.collection.LinearSeqOptimized.last(LinearSeqOptimized.scala:150)
[info]       at scala.collection.LinearSeqOptimized.last$(LinearSeqOptimized.scala:149)
[info]       at scala.collection.immutable.List.last(List.scala:89)
[info]       at org.squbs.stream.PerpetualStreamMatValueSpec.$anonfun$new$8(PerpetualStreamMatValueSpec.scala:77)
[info]       at org.squbs.stream.PerpetualStreamMatValueSpec.$anonfun$new$8$adapted(PerpetualStreamMatValueSpec.scala:74)
[info]       at org.squbs.stream.PerpetualStreamMatValueSpecHelper$.useSystem(PerpetualStreamMatValueSpec.scala:149)
[info]       at org.squbs.stream.PerpetualStreamMatValueSpec.$anonfun$new$7(PerpetualStreamMatValueSpec.scala:74)
[info]       at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
[info]       at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]       at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
```

akara:
One additional failure to look at:

```
[info] - should fail fast using default failure decider *** FAILED ***
[...]

PR #699, opened 2019-02-05
 #569 Poll a bit for actor system state to stabilize

crankydillo:
It's a bit of a hack, but likely justified.

- [X] Commits are squashed.
- [ ] Tests added.

### DIFF
--- a/squbs-unicomplex/src/test/scala/org/squbs/stream/PerpetualStreamMatValueSpec.scala
+++ b/squbs-unicomplex/src/test/scala/org/squbs/stream/PerpetualStreamMatValueSpec.scala
@@ -23,6 +23,8 @@ import akka.pattern._
 import akka.stream.{ActorMaterializer, ClosedShape, Materializer}
 import akka.stream.scaladsl.{Flow, GraphDSL, Keep, MergeHub, RunnableGraph, Sink, Source}
 import akka.util.Timeout
+import org.scalatest.concurrent.Eventually
+import org.scalatest.time.{Seconds, Span}
 import org.scalatest.{BeforeAndAfterAll, FunSpec, Matchers}
 import org.squbs.stream.PerpetualStreamMatValueSpecHelper.PerpStreamActors
 import org.squbs.unicomplex._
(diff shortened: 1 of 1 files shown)


squbs-unicomplex/src/test/scala/org/squbs/stream/PerpetualStreamMatValueSpec.scala
```diff
@@ -56,7 +61,7 @@ with BeforeAndAfterAll {
         useSystem[SinkMaterializingStream] {
           case Success(actor) =>
             val actorState = Await.result((actor ? StateRequest).mapTo[List[Long]], timeoutDuration)
-            actorState.last should be(PerpetualStreamMatValueSpecHelper.payload)
+            eventually { actorState.last should be(PerpetualStreamMatValueSpecHelper.payload) }
```
akara: While I think `eventually` is the right direction, `actorState` seems to be immutable. So if actorState was received and has no elements, eventually will also never be successful, right? The `eventually` clause should probably get a new list from the actor every time.

Please correct me if I'm wrong.

squbs-unicomplex/src/test/scala/org/squbs/stream/PerpetualStreamMatValueSpec.scala
```diff
@@ -74,7 +79,7 @@ with BeforeAndAfterAll {
             useSystem {
               case Success(actor) =>
                 val actorState = Await.result((actor ? StateRequest).mapTo[List[Long]], timeoutDuration)
-                actorState.last should be(PerpetualStreamMatValueSpecHelper.payload)
+                eventually { actorState.last should be(PerpetualStreamMatValueSpecHelper.payload) }
```
akara: Same comments here, too.

squbs-unicomplex/src/test/scala/org/squbs/stream/PerpetualStreamMatValueSpec.scala
crankydillo: Ugh, that is super embarrassing!

Issue #694, opened 2019-01-09
The time type in squbs should be consistent across all APIs 


sebady:

sebady:
We use both TimeUnit and Duration in various APIs.  We should be consistent here.

jiminhsieh:
I noticed that `TimeUnit` is only used at Java API.
https://github.com/paypal/squbs/blob/629803efdd1937f14057a93f39884281063b7c57/squbs-unicomplex/src/main/scala/org/squbs/lifecycle/GracefulStopHelper.scala#L117-L123
https://github.com/paypal/squbs/blob/629803efdd1937f14057a93f39884281063b7c57/squbs-unicomplex/src/main/scala/org/squbs/lifecycle/GracefulStopHelper.scala#L133-L141
https://github.com/paypal/squbs/blob/629803efdd1937f14057a93f39884281063b7c57/squbs-unicomplex/src/main/scala/org/squbs/lifecycle/GracefulStopHelper.scala#L146-L153

Since we need to consider Java developers, my idea is `java.time.Duration` for Java APIs and `scala.concurrent.duration.Duration` for Scala APIs.

sebady:
This sounds like a good idea to me.

akara:
I think we're pretty consistent on the Scala `Duration`, but all over the place for the Java time. So this change really focuses on the Java API.

jiminhsieh:
I noticed that some of Java APIs are used `scala.concurrent.duration.FiniteDuration`. I think it was not friendly for Java developers. They will need to import scala std library.
For example: 
https://github.com/paypal/squbs/blob/629803efdd1937f14057a93f39884281063b7c57/squbs-ext/src/main/scala/org/squbs/streams/circuitbreaker/impl/AtomicCircuitBreakerState.scala#L96-L116
https://github.com/paypal/squbs/blob/629803efdd1937f14057a93f39884281063b7c57/squbs-pattern/src/main/scala/org/squbs/pattern/timeoutpolicy/TimeoutPolicy.scala#L275-L279

My question is should we fix those APIs? If yes, should we fix with different PRs?

crankydillo:
@jiminhsieh It seems like that would fall under the scope of making a consistent experience for Java and Scala, so my vote is they should be fixed under the same PR that addresses this issue.  Is there some reason you think they may need a separate PR?

Anyhow, are you interested in working on this?  We're debating this being a part of the 0.12.0 release and are considering picking it up soon; however, we'd love to have a contribution here:)

[...]

PR #706, opened 2019-02-22
Java duration (#694)

jiminhsieh:

- [ ] Commits are squashed.
- [ ] Tests added.

It's still WIP, but I want to discuss some of the changes.

### DIFF
--- a/squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
+++ b/squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
@@ -16,7 +16,6 @@
 
 package org.squbs.streams
 
-import java.time.{Duration => JDuration}
 import java.util.concurrent.TimeUnit.NANOSECONDS
 import java.util.concurrent.{TimeUnit, TimeoutException}
 import java.util.function.Consumer
@@ -192,7 +191,7 @@ object Timeout {
   /**
     * Java API
     */
-  def create[In, Out, Context](timeout: JDuration):
+  def create[In, Out, Context](timeout: java.time.Duration):
   javadsl.BidiFlow[Pair[In, Context], Pair[In, Context], Pair[Out, Context], Pair[Try[Out], Context], NotUsed] =
     toJava(apply[In, Out, Context](FiniteDuration(timeout.toMillis, TimeUnit.MILLISECONDS)))
 
@@ -234,7 +233,7 @@ object TimeoutSettings {
   /**
     * Java API
     */
-  def create[In, Out, Context](timeout: JDuration): TimeoutSettings[In, Out, Context] =
+  def create[In, Out, Context](timeout: java.time.Duration): TimeoutSettings[In, Out, Context] =
     apply(FiniteDuration(timeout.toMillis, TimeUnit.MILLISECONDS))
 }
 
@@ -353,7 +352,7 @@ object TimeoutOrdered {
   /**
     * Java API
     */
-  def create[In, Out](timeout: JDuration,
+  def create[In, Out](timeout: java.time.Duration,
                       cleanUp: Consumer[Out]):
   akka.stream.javadsl.BidiFlow[In, In, Out, Try[Out], NotUsed] = {
     apply(FiniteDuration(timeout.toMillis, TimeUnit.MILLISECONDS), out => cleanUp.accept(out)).asJava
@@ -366,7 +365,7 @@ object TimeoutOrdered {
     * @tparam Out
     * @return
     */
-  def create[In, Out](timeout: JDuration):
+  def create[In, Out](timeout: java.time.Duration):
   akka.stream.javadsl.BidiFlow[In, In, Out, Try[Out], NotUsed] = {
     apply(FiniteDuration(timeout.toMillis, TimeUnit.MILLISECONDS), (_: Out) => ()).asJava
   }
--- a/squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
+++ b/squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
@@ -355,7 +355,7 @@ object TimeoutOrdered {
   def create[In, Out](timeout: java.time.Duration,
                       cleanUp: Consumer[Out]):
   akka.stream.javadsl.BidiFlow[In, In, Out, Try[Out], NotUsed] = {
-    apply(FiniteDuration(timeout.toMillis, TimeUnit.MILLISECONDS), out => cleanUp.accept(out)).asJava
+    apply(FiniteDuration(timeout.toMillis, TimeUnit.MILLISECONDS), out: Out => cleanUp.accept(out)).asJava
   }
 
   /**
--- a/squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
+++ b/squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
@@ -355,7 +355,7 @@ object TimeoutOrdered {
   def create[In, Out](timeout: java.time.Duration,
                       cleanUp: Consumer[Out]):
   akka.stream.javadsl.BidiFlow[In, In, Out, Try[Out], NotUsed] = {
-    apply(FiniteDuration(timeout.toMillis, TimeUnit.MILLISECONDS), out: Out => cleanUp.accept(out)).asJava
+    apply(FiniteDuration(timeout.toMillis, TimeUnit.MILLISECONDS), (out: Out) => cleanUp.accept(out)).asJava
   }
 
   /**
--- a/squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
+++ b/squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
@@ -194,7 +194,7 @@ object Timeout {
     */
   def create[In, Out, Context](timeout: JDuration):
   javadsl.BidiFlow[Pair[In, Context], Pair[In, Context], Pair[Out, Context], Pair[Try[Out], Context], NotUsed] =
-    toJava(apply[In, Out, Context](FiniteDuration(timeout.toMillis, TimeUnit.MICROSECONDS)))
+    toJava(apply[In, Out, Context](FiniteDuration(timeout.toMillis, TimeUnit.MILLISECONDS)))
 
 }
 
--- a/squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
+++ b/squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
@@ -16,8 +16,9 @@
 
 package org.squbs.streams
 
+import java.time.{Duration => JDuration}
 import java.util.concurrent.TimeUnit.NANOSECONDS
-import java.util.concurrent.TimeoutException
+import java.util.concurrent.{TimeUnit, TimeoutException}
 import java.util.function.Consumer
 
 import akka.NotUsed
@@ -191,9 +192,9 @@ object Timeout {
   /**
     * Java API
     */
-  def create[In, Out, Context](timeout: FiniteDuration):
+  def create[In, Out, Context](timeout: JDuration):
   javadsl.BidiFlow[Pair[In, Context], Pair[In, Context], Pair[Out, Context], Pair[Try[Out], Context], NotUsed] =
-    toJava(apply[In, Out, Context](timeout))
+    toJava(apply[In, Out, Context](FiniteDuration(timeout.toMillis, TimeUnit.MICROSECONDS)))
 
 }
 
@@ -233,7 +234,8 @@ object TimeoutSettings {
   /**
     * Java API
     */
-  def create[In, Out, Context](timeout: FiniteDuration): TimeoutSettings[In, Out, Context] = apply(timeout)
+  def create[In, Out, Context](timeout: JDuration): TimeoutSettings[In, Out, Context] =
+    apply(FiniteDuration(timeout.toMillis, TimeUnit.MILLISECONDS))
 }
 
 /**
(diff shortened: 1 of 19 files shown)


squbs-pattern/src/test/scala/org/squbs/pattern/timeoutpolicy/TimeoutPolicySpec.scala
```diff
@@ -42,8 +42,8 @@ class TimeoutPolicySpec extends FlatSpecLike with Matchers{
   "Run FixedTimeoutPolicy in function" should "work" in {
     val policy = TimeoutPolicy(name = Some("test"), initial = 1.second, fixedRule, minSamples = 1)
 
-    for(i <- 0 until 10) {
-      policy.execute(timeout => {
+    for(_ <- 0 until 10) {
+      policy.execute((timeout: FiniteDuration) => {
```
jiminhsieh: After I chaged the blow part: 
https://github.com/paypal/squbs/pull/706/files#diff-5a24577de789cacc6fa9751834aa36c3R97
https://github.com/paypal/squbs/pull/706/files#diff-6f9c73a3eba62c02bf35c3d0479a1f1fR23
I have to add type for the `timeout`, otherwise it will cause error:
```
[error] ~/repository/github/squbs/squbs-pattern/src/test/scala/org/squbs/pattern/timeoutpolicy/TimeoutPolicySpec.scala:46:14: ambiguous reference to overloaded definition,
[error] both method execute in class TimeoutPolicy of type [T](f: org.squbs.pattern.timeoutpolicy.TimedFn[T])T
[error] and  method execute in class TimeoutPolicy of type [T](f: scala.concurrent.duration.FiniteDuration => T)T
[error] match argument types (Comparable[_ >: java.time.Duration with scala.concurrent.duration.Duration <: Comparable[_ >: java.time.Duration with scala.concurrent.duration.Duration <: java.io.Serializable] with java.io.Serializable] with java.io.Serializable => Unit)
[error]       policy.execute(timeout => {
[error]              ^
[error] one error found
```

squbs-httpclient/src/test/java/org/squbs/httpclient/ClientFlowTest.java
```diff
@@ -190,9 +189,9 @@ public void testAllInputsAPI() throws Exception {
         assertJmxValue(oNameHC, "MaxConnections", 41);
         assertJmxValue(oNameHC, "Environment", "PROD");
         assertJmxValue(oNameCB, "MaxFailures", 11);
-        assertJmxValue(oNameCB, "CallTimeout", "12 seconds");
-        assertJmxValue(oNameCB, "ResetTimeout", "13 minutes");
-        assertJmxValue(oNameCB, "MaxResetTimeout", "14 days");
+        assertJmxValue(oNameCB, "CallTimeout", "12000 milliseconds");
+        assertJmxValue(oNameCB, "ResetTimeout", "780000 milliseconds");
+        assertJmxValue(oNameCB, "MaxResetTimeout", "1209600000 milliseconds");
         assertJmxValue(oNameCB, "ExponentialBackoffFactor", 16.0);
```
jiminhsieh: After I replaced to `java.time.duration`, I have to change the expected value for passing unit testing. The problem is `java.time.duration` doesn't have the original time unit.

squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
```diff
@@ -191,9 +192,9 @@ object Timeout {
   /**
     * Java API
     */
-  def create[In, Out, Context](timeout: FiniteDuration):
+  def create[In, Out, Context](timeout: JDuration):
```
anilgursel: For Java APIs, I personally prefer specifying `java.time.Duration` when there is confusion.  Just because Java users are usually not familiar with `import java.time.{Duration => JDuration}` syntax and makes it harder for them to understand at first look.

squbs-ext/src/test/java/org/squbs/streams/TimeoutTest.java
```diff
@@ -87,7 +85,7 @@ public void testFlowsWithMessageOrderGuaranteeAndCleanUp() throws ExecutionExcep
         Consumer<String> cleanUp = s -> counter.incrementAndGet();
 
         final BidiFlow<String, String, String, Try<String>, NotUsed> timeoutBidiFlow =
-                TimeoutOrdered.create(Timing.timeout(), cleanUp);
+                TimeoutOrdered.create(Duration.ofMinutes(Timing.timeout().toMinutes()), cleanUp);
```
anilgursel: `Timing.timeout()` is set to `1 second`.  Any specific reason to convert it to minutes here?

squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
```diff
@@ -191,9 +192,9 @@ object Timeout {
...
   javadsl.BidiFlow[Pair[In, Context], Pair[In, Context], Pair[Out, Context], Pair[Try[Out], Context], NotUsed] =
-    toJava(apply[In, Out, Context](timeout))
+    toJava(apply[In, Out, Context](FiniteDuration(timeout.toMillis, TimeUnit.MICROSECONDS)))
```
anilgursel: Is `TimeUnit.MICROSECONDS` correct here?  Shouldn't it be `TimeUnit.MILLISECONDS`?

squbs-ext/src/test/java/org/squbs/streams/TimeoutTest.java
```diff
@@ -256,7 +254,7 @@ public void testWithCleanUp() throws ExecutionException, InterruptedException {
                         .map(elem -> (Pair<String, UUID>)elem);
 
         Consumer<String> cleanUp = s -> counter.incrementAndGet();
-        TimeoutSettings settings = TimeoutSettings.<String, String, UUID>create(Timing.timeout()).withCleanUp(cleanUp);
+        TimeoutSettings settings = TimeoutSettings.<String, String, UUID>create(Duration.ofMinutes(Timing.timeout().toMinutes())).withCleanUp(cleanUp);
```
anilgursel: Ditto

squbs-ext/src/test/java/org/squbs/streams/TimeoutTest.java
```diff
@@ -295,7 +293,7 @@ public MyContext(String s, UUID uuid) {
                         .map(elem -> (Pair<String, MyContext>)elem);
 
         Consumer<String> cleanUp = s -> counter.incrementAndGet();
-        TimeoutSettings settings = TimeoutSettings.<String, String, MyContext>create(Timing.timeout())
+        TimeoutSettings settings = TimeoutSettings.<String, String, MyContext>create(Duration.ofMinutes(Timing.timeout().toMinutes()))
```
anilgursel: Ditto

squbs-ext/src/test/java/org/squbs/streams/circuitbreaker/CircuitBreakerTest.java
```diff
@@ -37,7 +37,7 @@
 import org.squbs.streams.Timing;
 import org.squbs.streams.circuitbreaker.impl.AtomicCircuitBreakerState;
 import org.squbs.streams.circuitbreaker.japi.CircuitBreakerSettings;
-import scala.concurrent.duration.Duration;
+//import scala.concurrent.duration.Duration;
```
anilgursel: Please delete this line.

squbs-ext/src/test/java/org/squbs/streams/circuitbreaker/CircuitBreakerTest.java
```diff
@@ -51,14 +51,15 @@
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
+import java.time.Duration;
```
anilgursel: Can you please do a "Optimize Imports" in IntelliJ or the similar feature in the IDE of your choice.  I would expect this line to come earlier than the `java.util` ones.

squbs-httpclient/src/test/java/org/squbs/httpclient/ClientFlowTest.java
```diff
@@ -143,9 +142,9 @@ public void testBuilderAPI() throws Exception {
         assertJmxValue(oNameHC, "MaxConnections", 41);
         assertJmxValue(oNameHC, "Environment", "PROD");
         assertJmxValue(oNameCB, "MaxFailures", 11);
-        assertJmxValue(oNameCB, "CallTimeout", "12 seconds");
-        assertJmxValue(oNameCB, "ResetTimeout", "13 minutes");
-        assertJmxValue(oNameCB, "MaxResetTimeout", "14 days");
+        assertJmxValue(oNameCB, "CallTimeout", "12000 milliseconds");
+        assertJmxValue(oNameCB, "ResetTimeout", "780000 milliseconds");
```
anilgursel: It would be helpful to add what these numbers mean in comments, e.g., `// 12 seconds`

squbs-pattern/src/main/scala/org/squbs/pattern/timeoutpolicy/TimeoutPolicy.scala
```diff
@@ -17,6 +17,8 @@
 package org.squbs.pattern.timeoutpolicy
 
 import java.lang.management.ManagementFactory
+import java.time.{Duration => JDuration}
```
anilgursel: The same comment regarding Java users..

squbs-unicomplex/src/test/java/org/squbs/stream/ProperShutdownStreamJ.java
```diff
@@ -75,7 +76,7 @@ public Receive createReceive() {
         ActorRef actorRef = matValue().first();
         CompletionStage<Long> fCount = matValue().second();
         CompletionStage<Boolean> fStopped =
-                gracefulStop(actorRef, Timeouts.awaitMax().toMillis(), TimeUnit.MILLISECONDS);
+                gracefulStop(actorRef, Duration.ofMinutes(Timeouts.awaitMax().toMinutes()));
```
anilgursel: I would prefer using seconds here in case `awaitMax` gets decreases to lower than a minute.  `awaitMax` was increased later to provide better stability for some tests, but if that issue gets addresses, better to decrease it to a lower value.

squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
akara: Please see https://github.com/scala/scala-java8-compat/pull/86. You should be able to use the Java8 compat library for the conversions. It takes care of more use cases.

squbs-unicomplex/src/test/java/org/squbs/stream/ProperShutdownStreamJ.java
akara: Again, I'd use the time converter directly.

squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
jiminhsieh: I saw that PR too. The problem is we have to bump `scala-java8-compat` to 0.9.0. Now, akka still uses 0.8.0. They didn't use 0.9.0, because it probably causes the breaking change.
https://github.com/akka/akka/blob/26b086931804cf2bc37481f08ed9f14e8ebf343b/project/Dependencies.scala#L33-L39
Here is part of commit message.
> As it looks like 0.8.0 is not bincompat with 0.9.0

squbs-ext/src/test/java/org/squbs/streams/TimeoutTest.java
jiminhsieh: You're right. Thanks for finding this. I will fix soon.

squbs-ext/src/test/java/org/squbs/streams/circuitbreaker/CircuitBreakerTest.java
jiminhsieh: Got it!

squbs-unicomplex/src/test/java/org/squbs/stream/ProperShutdownStreamJ.java
jiminhsieh: We will need to use `scala-java8-compat` 0.9.0. The thing is akka team only updated at scala 2.13.x. Here is that commit: https://github.com/akka/akka/commit/49322afae0c7c0f96c76d5131207ac55a8f92cd3

squbs-httpclient/src/test/java/org/squbs/httpclient/ClientFlowTest.java
jiminhsieh: Thanks!

squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala
akara: Got it, thanks!

akara:
The time display in milliseconds is a little disturbing to me. Let me think it over and see what I can do. Otherwise I'm all good with this PR.

squbs-httpclient/src/test/java/org/squbs/httpclient/ClientFlowTest.java
```diff
@@ -143,9 +142,9 @@ public void testBuilderAPI() throws Exception {
         assertJmxValue(oNameHC, "MaxConnections", 41);
         assertJmxValue(oNameHC, "Environment", "PROD");
         assertJmxValue(oNameCB, "MaxFailures", 11);
-        assertJmxValue(oNameCB, "CallTimeout", "12 seconds");
-        assertJmxValue(oNameCB, "ResetTimeout", "13 minutes");
-        assertJmxValue(oNameCB, "MaxResetTimeout", "14 days");
+        assertJmxValue(oNameCB, "CallTimeout", "12000 milliseconds"); // 12 seconds
+        assertJmxValue(oNameCB, "ResetTimeout", "780000 milliseconds"); // 13 minutes
+        assertJmxValue(oNameCB, "MaxResetTimeout", "1209600000 milliseconds"); // 14 days
```
akara: I actually do not like such JMX displays at all. I understand the limitations of `java.time.Duration`, however, I think we can do much better. Instead of diluting the check to milliseconds, would it be better to change `CircuitBreakerStateMXBeanImpl`? For instance, we can change `override def getMaxResetTimeout: String = maxResetTimeout.toString` to `override def getMaxResetTimeout: String = maxResetTimeout.toCoarsest.toString`. That would be much more consumable in JMX.
Given, `toCoarsest` goes through a few recursions. So it is not totally free or super low cost. Placing this conversion at strategic points is important. Should we do this in the JMX bean? Should we do it at construction of the state object? Which one has the least runtime cost? JMX bean is generally not too bad. But since these are timeouts, the creation of objects that are only done at creation/materialization time is also not a bad choice. @anilgursel, what are your thoughts?

anilgursel: I agree @akara.  Creation/materialization time makes sense.

akara:
Looks like there is a good amount of merge conflict with the previously approved PR on Scala 2.11 support. Pulling code down and fixing those conflicts.

PR #721, opened 2019-04-27
Squbs bootstrap process scans class path or class loader?

mdufresne:
I've never heard of scanning the class loader.  I'm familiar with class path scanning.
Did a quick google search as well and nothing came up for "scanning class loader".

I'm assuming class path was meant.


sebady:
Thank you for your contribution.

Issue #711, opened 2019-03-11
Service infra timeout needs to be more configurable


crankydillo:
From what I can tell, the only way to increase the service infra timeout is to run additional listeners ([source](https://github.com/paypal/squbs/blob/1f2398c20ea318fe1f611289a2a466174aee4996/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala#L465)).  We need to be able to change that service infra start timeout via configuration..

Update:  Make sure you have good error messages.  The [code](https://github.com/paypal/squbs/blob/af38fa2e7873540dae0101bbfda1919de5766434/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala#L494) as it stand at time of writing uses the same string for both a timeout and a 'start failure', which may also be a timeout failure.  So when you get a timeout, it's not crystal clear if the timeout is happening in the proposed 'fix' or some underlying layer.

akara:
In essence, make the timeout in startServiceInfra configurable. Sure.

crankydillo:
Work done in linked PRs.

PR #728, opened 2019-08-09
 #711 proposal for better error messages..

crankydillo:
The fix we have on master for 711 has error messages that make it hard to triage.  This addresses that.

- [ x] Title includes issue id.
- [x ] Description of the change added.
- [ x] Commits are squashed.
- [ ] Tests added.

### DIFF
--- a/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
+++ b/squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
@@ -491,8 +491,12 @@ object UnicomplexBoot extends LazyLogging {
       for ((listenerName, config) <- boot.listeners) yield {
         val responseFuture = Unicomplex(actorSystem).uniActor ? StartListener(listenerName, config)
         responseFuture.onComplete {
-          case Failure(t) => logger.error(s"Failed to start the listener $listenerName.", t)
-          case Success(StartFailure(t)) => logger.error(s"Failed to start the listener $listenerName.", t)
+          case Failure(t) if (t.isInstanceOf[TimeoutException]) =>
+            logger.error(s"The Unicomplex could not start the listener, $listenerName, within $to.", t)
+          case Failure(t) =>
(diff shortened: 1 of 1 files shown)


squbs-unicomplex/src/main/scala/org/squbs/unicomplex/UnicomplexBoot.scala
```diff
@@ -491,8 +491,12 @@ object UnicomplexBoot extends LazyLogging {
       for ((listenerName, config) <- boot.listeners) yield {
         val responseFuture = Unicomplex(actorSystem).uniActor ? StartListener(listenerName, config)
         responseFuture.onComplete {
-          case Failure(t) => logger.error(s"Failed to start the listener $listenerName.", t)
-          case Success(StartFailure(t)) => logger.error(s"Failed to start the listener $listenerName.", t)
+          case Failure(t) if (t.isInstanceOf[TimeoutException]) =>
+            logger.error(s"The Unicomplex could not start the listener, $listenerName, within $to.", t)
```
anilgursel: is it guaranteed that `TimeoutException` would be thrown only after `to`?  Some other component may timeout on some other operation and could potentially send `TimeoutException` back to Unicomplex?

crankydillo: That is the problem I'm trying to deal with.  That is happening a lot and the error message makes it confusing what's happening.  If some other operation times out, that should come back in `Success(StartFailure(t))` case and you will see that in the stack trace.

crankydillo: So in your case what that user would see is:
```
The Unicomplex reported a start failure for the listener, foo listener.  {TimeoutException.stacktrace}
```

crankydillo: @akara Any opinion on this?

And I hope all is well!

PR #729, opened 2019-08-15
 #711 Document service infra timeouts.

crankydillo:

- [x ] Title includes issue id.
- [ x] Description of the change added.
- [ x] Commits are squashed.
- [ ] Tests added.
- [ x] Documentation added/updated.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -21,6 +21,14 @@ squbs {
   # loaded during squbs initialization for Actor System settings. Implicit "application.conf" will be loaded
   # besides this file name list
   external-config-files = []
+
+  # Service infra configuration.
+  service-infra {
+    # Maximum amount of time to wait for all listeners to be started.
+    timeout = 60s
+    # Maximum amount of time each listener is given to start.
+    listener-timeout = 10s
+  }
 }
 
 default-listener {
--- a/squbs-unicomplex/src/main/resources/reference.conf
+++ b/squbs-unicomplex/src/main/resources/reference.conf
@@ -21,7 +21,7 @@ squbs {
   service-infra {
     # Maximum amount of time to wait for all listeners to be started.
     timeout = 60s
-    # Maximum mount of time each listener is given to start.
+    # Maximum amount of time each listener is given to start.
     listener-timeout = 10s
   }
 }


akara:
I'm good with this change, assuming it reflects the code change. Thanks for doing this.

docs/configuration.md
```diff
@@ -134,4 +142,4 @@ A materializer is declared at the root level of the configuration file. The name
 
 ## Pipeline
 
-If defined, a default pipeline is installed for pre-processing every single request and post-processing every response. Services can specify a different pipeline, or none at all as described under [Bootstrapping squbs](bootstrap.md#services). Applications or infrastructure can implement their own pipelines for pre-processing needs such as logging or tracing. Please see detailed description of pipelines under [Request/Response Pipeline](pipeline.md).
\ No newline at end of file
+If defined, a default pipeline is installed for pre-processing every single request and post-processing every response. Services can specify a different pipeline, or none at all as described under [Bootstrapping squbs](bootstrap.md#services). Applications or infrastructure can implement their own pipelines for pre-processing needs such as logging or tracing. Please see detailed description of pipelines under [Request/Response Pipeline](pipeline.md).
```
anilgursel: I can't see what the change here is. CRLF?

docs/configuration.md
```diff
@@ -21,6 +21,14 @@ squbs {
   # loaded during squbs initialization for Actor System settings. Implicit "application.conf" will be loaded
   # besides this file name list
   external-config-files = []
+
+  # Service infra configuration.
+  service-infra {
```
anilgursel: Since the code is already in, please ignore this comment.  I should have commented earlier.  But, I think `service-infra` is an implementation detail in code, so I believe the following would be a better configuration name:

```
squbs
    startup {
        # Maximum amount of time to wait for all listeners to be started.
        timeout = 60s
        # Maximum amount of time each listener is given to start.
        listener-timeout = 10s
      }
}
```

docs/configuration.md
crankydillo: I couldn't either, but I didn't try very hard..

docs/configuration.md
crankydillo: It's easy to revisit..  I didn't really like `service-infra` either; however, I think `squbs.startup.timeout` is too broad IMO.  Of course, the comment helps.  Anyhow, it's definitely not too late to change, and we don't have to have a long debate on what the names should be.  If you want to change to your suggestion or make a slight tweak based  on my feedback, just let me know.  Easy to do.

PR #736, opened 2020-02-28
PersistentBufferAtLeastOnce: Don't drop elements if no commits are made

longshorej:
When the queue is started, it attempts to resume from the old
index. However, this operation may fail if e.g. no commits have
occurred yet.

This ensures that the first element is only dropped if resuming
from an index succeeds, and adds a test to validate this behavior.


longshorej:
Thanks for the review @sebady. I'm just curious what the release timeframe is like after this is merged, is it a big deal for you guys to kick off a release?

sebady:
We are working on an internal GA release at the moment on another stack.  Once your PR merges we should be able to get a squbs release out in a few days. It may contain just this fix and could be released as `0.13.1`

sebady:
@anilgursel, @crankydillo Can one of you take a second look at this PR.

longshorej:
Hi @sebady - any update on this? Thanks!

sebady:
@anilgursel, @crankydillo, @akara This change looks good but can one of you give another thumbs up on this one?

longshorej:
Thanks!

Issue #750, opened 2022-03-06
MaterializationMetricsCollectorSpec fails intermittently


akara:
Especially on GitHub Actions, we see a good failure rate on this test alone.

akara:
Issue fixed with PR

PR #751, opened 2022-03-06
#750 Fixed intermittent test failures in MaterializationMetricsCollectorSpec

akara:

- [x] Commits are squashed.
- [ ] Tests added.


squbs-ext/src/main/scala/org/squbs/metrics/MetricsFlow.scala
```diff
@@ -80,7 +80,8 @@ object MaterializationMetricsCollector {
   def create[T](name: String, system: ActorSystem) = apply[T](name)(system)
 }
 
-class MaterializationMetricsCollector[T] private[metrics] (name: String)(implicit system: ActorSystem) extends GraphStage[FlowShape[T, T]] {
+class MaterializationMetricsCollector[T] private[metrics] (name: String)(implicit system: ActorSystem)
```
akara, line 83: Just style changes

squbs-ext/src/test/scala/org/squbs/metrics/MaterializationMetricsCollectorSpec.scala
```diff
@@ -62,11 +62,13 @@ class MaterializationMetricsCollectorSpec extends TestKit(ActorSystem("Materiali
     probe2.sendComplete()
     probe3.sendComplete()
 
-    Future.sequence(future2 :: future3:: Nil) map { _ =>
-      jmxValue("upstream-finishes-active-count", "Count").value shouldBe 0
-      jmxValue("upstream-finishes-creation-count", "Count").value shouldBe 3
-      jmxValue("upstream-finishes-termination-count", "Count").value shouldBe 3
-    }
```
akara, line 69: There is a slight problem with doing the assertions in `map`. There is no guarantee the future would complete. In the test this future does not complete in time and the assertions are never exercised, causing the test to pass. When they are exercised, the test sometimes fail. Changing to the new way where we really block on each future and assert subsequently guarantees the future complets (success or failure). Then we do the assertions in the same thread never escaping the thread scope of the test. This test is more reliable.

crankydillo (approved):
Looks good to me.

On a related note, asserting on asynchronously-provided values has caused me much grief over the past 2 years!

